PR #2, opened 2019-02-18
Allow MonitorManager to send kafka events

harageth:
This creates a class that will send Monitor Events to the Ambassador through Kafka.
It rely's on https://github.com/racker/salus-telemetry-model/pull/26 to provide the common data model.

Since we are not doing anything with labels right now and the CRUD operations aren't merged in I added a few sample lines of code on after consuming the ResourceEvents to post one of these to Kafka so that we can see that it is working for the demo.


src/main/java/com/rackspace/salus/monitor_management/services/MonitorEventProducer.java
```diff
@@ -0,0 +1,36 @@
...
+    @Autowired
+    public MonitorEventProducer(KafkaTemplate<String,Object> kafkaTemplate, KafkaTopicProperties properties) {
+        this.kafkaTemplate = kafkaTemplate;
+        this.properties= properties;
+    }
+
+    // ok so we need to figure out what we are sending
+    //unless we want to just pass into this function everything
+    public void sendMonitorEvent(MonitorEvent event) {
+        final String topic = properties.getEvents();
```
itzg: Oh, I guess you'll need to add a new topic and therefore a new entry in `KafkaTopicProperties` like how `resources` is. The events topic was added during presence monitor development for external facing events...like alert style events.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -9,9 +11,11 @@
 @Slf4j
 @Service
 public class MonitorManagement {
+    private final MonitorEventProducer monitorEventProducer;
 
     @Autowired
-    public MonitorManagement() {
+    public MonitorManagement(MonitorEventProducer monitorEventProducer) {
+        this.monitorEventProducer = monitorEventProducer;
     }
```
GeorgeJahad: there are going to be a few merge conflicts, between your code and mine, but nothing too bad.

GeorgeJahad: where is the handleResourceEvent() called from?

harageth: It gets called from the ResourceEventListener which is the Kafka Ingestion class.

PR #3, opened 2019-02-19
Add Monitor Manager crud/rest api calls

GeorgeJahad:
This is a basically just a copy of the resource manager's CRUD/Rest Api methods, that has been repurposed to work for monitors.  (So that means I get to ask @jjbuchan for a review even though he is working on UMB

# Resolves:
https://jira.rax.io/browse/SALUS-187
https://jira.rax.io/browse/SALUS-130


src/main/resources/application.yml
```diff
@@ -16,5 +16,28 @@ spring:
 logging:
   level:
     web: debug
+---
```
itzg: Rather than a sub-document for the `dev` profile, please move this to a separate `application-dev.yml` file. With that you can remove the `spring.profiles` key.

GeorgeJahad: @itzg out of curiosity, why keep multiple yml files?  isn't it "DRY"er to use sub-documents and keep it all in one file?

itzg: There's no repetition between the files (or shouldn't be), so technically not violating DRY :)

After having tried out both approaches, I just found it much clearer to locate and know the Spring profile overrides exist by separate files. Otherwise you have to look into the `application.yml` to see the dev profile is there.

GeorgeJahad: that's fine with me, spring can be pretty opaque, and any conventions to make things clearer are a win in my book.  but ,  just to explain what i meant by "dry", all the spring.kafka parameters at the top of the file are shared both by prod and dev.  won't they have to now be duplicated in the second file?

or am i misunderstanding how things work?

itzg: Ah, now I understand where you're coming from.

First, please switch over to the consolidated kafka topic config like what I did here https://github.com/racker/salus-telemetry-resource-management/pull/8 and with that you'll see the topic config can be removed entirely from the yaml since defaults suffice.

As for the other part, the `application.yml` is always processed first. Spring then looks for each `application-{profile}.yml` for each activated profile and overlays the config in there. So the only "duplication" should be things overridden between "default" and the given profile.

GeorgeJahad:
the last commit should contain all the requested changes, plus those needed for the review changes done for the model pr: https://github.com/racker/salus-telemetry-model/pull/27

GeorgeJahad:
oops, just commited the application.yml changes

GeorgeJahad:
@itzg the kafka properties in the application.yml file were not topic related and I believe are all still necessary.  @harageth had already added the topic annotation to the master branch, and i rebased this branch against it before pushing.  so i don't think any more needs to be done there.

src/main/resources/application.yml
itzg: Sorry, I had gotten myself confused about which kafka properties were there.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1,19 +1,203 @@
...
+    }
+
+    /**
+     * Gets an individual monitor object by the public facing id.
+     *
+     * @param tenantId  The tenant owning the monitor.
+     * @param monitorId The unique value representing the monitor.
+     * @return The monitor object.
+     */
+    public Monitor getMonitor(String tenantId, String monitorId) {
```
jjbuchan: I mentioned this in the model pr, but I do wonder about whether we should use this type of `monitorId` field vs. an auto-genarated one.  We could generate uuids for these ids.

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorUpdate.java
```diff
@@ -0,0 +1,37 @@
...
+import org.hibernate.validator.constraints.NotBlank;
+import java.io.Serializable;
+import java.util.Map;
+
+@Data
+public class MonitorUpdate implements Serializable {
+    Map<String,String> labels;
+
+    @NotBlank
+    String content;
```
jjbuchan: Should this not be optional?  Maybe they just want to change the labels and leave the content as is.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1,19 +1,203 @@
...
+        Monitor monitor = getMonitor(tenantId, monitorId);
+        if (monitor == null) {
+            throw new NotFoundException(String.format("No monitor found for %s on tenant %s",
+                    monitorId, tenantId));
+        }
+        PropertyMapper map = PropertyMapper.get();
+        map.from(updatedValues.getLabels())
+                .whenNonNull()
+                .to(monitor::setLabels);
+        monitor.setContent(updatedValues.getContent());
```
jjbuchan: Added comment below about this being optional, which may lead to this line needing to become a `map.from` line.

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorUpdate.java
```diff
@@ -0,0 +1,37 @@
...
+
+    String targetTenant;
+
+    String selectorScope = "ALL_OF";
```
jjbuchan: I don't think we'd want this to be available for customers to change, do we?

src/main/java/com/rackspace/salus/monitor_management/web/error/GenericError.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.monitor_management.web.error;
+
+import lombok.Data;
+
+@Data
+public class GenericError {
```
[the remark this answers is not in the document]
jjbuchan: It's so the erorr in https://github.com/racker/salus-telemetry-monitor-management/pull/3/files#diff-d92e2722f1d4d94a45de8b4e9c2b9c9aR38 gets converted to json rather than a string.

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorUpdate.java
itzg: Good point. I'll have to remember my logic for including that in the public api-model from the start. It might have been just temporary until we got to the poller aspect of monitors.

PR #4, opened 2019-02-26
replaced monitorId with id, made id a uuid, and added optional monitor Name

GeorgeJahad:
https://jira.rax.io/browse/SALUS-202

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApi.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApi.java
@@ -74,15 +74,14 @@ public class MonitorApi {
         return emitter;
     }
 
-    @GetMapping("/tenant/{tenantId}/monitors/{id}")
+    @GetMapping("/tenant/{tenantId}/monitors/{uuid}")
     public Monitor getById(@PathVariable String tenantId,
-                                  @PathVariable String id) throws NotFoundException {
-        UUID uuid = UUID.fromString(id);
+                                  @PathVariable UUID uuid) throws NotFoundException {
 
         Monitor monitor = monitorManagement.getMonitor(tenantId, uuid);
         if (monitor == null) {
             throw new NotFoundException(String.format("No monitor found for %s on tenant %s",
-                    id, tenantId));
+                    uuid, tenantId));
         }
         return monitor;
     }
@@ -103,19 +102,17 @@ public class MonitorApi {
         return monitorManagement.createMonitor(tenantId, input);
     }
 
-    @PutMapping("/tenant/{tenantId}/monitors/{id}")
+    @PutMapping("/tenant/{tenantId}/monitors/{uuid}")
     public Monitor update(@PathVariable String tenantId,
-                          @PathVariable String id,
+                          @PathVariable UUID uuid,
                           @Valid @RequestBody final MonitorUpdate input) throws IllegalArgumentException {
-        UUID uuid = UUID.fromString(id);
         return monitorManagement.updateMonitor(tenantId, uuid, input);
     }
 
-    @DeleteMapping("/tenant/{tenantId}/monitors/{id}")
+    @DeleteMapping("/tenant/{tenantId}/monitors/{uuid}")
     @ResponseStatus(HttpStatus.NO_CONTENT)
     public void delete(@PathVariable String tenantId,
-                       @PathVariable String id) {
-        UUID uuid = UUID.fromString(id);
+                       @PathVariable UUID uuid) {
         monitorManagement.removeMonitor(tenantId, uuid);
     }
 }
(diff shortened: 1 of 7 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApi.java
```diff
@@ -73,14 +74,15 @@ public SseEmitter getAllAsStream() {
         return emitter;
     }
 
-    @GetMapping("/tenant/{tenantId}/monitors/{monitorId}")
-    public Monitor getByMonitorId(@PathVariable String tenantId,
-                                  @PathVariable String monitorId) throws NotFoundException {
+    @GetMapping("/tenant/{tenantId}/monitors/{id}")
+    public Monitor getById(@PathVariable String tenantId,
+                                  @PathVariable String id) throws NotFoundException {
```
itzg: I wonder if `@PathVariable` supports UUID natively rather than having to do the `fromString` conversion.

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorCreate.java
```diff
@@ -29,8 +29,8 @@
 
 @Data
 public class MonitorCreate implements Serializable {
-    @NotBlank
-    String monitorId;
```
itzg: The monitor events and ultimately the Envoy still need the primary key of the monitor. The Envoy uses that to guarantee uniquely and deterministically identified configuration files, one per monitor.

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorUpdate.java
```diff
@@ -28,6 +28,8 @@
 public class MonitorUpdate implements Serializable {
     Map<String,String> labels;
 
+    String monitorName;
```
itzg: Needs the ID

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorCreate.java
```diff
@@ -29,8 +29,8 @@
 
 @Data
 public class MonitorCreate implements Serializable {
```
itzg: Not directly related to this PR, but I'm now thinking these event objects belong over in the salus-telemetry-model repo since the Ambassador will need deserialize these out of Kafka.

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorCreate.java
GeorgeJahad: The primary key is actually created by the SQL create function.  It doesn't exist at the time this  class is used.  The line removed here is not the primary key.  What is actually happening in this line is a rename of the field from monitorId to monitorName.

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorUpdate.java
GeorgeJahad: perhaps my understanding of spring boot request bodies is wrong, but the intent of this class is to be used here: https://github.com/racker/salus-telemetry-monitor-management/blob/george/addMonitorName/src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApi.java#L109

the id is a path variable and not in the body.

am I misunderstanding the usage?

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorCreate.java
GeorgeJahad: i don't think these get sent to kafka.  as i said they are intended to be webmvc @RequestBody parameters

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorCreate.java
itzg: I'm sorry, yeah, this is the web model, not the kafka message model.

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorUpdate.java
itzg: No, you're right. Again, I got confused with web model vs kafka model.

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApi.java
GeorgeJahad: Of course, you are correct.  Fixed in the latest commit

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorCreate.java
itzg: ...I later realized the public API module will need these for building the request body, but I'm going to research some maven techniques to pull that off.

PR #6, opened 2019-03-03
Lookup envoys and send corresponding events, as resources/monitors change

GeorgeJahad:
# Resolves

https://jira.rax.io/browse/SALUS-126

# What
Publishes events to kafka for Monitor and Resource changes

# How
Finds the envoy corresponding to the Monitor/Resource being changed and addresses the event accordingly.

## How to test

Added unit tests

## Note
This is dependent on an API function being created by Adams WIP pr:
https://github.com/racker/salus-telemetry-resource-management/blob/6c64efcd5f050eeff8dc7171530f024983b03c87/src/main/java/com/rackspace/salus/resource_management/web/controller/ResourceApi.java#L121

I wasn't able to connect to the api defined in that manner.  I had to change it like so:

https://github.com/racker/salus-telemetry-resource-management/compare/6c64efcd5f050eeff8dc7171530f024983b03c87...george/fixApi

In addition, the SQL doesn't seem to work yet.  I'll take a look at that next if I get a chance


src/main/java/com/rackspace/salus/monitor_management/config/MonitorManagementProperties.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.monitor_management.config;
+
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+@ConfigurationProperties("monitor-management")
```
itzg: In the public API module I'm thinking of using a config prefix of `services` and naming the properties class `ServicesProperties`. Then the deployment configs can consistently refer to `services.someBackendUrl`.

src/main/java/com/rackspace/salus/monitor_management/config/MonitorManagementProperties.java
```diff
@@ -0,0 +1,28 @@
...
+@Component
+@Data
+public class MonitorManagementProperties {
+    String resourceManagerUrl;
```
itzg: Nit: perhaps name this `resourceManagementUrl` to match the module name.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -145,20 +164,82 @@ public Monitor createMonitor(String tenantId, @Valid MonitorCreate newMonitor) t
...
+    /**
+     * Get a list of resources for the tenant that match the given labels
+     *
+     * @param tenantId tenant whose resources are to be found
+     * @param labels labels to be matched
+     * @return The list found
+     */
+    private List<Resource> getResourcesWithLabels(String tenantId, Map<String, String> labels) {
+        List<Resource> emptyList = new ArrayList<>();
+        String endpoint = monitorManagementProperties.getResourceManagerUrl() +
```
itzg: It feels like some URI templating could help you out here:

https://docs.spring.io/spring/docs/current/spring-framework-reference/integration.html#rest-resttemplate-uri

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -54,22 +61,34 @@
...
+                             EnvoyResourceManagement envoyResourceManagement,
+                             MonitorEventProducer monitorEventProducer,
+                             RestTemplateBuilder restTemplateBuilder,
+                             MonitorManagementProperties monitorManagementProperties) {
         this.monitorRepository = monitorRepository;
         this.entityManager = entityManager;
-	    this.monitorEventProducer = monitorEventProducer;
+        this.envoyResourceManagement = envoyResourceManagement;
+        this.monitorEventProducer = monitorEventProducer;
+        this.restTemplate = restTemplateBuilder.build();
```
itzg: I'm about to try this out in the public API code, but using

https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/web/client/RestTemplateBuilder.html#rootUri-java.lang.String-

to root the URLs on resource management would simplify the URL building later in the code.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -145,20 +164,82 @@ public Monitor createMonitor(String tenantId, @Valid MonitorCreate newMonitor) t
...
+        String endpoint = monitorManagementProperties.getResourceManagerUrl() +
+                "/api/tenant/tenantId/resourceLabels".replace("tenantId", tenantId);
+        try {
+            UriComponentsBuilder uriComponentsBuilder = UriComponentsBuilder.fromUriString(endpoint);
+            for (Map.Entry<String, String> e : labels.entrySet()) {
+                uriComponentsBuilder.queryParam(e.getKey(), e.getValue());
+            }
+            String uriString = uriComponentsBuilder.toUriString();
+            RequestEntity<Void> requestEntity = RequestEntity.get(new URI(uriString)).build();
+            ResponseEntity<List<Resource>> resp = restTemplate.exchange(requestEntity,
```
itzg: It sucks that Spring makes us use `exchange` when requesting a list of objects, but using this variation of `exchange` would at least be a little less awkward:

https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/client/RestTemplate.html#exchange-java.lang.String-org.springframework.http.HttpMethod-org.springframework.http.HttpEntity-org.springframework.core.ParameterizedTypeReference-java.util.Map-

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -187,66 +269,76 @@ public Monitor updateMonitor(String tenantId, UUID id, @Valid MonitorUpdate upda
...
+     */
     public void handleResourceEvent(ResourceEvent event) {
-        log.debug("");
-        if (event.getOldLabels() != null) {
-            Set<String> keys = event.getOldLabels().keySet();
-            keys.removeAll(event.getResource().getLabels().keySet());//now we should have the difference of labels.
+        Resource r = event.getResource();
+        String[] identifiers = r.getResourceId().split(":");
+        ResourceInfo resourceInfo = envoyResourceManagement
+                .getOne(r.getTenantId(), identifiers[0], identifiers[1]).join().get(0);
```
itzg: In a follow-up we should probably have `envoyResourceManagement.getOne(...)` take the opaque/raw resource ID rather than counting on the convention of a colon delimiter.

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorCreate.java
```diff
@@ -37,11 +40,11 @@
     @NotBlank
     String content;
 
-    @NotBlank
-    String agentType;
+    @NotNull
+    AgentType agentType;
```
itzg: Thanks for the enum change.

src/main/resources/application.yml
```diff
@@ -1,3 +1,6 @@
+monitor-management:
+  resourceManagerUrl:
+    http://localhost:8085
```
itzg: This localhost value should go into `application-dev.yml`.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
GeorgeJahad: I ended up needed to fix this here, (because it was already broken by the  incomplete resourceId change):
https://github.com/racker/salus-telemetry-etcd-adapter/pull/24

GeorgeJahad:
@itzg I think I've  satisfied all your requests, they were all excellent!  Thank you!

PR #7, opened 2019-03-05
Use resource mgmt query for resource IDs given Envoy labels

itzg:
## Resolves

For end to end integration of https://jira.rax.io/browse/SALUS-205

## What

Narrows the Envoy computation to use the resource management query focused on providing just the resource IDs for the given Envoy advertised labels. I originally added the query to temporarily simplify/avoid the object deserialization of a `Resource` via the JDBC query; however, I later realized that monitors should really only use the Envoy advertised labels, which the resource management app takes care of namespacing.

Also for integration of the public API, I added an attached Maven artifact with the `client` qualifier. That will expose just the REST model classes for use by the public API application.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -118,6 +118,23 @@
         <groupId>com.google.cloud.tools</groupId>
         <artifactId>jib-maven-plugin</artifactId>
       </plugin>
+
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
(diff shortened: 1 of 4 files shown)


itzg:
Thanks for the unit test fix. I'll merge this after https://github.com/racker/salus-telemetry-resource-management/pull/10

PR #8, opened 2019-03-06
Create MonitorManagement label matching query

harageth:
# Resolves

This is for https://jira.rax.io/browse/SALUS-125

# What

Create a query that will match to all of the labels we have provided

# How

It creates a query function to manually create the SQL query since this is a fairly complex query

# How to test

Run the MonitorManagementTest unit tests

# Why

This query incurs fewer JOIN penalties from the suggested SO approach.

# TODO

Probably a little cleanup

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -341,19 +341,7 @@ public class MonitorManagement {
      * @param tenantId The tenant associated to the resource
      * @return the list of Monitor's that match the labels
      */
-    public List<Monitor> getMonitorsFromLabels(Map<String, String> labels, String tenantId) throws IllegalArgumentException{
-        /*
-        SELECT * FROM resources where id IN (SELECT id from resource_labels WHERE id IN (select id from resources)
-        AND ((labels = "windows" AND labels_key = "os") OR (labels = "prod" AND labels_key="env")) GROUP BY id
-        HAVING COUNT(id) = 2) AND tenant_id = "aaaad";
-        */
-
-
-        /*
-        two scenarios for testing are envoy comes first or monitor comes first
-
-         */
-
+    public List<Monitor> getMonitorsFromLabels(Map<String, String> labels, String tenantId) throws IllegalArgumentException {
         if(labels.size() == 0) {
             throw new IllegalArgumentException("Labels must be provided for search");
         }
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -307,12 +307,12 @@ public class MonitorManagement {
         }
 
         List<Monitor> oldMonitors = null;
-        List<Monitor> monitors = getMonitorsFromLabels(event.getResource().getLabels(), event.getResource().getTenantId(), MatchOptions.MATCH_ALL);
+        List<Monitor> monitors = getMonitorsFromLabels(event.getResource().getLabels(), event.getResource().getTenantId());
         if (monitors == null) {
             monitors = new ArrayList<>();
         }
         if (event.getOldLabels() != null && !event.getOldLabels().equals(event.getResource().getLabels())) {
-            oldMonitors = getMonitorsFromLabels(event.getOldLabels(), event.getResource().getTenantId(), MatchOptions.MATCH_ALL);
+            oldMonitors = getMonitorsFromLabels(event.getOldLabels(), event.getResource().getTenantId());
         }
         if (oldMonitors != null) {
             monitors.addAll(oldMonitors);
@@ -341,7 +341,7 @@ public class MonitorManagement {
      * @param tenantId The tenant associated to the resource
      * @return the list of Monitor's that match the labels
      */
-    public List<Monitor> getMonitorsFromLabels(Map<String, String> labels, String tenantId, MatchOptions option) {
+    public List<Monitor> getMonitorsFromLabels(Map<String, String> labels, String tenantId) throws IllegalArgumentException{
         /*
         SELECT * FROM resources where id IN (SELECT id from resource_labels WHERE id IN (select id from resources)
         AND ((labels = "windows" AND labels_key = "os") OR (labels = "prod" AND labels_key="env")) GROUP BY id
@@ -353,6 +353,10 @@ public class MonitorManagement {
         two scenarios for testing are envoy comes first or monitor comes first
 
          */
+
+        if(labels.size() == 0) {
+            throw new IllegalArgumentException("Labels must be provided for search");
+        }
 
         MapSqlParameterSource paramSource = new MapSqlParameterSource();
         paramSource.addValue("tenantId", tenantId);
@@ -409,9 +413,5 @@ public class MonitorManagement {
         });
 
         return monitors;
-    }
-
-    public static enum MatchOptions {
-        MATCH_SOME, MATCH_ALL;
     }
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -346,7 +346,7 @@ public class MonitorManagement {
      * @param tenantId The tenant associated to the resource
      * @return the list of Monitor's that match the labels
      */
-    public List<Monitor> getMonitorsFromLabels(Map<String, String> labels, String tenantId, MATCH_OPTIONS option) {
+    public List<Monitor> getMonitorsFromLabels(Map<String, String> labels, String tenantId, MatchOptions option) {
         /*
         SELECT * FROM resources where id IN (SELECT id from resource_labels WHERE id IN (select id from resources)
         AND ((labels = "windows" AND labels_key = "os") OR (labels = "prod" AND labels_key="env")) GROUP BY id
@@ -417,7 +417,7 @@ public class MonitorManagement {
         return monitors;
     }
 
-    public static enum MATCH_OPTIONS {
+    public static enum MatchOptions {
         MATCH_SOME, MATCH_ALL;
     }
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -24,11 +24,7 @@ import com.rackspace.salus.telemetry.etcd.services.EnvoyResourceManagement;
 import com.rackspace.salus.telemetry.messaging.MonitorEvent;
 import com.rackspace.salus.telemetry.messaging.OperationType;
 import com.rackspace.salus.telemetry.messaging.ResourceEvent;
-import com.rackspace.salus.telemetry.model.Monitor;
-import com.rackspace.salus.telemetry.model.Monitor_;
-import com.rackspace.salus.telemetry.model.NotFoundException;
-import com.rackspace.salus.telemetry.model.Resource;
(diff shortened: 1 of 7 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApi.java
```diff
@@ -114,4 +117,11 @@ public void delete(@PathVariable String tenantId,
                        @PathVariable UUID uuid) {
         monitorManagement.removeMonitor(tenantId, uuid);
     }
+
+    @GetMapping("/tenant/{tenantId}/monitorLabels")
+    public List<Resource> getResourcesWithLabels(@PathVariable String tenantId,
```
GeorgeJahad: List<Monitor>?
getMonitorsWithLabels()?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -327,4 +333,76 @@ public void handleResourceEvent(ResourceEvent event) {
...
+
+        NamedParameterJdbcTemplate namedParameterTemplate = new NamedParameterJdbcTemplate(jdbcTemplate.getDataSource());
+        List<Monitor> monitors = new ArrayList<>();
+        namedParameterTemplate.query(builder.toString(), paramSource, (resultSet)->{
+
+            long prevId = 0;
+            Monitor prevMonitor = null;
+            boolean iterate = true;
+
+            while(iterate){
```
GeorgeJahad: while(resultSet.next()) ?

itzg: I agree with George especially since the docs say

> Initially, the cursor is positioned before the first row.

from https://docs.oracle.com/javase/tutorial/jdbc/basics/retrieving.html

itzg:
I'm not sure if exactly the same query will work for monitor lookups. For example, an Envoy could attach with three labels:

Label | Value
---|----------
os | DARWIN
arch | X86_64
cluster | staging

where matching monitors could have, for example

Label | Value
-------|------
cluster | staging

or 

Label | Value
-------|------
os | DARWIN
arch | X86_64

In this case, `getMonitorsFromLabels` will get called with the three labels of the Envoy and matching monitors will be those such that all labels of the monitor are included in the Envoy/query labels. This differs from the resource manager's `getResourcesFromLabels` where matching resources must include all of the labels given to the query.

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApi.java
harageth: Oops.... I thought I had converted everything over

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
harageth: That is interesting... I found while I was working with it that I could access the first row without doing anything and if I immediately did `resultSet.next()` that it iterated past the first row. 

I did try it as a `do while` loop but I was just realizing that I messed up the syntax and made it iterate over itself. Swapping to that syntax would make this easier to read.

harageth:
@itzg wouldn't we want to have both available? I think I left a comment in my other PR that altering the query to MATCH SOME instead of MATCH ALL is super simple. It is just the removal of the COUNT at the end of the inner query.

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApi.java
GeorgeJahad: That comment was supposed to read:
```
List<Monitor>?
```
I always forget to escape the
```
<>
```

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -399,10 +405,13 @@ public void handleResourceEvent(ResourceEvent event) {
...
                 }
-                iterate = resultSet.next();
-            }
+            } while(resultSet.next());
         });
 
         return monitors;
     }
+
+    public static enum MATCH_OPTIONS {
```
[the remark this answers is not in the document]
harageth: oh right. I was tired last night and not thinking

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -327,4 +333,85 @@ public void handleResourceEvent(ResourceEvent event) {
...
+            }
+            builder.append("(labels = :label"+ i +" AND labels_key = :labelKey" + i + ")");
+            paramSource.addValue("label"+i, entry.getValue());
+            paramSource.addValue("labelKey"+i, entry.getKey());
+            i++;
+        }
+
+        builder.append(" GROUP BY id");
+        //HAVING COUNT(id) = :i)
+        switch(option) {
```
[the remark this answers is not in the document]
harageth: I think we need to talk about this to actually understand what is needed (It sounds like there is some form of contradiction between your two comments). But you are correct, removing the `HAVING` will make this not work correctly.

harageth:
Hmmmm I am currently trying to figure out what happened to make the jdbcTemplate not get assigned anymore...

harageth:
Ok I am calling this one ready for full review

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApi.java
```diff
@@ -114,4 +117,11 @@ public void delete(@PathVariable String tenantId,
                        @PathVariable UUID uuid) {
         monitorManagement.removeMonitor(tenantId, uuid);
     }
+
+    @GetMapping("/tenant/{tenantId}/monitorLabels")
+    public List<Monitor> getResourcesWithLabels(@PathVariable String tenantId,
```
GeorgeJahad: getResources??

GeorgeJahad: why does getResourcesWithLabels() return a list of monitors?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -327,4 +334,84 @@ public void handleResourceEvent(ResourceEvent event) {
...
+
+    /**
+     * takes in a Mapping of labels for a tenant, builds and runs the query to match to those labels
+     * @param labels the labels that we need to match to
+     * @param tenantId The tenant associated to the resource
+     * @return the list of Monitor's that match the labels
+     */
+    public List<Monitor> getMonitorsFromLabels(Map<String, String> labels, String tenantId) throws IllegalArgumentException{
+        /*
+        SELECT * FROM resources where id IN (SELECT id from resource_labels WHERE id IN (select id from resources)
```
GeorgeJahad: is this comment correct?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -327,4 +334,84 @@ public void handleResourceEvent(ResourceEvent event) {
...
+                }else {
+                    Map<String, String> theseLabels = new HashMap<String, String>();
+                    theseLabels.put(
+                            resultSet.getString("labels_key"),
+                            resultSet.getString("labels"));
+                    Monitor m = new Monitor()
+                            .setId(UUID.fromString(resultSet.getString("id")))
+                            .setTenantId(resultSet.getString("tenant_id"))
+                            .setContent(resultSet.getString("content"))
+                            //.setMonitorName(resultSet.getString("monitor_name"))
```
GeorgeJahad: why are these commented out?

src/test/java/com/rackspace/salus/monitor_management/MonitorManagementTest.java
```diff
@@ -306,4 +312,160 @@ public void testPublishMonitor() {
...
+        String tenantId = RandomStringUtils.randomAlphanumeric(10);
+        String tenantId2 = RandomStringUtils.randomAlphanumeric(10);
+        monitorManagement.createMonitor(tenantId, create);
+        monitorManagement.createMonitor(tenantId2, create);
+        entityManager.flush();
+
+        List<Monitor> monitors = monitorManagement.getMonitorsFromLabels(labels, tenantId);
+        assertEquals(1, monitors.size()); //make sure we only returned the one value
+        assertEquals(tenantId, monitors.get(0).getTenantId());
+        //assertEquals(create.get(), resources.get(0).getResourceId());
```
GeorgeJahad: why is this commented out?

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApi.java
harageth: It was copied over and I apparently missed that name. Its fixed now

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
harageth: Oh right... I forgot I still need to finish decomposing the object. Since the monitor is storing enums and hibernate properly is storing integers and not enums in the database, at the time I started writing that I wanted to make sure the query worked and then I forgot that I hadn't finished creating the object

src/test/java/com/rackspace/salus/monitor_management/MonitorManagementTest.java
```diff
@@ -306,4 +312,160 @@ public void testPublishMonitor() {
...
+        labels.put("env", "test");
+
+        MonitorCreate create = podamFactory.manufacturePojo(MonitorCreate.class);
+        create.setLabels(resourceLabels);
+        String tenantId = RandomStringUtils.randomAlphanumeric(10);
+        monitorManagement.createMonitor(tenantId, create);
+        entityManager.flush();
+
+        List<Monitor> resources = monitorManagement.getMonitorsFromLabels(labels, tenantId);
+        assertEquals(1, resources.size()); //make sure we only returned the one value
```
itzg: This should not match, since the primary use of `getMonitorsFromLabels` is to query for those monitors that should be applicable for a given envoy/resource. As such, the returned monitors should only be those such that ALL of the monitor's labels are included in the queried set of labels.

src/test/java/com/rackspace/salus/monitor_management/MonitorManagementTest.java
```diff
@@ -306,4 +312,160 @@ public void testPublishMonitor() {
...
+
+
+        MonitorCreate create = podamFactory.manufacturePojo(MonitorCreate.class);
+        create.setLabels(monitorLabels);
+        String tenantId = RandomStringUtils.randomAlphanumeric(10);
+        monitorManagement.createMonitor(tenantId, create);
+        entityManager.flush();
+
+        List<Monitor> resources = monitorManagement.getMonitorsFromLabels(labels, tenantId);
+        assertEquals(1, resources.size()); //make sure we only returned the one value
```
itzg: Relative to my previous comment, this is indeed the positive matching scenario since all labels of the returned monitor are contained in the given query set of labels.

src/test/java/com/rackspace/salus/monitor_management/MonitorManagementTest.java
itzg: @GeorgeJahad or @jjbuchan , can you double check me that I'm not confusing the scenarios now?

GeorgeJahad: My understanding is that if the resources labels are a superset of the monitors, then it is a match.  So I think @itzg is correct here, in this example, the resources labels, (in the variable "labels",) are not a superset of the monitors labels, (in the variable "resourceLabels"), so this should not be a match.

PR #12, opened 2019-04-06
add zone to api

GeorgeJahad:
# Resolves
part of https://jira.rax.io/browse/SALUS-290

# What
adds zones to the api.  also fixed an indexing bug

## How to test
unit tests

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -174,7 +174,8 @@ public class MonitorManagement {
                 .setContent(newMonitor.getContent())
                 .setAgentType(newMonitor.getAgentType())
                 .setSelectorScope(newMonitor.getSelectorScope())
-                .setTargetTenant(newMonitor.getTargetTenant());
+                .setTargetTenant(newMonitor.getTargetTenant())
+                .setZones(newMonitor.getZones());
 
 
         monitorRepository.save(monitor);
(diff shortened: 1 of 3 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -226,13 +227,13 @@ public void publishMonitor(Monitor monitor, OperationType operationType, Map<Str
         final Set<String> deduped = new HashSet<>(resources);
 
         for (String resourceId : deduped) {
-            ResourceInfo resourceInfo = envoyResourceManagement
-                    .getOne(monitor.getTenantId(), resourceId).join().get(0);
-            if (resourceInfo != null) {
+            List<ResourceInfo> list = envoyResourceManagement
+                    .getOne(monitor.getTenantId(), resourceId).join();
```
itzg: I'm making changes already in etcd-adapter -- I think I'll change this `getOne` to return a single-value and not a list.

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorCreate.java
```diff
@@ -47,4 +47,5 @@
 
     ConfigSelectorScope selectorScope = ConfigSelectorScope.ALL_OF;
 
+    String[] zones;
```
itzg: If you make this change

https://github.com/racker/salus-telemetry-model/pull/39#discussion_r272806136

then this will need to be a `List<String>` also for ease of use.

GeorgeJahad:
Latest commits change zone to list and add more label validation:
https://github.com/racker/salus-telemetry-monitor-management/compare/d7acbec...62c2883

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -177,12 +178,28 @@ public Monitor createMonitor(String tenantId, @Valid MonitorCreate newMonitor) t
                 .setTargetTenant(newMonitor.getTargetTenant())
                 .setZones(newMonitor.getZones());
 
+        if (newMonitor.getLabels() != null) {
+            checkLabels(newMonitor.getLabels());
```
itzg: Hmm, the labels of monitors are different since they are "selectors". We actually want them to be able to select both system labels and user labels of resources.

GeorgeJahad: dumb of me.  i'll fix it.  thanks

GeorgeJahad: ok i reverted the label validation commits.

PR #13, opened 2019-04-08
migrate data model objects and unit tests over from api project

harageth:
# Resolves

    Half of https://jira.rax.io/browse/SALUS-261

# What

    Accept the data proxied through the API and convert it here rather than in the API

# How

    Use the MoniterConversionService class to convert the input that we receive into the proper input and output objects

## How to test

    Start up the api and the monitor management service and send a request through the API to the Monitor Management Service
There is also the MonitorConversionServiceTest unit tests file

# Why

    Because this was the approach taken in the other API's

# TODO

    Make sure all of the objects are created correctly and that the services are interacting properly together


itzg:
Let me know if/when this is ready for final review.

itzg:
...oh, looks like you need to merge (not rebase) master into your branch. I'd suggest just accepting your changes; however, bringing the new `zones` field into your `MonitorCU`:

https://github.com/racker/salus-telemetry-monitor-management/commit/ebc44e88b913f03e92b1ea24a11818456e33f1c4#diff-81d62bcb3870f7d769054f0703ab5bb9R51

jjbuchan:
@itzg 
> looks like you need to merge (not rebase) master into your branch

I'd be curious to hear your thoughts on this.  I never merge master into my own branches and feel like it complicates things more when others do.

PR #14, opened 2019-04-12
Initial support for remote monitors and bound monitor management

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-305
All of https://jira.rax.io/browse/SALUS-297

# What

Adds the initial support for remote monitors specifically ONLY this scenario so far:

image

These changes introduce `BoundMonitor` and the management of those via the `ZoneStorage` added in https://github.com/racker/salus-telemetry-etcd-adapter/pull/27

The `MonitorConversionService` has been updated to convert the TODOs into initial support for remote plugins, starting with ping.

`MonitorApi` is now a refactored interface that introduces a concept I'd like to start using for these inter-service REST API calls. It allowed for introduce `MonitorApiClient`, which is a client-side component that abstracts the calls into this app. Its usage will be seen in the corresponding ambassador PR.

It also gets rid of the use of the one-off "get resource IDs" query that was removed in https://github.com/racker/salus-telemetry-resource-management/pull/17

## How to test

Unit tests were updated


itzg:
Source for sequence diagram is [here](https://www.websequencediagrams.com/?lz=dGl0bGUgRW52b3kgYXR0YWNoZXMsIHRoZW4gbmV3IG1vbml0b3IKCnBhcnRpY2lwYW50ACgGAAUNVXNlcgoKAEEFLT5BbWJhc3NhZG9yOgBMBwpVc2VyLT4rTQBIBk1hbmFnZW1lbnQ6IGNyZWF0ZSByZW1vdGUvYWdlbnQAbAkAHhEtPisAUQxzZW5kIGV2ZW50CgBqCi0-AIFIBTogY29uZmlnCgo&s=napkin)

src/main/java/com/rackspace/salus/monitor_management/entities/BoundMonitor.java
```diff
@@ -0,0 +1,90 @@
...
+
+    // NOTE the ordering of the following is important since potentially null values need to be
+    // last in the primary key
+
+    UUID monitorId;
+    String resourceId;
+    // agent monitors will have null zone and zoneTenantId
+    String zone;
+    // public zones will have null zoneTenantId
+    String zoneTenantId;
```
harageth: Do we want the TenantId as part of the primary key? The minimal candidate key is actually the monitorId, ResourceId, zone. I think we should index this tenantId but I am not convinced its necessary in the primary key.

itzg: For private zones we are allowing the customer/tenant to give the zone whatever identifier they want. It could be something simple like "us-west" which would likely collide with a private zone named by other tenants. So...we need to qualify private zone IDs by tenant except the special case of public zones where an empty string is used in place of tenant.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
```diff
@@ -14,19 +14,28 @@
...
+import com.rackspace.salus.monitor_management.web.model.RemoteMonitorDetails;
+import com.rackspace.salus.monitor_management.web.model.RemotePlugin;
 import com.rackspace.salus.telemetry.model.ConfigSelectorScope;
 import com.rackspace.salus.telemetry.model.Monitor;
 import java.io.IOException;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
+;
```
harageth: I havent checked this out... But I feel like this semi-colon shouldn't be there.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
```diff
@@ -65,56 +72,90 @@ public DetailedMonitorOutput convertToOutput(Monitor monitor) {
...
-        throw new IllegalStateException("Inconsistent AgentType");
+      try {
+        remotePlugin = objectMapper.readValue(monitor.getContent(), RemotePlugin.class);
+      } catch (IOException e) {
+        log.warn("Failed to deserialize RemotePlugin for monitor={}", monitor, e);
+        throw new IllegalStateException("Failed to deserialize RemotePlugin");
       }
 
-      monitorDetails.setPlugin(localPlugin);
+      verifyPluginAgentType(monitor, remotePlugin);
```
harageth: Hmmm I am not sure I like that this function is throwing the exception itself. I guess it doesn't really matter. The thing that stopped me on the function call is the fact that its not returning anything and just a function call standing alone.

harageth: I get it needs to do the same thing in both instances. So maybe a rename of the function because a verification function I feel needs to return a boolean.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
```diff
@@ -65,56 +72,90 @@ public DetailedMonitorOutput convertToOutput(Monitor monitor) {
...
+
+      monitorDetails.setPlugin(remotePlugin);
     }
 
     return detailedMonitorOutput;
   }
 
-  public MonitorCU convertFromInput(DetailedMonitorInput create) {
+  public MonitorCU convertFromInput(DetailedMonitorInput input) {
```
harageth: Good catch renaming this for me. Thanks

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -166,45 +189,176 @@ public Monitor getMonitor(String tenantId, UUID id) {
...
+            return new ResolvedZone()
+                .setPublicZone(false)
+                .setTenantId(tenantId)
+                .setId(zone);
+        }
+    }
+
+    private List<String> determineMonitoringZones(Monitor monitor) {
+        if (monitor.getZones() == null || monitor.getZones().isEmpty()) {
+            return zonesProperties.getDefaultZones();
```
harageth: reminder to myself to look at the this getDefaultZones code

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -166,45 +189,176 @@ public Monitor getMonitor(String tenantId, UUID id) {
...
+        ResponseEntity<List<Resource>> resp = restTemplate.exchange(uriString, HttpMethod.GET, null,
+                new ParameterizedTypeReference<List<Resource>>() {
                 });
-        if (resp.getStatusCode() != HttpStatus.OK) {
+        if (resp.getStatusCode() != HttpStatus.OK || resp.getBody() == null) {
             log.error("get failed on: " + uriString, resp.getStatusCode());
             return emptyList;
         }
-        return resp.getBody();
+        return resp.getBody().stream()
```
harageth: this does seem so much easier than actually maintaining the two different fields. If we need to speed this up at some point in the future for large customers (Comcast should end up with 4500 servers if they use us on all of their deployments across clouds) we can refactor the query builder to build out the query string and return that to a function that returns the proper Objects. Rather than building the query and running it in the same function.

src/main/java/com/rackspace/salus/monitor_management/entities/BoundMonitor.java
harageth: But doesn't the UUID of the monitor/resource give us that uniqueness? Or were we using this table to imply the zones that we have configured?

src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
```diff
@@ -0,0 +1,46 @@
...
+  };
+  private final RestTemplate restTemplate;
+
+  public MonitorApiClient(RestTemplate restTemplate) {
+    this.restTemplate = restTemplate;
+  }
+
+  @Override
+  public List<BoundMonitor> getBoundMonitors(String envoyId) {
+    return restTemplate.exchange(
```
harageth: What is happening here?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
itzg: Nope, I'll delete it

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
itzg: Yeah, there's too much bulk in the log and exception building to repeat it...which is why I refactored this out in the first place. What name/prefix would you suggest?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
itzg: Agreed. Either way having one main query with a few conditionals rather than the duplication I ended up with there was the main thing I wanted to avoid.

harageth: Yep. I mostly just wanted to make this comment for posterity sake. Make sure we are on the same page for future ideas.

src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
itzg: Yeah, after pushing the PR I realized I needed javadocs on this explain assumptions about the `RestClient` passed into this.

itzg:
Rather than create yet another PR adding to the pile, I went ahead and included the templated content rendering into this same PR with these commits

https://github.com/racker/salus-telemetry-monitor-management/pull/14/files/015fd37afa58c1ffb9c32701fcc186b563bd4bf6..cb487da04b055374ddf24af4894198f5a28b4371

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -166,43 +193,166 @@ public Monitor getMonitor(String tenantId, UUID id) {
...
+        log.debug("Distributing new monitor={} to resources={}", monitor, resources);
+
+        final List<BoundMonitor> boundMonitors = new ArrayList<>();
+
+        if (monitor.getSelectorScope() == ConfigSelectorScope.ALL_OF) {
+            // AGENT MONITOR
+
+            for (Resource resource : resources) {
+                final ResourceInfo resourceInfo = envoyResourceManagement
+                    .getOne(monitor.getTenantId(), resource.getResourceId())
```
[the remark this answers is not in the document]
itzg: I agree the mismatch in pattern seems odd at first. Anything lease related would not be able to be moved into a microservice like our SQL pattern, so then we'd be left with half the etcd code one way and half the other.

This is worth thinking about more especially after the agent catalog parts are moved to SQL.

src/main/java/com/rackspace/salus/monitor_management/entities/BoundMonitor.java
```diff
@@ -0,0 +1,92 @@
...
+import javax.validation.constraints.NotNull;
+import lombok.Data;
+import org.hibernate.annotations.Type;
+
+// Using the old validation exceptions for podam support
+// Will move to the newer ones once they're supported.
+//import javax.validation.constraints.NotBlank;
+
+@Entity
+@IdClass(BoundMonitor.PrimaryKey.class)
```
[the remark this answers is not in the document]
itzg: George had discovered this approach, but...I see `@Embedded`, but not `@EmbeddedKey`. You might be mixing up the JPA embedded object annotations.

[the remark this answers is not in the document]
jjbuchan: `@EmbeddedId` is what I meant... you'd specify that on the composite "embedded" object.

Definitely me mixing things up.

src/main/java/com/rackspace/salus/monitor_management/entities/BoundMonitor.java
```diff
@@ -0,0 +1,92 @@
...
+  @Column(length = 100)
+  String zone;
+
+  /**
+   * For remote monitors, this field must be non-empty to indicate to the assigned Envoy that
+   * the measurements are being collected on behalf of the target tenant rather than the tenant
+   * of the Envoy.
+   */
+  @NotNull
+  String targetTenant;
```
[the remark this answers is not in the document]
itzg: First `monitorId` is the `id` of the `Monitor` -- it's just not formally declared as a foreign key.

As for tenants,`BoundMonitor` has two tenants going on:
- the tenant that has a resource to be monitored
- the tenant owning the zone, if it's a remote, private monitor

The first, is the `targetTenant` and is set to the tenant of the `Monitor` for remote monitors [here](https://github.com/racker/salus-telemetry-monitor-management/blob/a49d872cb4ca98ef6f3deb89cf6d92d23960ee1c/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java#L305).

For local/agent monitors `targetTenant` is currently set to an empty string so that the ambassador can use that as condition to override the telegraf plugin tags [here](https://github.com/racker/salus-telemetry-ambassador/blob/f9314d2cc858bf12c526b103c7333f527d3d7d4b/src/main/java/com/rackspace/salus/telemetry/ambassador/services/ConfigInstructionsBuilder.java#L67). In hindsight, it was probably better to use a separate boolean field to indicate that logic should be activated or not and just always populate the monitor's tenant in `BoundMonitor`. With that inconsistency, local monitors can't be located by tenant right now without doing a join via the `monitorId`.

FYI, for the second items, the tenant owning the zone, that's a field I'm now adding back to the composite key since my current jira story requires looking up bound monitors only given the zone. For private zones, the zone ID alone is not enough since they are not required to be unique across tenants.

Yes, I also think we should eventually move `Monitor` into the same entities package as `BoundMonitor`. I'm wanting us to wait to make "larger" changes like that until a majority of the remote polling retrofitting is done.

PR #15, opened 2019-04-17
Provide a client facing DTO version of BoundMonitor entity

itzg:
# Resolves

Fixes a bug introduced in https://github.com/racker/salus-telemetry-monitor-management/pull/14

# What

Clients of the `MonitorApi` interface could not compile since it referenced `BoundMonitor`, which is purposely not included in the client jar.

# How

Introduces a Data Transfer Object (DTO) version of `BoundMonitor` to decouple the REST API clients from the "private" JPA entity. This approach deviates from [what we currently do in model](https://github.com/racker/salus-telemetry-model/tree/master/src/main/java/com/rackspace/salus/telemetry/model) since it is an experiment to see if this is a better and feasible coding practice.

## How to test

A test was added to confirm the likeness of the DTO.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
@@ -16,7 +16,7 @@
 
 package com.rackspace.salus.monitor_management.web.client;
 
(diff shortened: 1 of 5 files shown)


itzg:
@harageth just as a reminder, this PR is required for https://github.com/racker/salus-telemetry-ambassador/pull/32

PR #16, opened 2019-04-18
Rename the labels field to labelSelector in the monitor DTO

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-319

# What

During sprint demo we realized it would be better to rename this user facing field to "labelSelector" emphasize that it selects the labels of the resource and is not itself a set of labels of the monitor.

## How to test

Existing unit tests

# Depends on

- https://github.com/racker/salus-telemetry-model/pull/42

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
@@ -52,7 +52,7 @@ public class MonitorConversionService {
     final DetailedMonitorOutput detailedMonitorOutput = new DetailedMonitorOutput()
         .setId(monitor.getId().toString())
         .setName(monitor.getMonitorName())
-        .setLabels(monitor.getLabels());
+        .setLabelSelector(monitor.getLabels());
 
     final ConfigSelectorScope selectorScope = monitor.getSelectorScope();
 
@@ -99,7 +99,7 @@ public class MonitorConversionService {
   public MonitorCU convertFromInput(DetailedMonitorInput input) {
     final MonitorCU monitor = new MonitorCU()
         .setMonitorName(input.getName())
-        .setLabels(input.getLabels());
+        .setLabels(input.getLabelSelector());
 
     final MonitorDetails details = input.getDetails();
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
```diff
@@ -99,7 +99,7 @@ public DetailedMonitorOutput convertToOutput(Monitor monitor) {
   public MonitorCU convertFromInput(DetailedMonitorInput input) {
     final MonitorCU monitor = new MonitorCU()
         .setMonitorName(input.getName())
-        .setLabels(input.getLabels());
+        .setLabels(input.getLabelSelector());
```
jjbuchan: Should it change in `MonitorCU` as well?

harageth: Yeah it probably should

itzg: Yeah, for consistency probably should...but mainly I was wanting to change the user facing class. If I change `MonitorCU`...should probably change `Monitor` also to be complete. Seem ok?

harageth: Yeah I am good with that

jjbuchan: Me too.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -520,37 +519,15 @@ public void handleResourceEvent(ResourceEvent event) {
...
-                            .setLabels(theseLabels);
-                    prevId = resultSet.getString("id");
-                    prevMonitor = m;
-                    monitors.add(m);
-
-                }
-            } while(resultSet.next());
-        });
-
+        final List<UUID> monitorIds = namedParameterTemplate.query(builder.toString(), paramSource,
```
itzg: Since the column names changed, I took the opportunity to consistently implement this like I did over here

https://github.com/racker/salus-telemetry-resource-management/pull/17/files#r275867928

PR #17, opened 2019-04-25
Support new zone events and modified resource event structure

itzg:
# Resolves

Part of
- https://jira.rax.io/browse/SALUS-263
- https://jira.rax.io/browse/SALUS-311
- https://jira.rax.io/browse/SALUS-316

# What

Referring to the diagram sections starting at 

https://one.rackspace.com/pages/viewpage.action?pageId=577472836#ExplicitPollerPickingDesign-EntityDiagrams 

this PR supports processing of the new zone events introduced in https://github.com/racker/salus-telemetry-zone-management/pull/1 by handling the cases when a new envoy-resource registers in a zone and when a new envoy-resource instance re-registers for the same resource in the expected zones.

It also adjusts the resource event handling code to adapt to the new event structure from https://github.com/racker/salus-telemetry-model/pull/40 by computing corresponding monitoring binding creations, change, and removal based on resource changes.

## How to test

Lots of new unit test cases were added for the new scenarios.

# Depends on
- https://github.com/racker/salus-telemetry-model/pull/43
- https://github.com/racker/salus-common/pull/15

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/repositories/BoundMonitorRepository.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/repositories/BoundMonitorRepository.java
@@ -25,6 +25,9 @@ public interface BoundMonitorRepository extends CrudRepository<BoundMonitor, Bou
 
   List<BoundMonitor> findByEnvoyId(String envoyId);
 
-  @Query("select b from BoundMonitor b where b.zoneTenantId = :zoneTenantId and b.zone = :zone and envoyId is null")
-  List<BoundMonitor> findOnesWithoutEnvoy(String zoneTenantId, String zone);
+  @Query("select b from BoundMonitor b where b.zoneTenantId = :zoneTenantId and b.zoneId = :zoneId and b.envoyId is null")
+  List<BoundMonitor> findOnesWithoutEnvoy(String zoneTenantId, String zoneId);
+
+  @Query("select b from BoundMonitor b where b.zoneTenantId = :zoneTenantId and b.zoneId = :zoneId and b.envoyId = :envoyId")
+  List<BoundMonitor> findOnesWithEnvoy(String zoneTenantId, String zoneId, String envoyId);
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/repositories/BoundMonitorRepository.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/repositories/BoundMonitorRepository.java
@@ -18,6 +18,7 @@ package com.rackspace.salus.monitor_management.repositories;
 
 import com.rackspace.salus.monitor_management.entities.BoundMonitor;
 import java.util.List;
+import java.util.UUID;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
 
@@ -30,4 +31,6 @@ public interface BoundMonitorRepository extends CrudRepository<BoundMonitor, Bou
 
   @Query("select b from BoundMonitor b where b.zoneTenantId = :zoneTenantId and b.zoneId = :zoneId and b.envoyId = :envoyId")
   List<BoundMonitor> findOnesWithEnvoy(String zoneTenantId, String zoneId, String envoyId);
+
+  List<BoundMonitor> findByMonitor_IdAndResourceId(UUID monitorId, String resourceId);
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/repositories/BoundMonitorRepository.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/repositories/BoundMonitorRepository.java
@@ -24,13 +24,13 @@ import org.springframework.data.repository.CrudRepository;
 
 public interface BoundMonitorRepository extends CrudRepository<BoundMonitor, BoundMonitor.PrimaryKey> {
 
-  List<BoundMonitor> findByEnvoyId(String envoyId);
+  List<BoundMonitor> findAllByEnvoyId(String envoyId);
 
   @Query("select b from BoundMonitor b where b.zoneTenantId = :zoneTenantId and b.zoneId = :zoneId and b.envoyId is null")
-  List<BoundMonitor> findOnesWithoutEnvoy(String zoneTenantId, String zoneId);
+  List<BoundMonitor> findAllWithoutEnvoy(String zoneTenantId, String zoneId);
 
   @Query("select b from BoundMonitor b where b.zoneTenantId = :zoneTenantId and b.zoneId = :zoneId and b.envoyId = :envoyId")
-  List<BoundMonitor> findOnesWithEnvoy(String zoneTenantId, String zoneId, String envoyId);
+  List<BoundMonitor> findAllWithEnvoy(String zoneTenantId, String zoneId, String envoyId);
 
-  List<BoundMonitor> findByMonitor_IdAndResourceId(UUID monitorId, String resourceId);
+  List<BoundMonitor> findAllByMonitor_IdAndResourceId(UUID monitorId, String resourceId);
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/repositories/BoundMonitorRepository.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/repositories/BoundMonitorRepository.java
@@ -18,9 +18,13 @@ package com.rackspace.salus.monitor_management.repositories;
 
 import com.rackspace.salus.monitor_management.entities.BoundMonitor;
 import java.util.List;
+import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
 
 public interface BoundMonitorRepository extends CrudRepository<BoundMonitor, BoundMonitor.PrimaryKey> {
 
   List<BoundMonitor> findByEnvoyId(String envoyId);
+
+  @Query("select b from BoundMonitor b where b.zoneTenantId = :zoneTenantId and b.zone = :zone and envoyId is null")
+  List<BoundMonitor> findOnesWithoutEnvoy(String zoneTenantId, String zone);
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -16,12 +16,12 @@
 
 package com.rackspace.salus.monitor_management.services;
 
-import com.rackspace.salus.monitor_management.config.ServicesProperties;
 import com.rackspace.salus.monitor_management.config.ZonesProperties;
 import com.rackspace.salus.monitor_management.entities.BoundMonitor;
 import com.rackspace.salus.monitor_management.repositories.BoundMonitorRepository;
 import com.rackspace.salus.monitor_management.repositories.MonitorRepository;
 import com.rackspace.salus.monitor_management.web.model.MonitorCU;
+import com.rackspace.salus.resource_management.web.client.ResourceApi;
 import com.rackspace.salus.telemetry.errors.AlreadyExistsException;
 import com.rackspace.salus.telemetry.etcd.services.EnvoyResourceManagement;
 import com.rackspace.salus.telemetry.etcd.services.ZoneStorage;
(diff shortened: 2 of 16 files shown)


src/main/java/com/rackspace/salus/monitor_management/repositories/BoundMonitorRepository.java
```diff
@@ -18,9 +18,19 @@
...
 import java.util.List;
+import java.util.UUID;
+import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
 
 public interface BoundMonitorRepository extends CrudRepository<BoundMonitor, BoundMonitor.PrimaryKey> {
 
   List<BoundMonitor> findByEnvoyId(String envoyId);
+
+  @Query("select b from BoundMonitor b where b.zoneTenantId = :zoneTenantId and b.zoneId = :zoneId and b.envoyId is null")
```
harageth: Glad you did them in the same order as the index.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -578,15 +578,16 @@ void handleResourceEvent(ResourceEvent event) {
 
         for (Monitor monitor : monitors) {
             final List<BoundMonitor> existing = boundMonitorRepository
-                .findByMonitor_IdAndResourceId(monitor.getId(), resourceInfo.getResourceId());
+                .findByMonitor_IdAndResourceId(monitor.getId(), resource.getResourceId());
```
harageth: It wont let me leave a comment on line 573 since its not modified but since we aren't using the resourceInfo in this function can we remove that too?

itzg: It's still used here

https://github.com/racker/salus-telemetry-monitor-management/pull/17/files/9fc6c0f6eb1c9641e3bc6ef934078f92a797bf6f..585c8709bb24ffcc8079f3f3283964987df4d6ca#diff-9f3caf1ed599810e6ed74b7a5556cd14R590

harageth: I apparently glossed over that portion

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -322,14 +324,16 @@ private static String emptyStringForNull(String input) {
...
-    public void handleNewEnvoyInZone(String zoneTenantId, String zoneId) {
+    public void handleNewEnvoyInZone(@Nullable String zoneTenantId, String zoneId) {
         log.debug("Locating bound monitors without assigned envoy with zoneId={} and zoneTenantId={}",
             zoneId, zoneTenantId);
 
         final ResolvedZone resolvedZone = resolveZone(zoneTenantId, zoneId);
 
         final List<BoundMonitor> onesWithoutEnvoy = boundMonitorRepository
-            .findOnesWithoutEnvoy(zoneTenantId, zoneId);
+            .findOnesWithoutEnvoy(emptyStringForNull(zoneTenantId),  zoneId);
```
harageth: In the case of a null->emptyString conversion here. What does that exactly mean? There is no tenantId owning the zone? Which implies one of our own zones?

itzg: Yes, a null zone tenant ID means it's one of our public zones. ResolvedZone encapsulates most of that logic:

https://github.com/racker/salus-telemetry-etcd-adapter/blob/master/src/main/java/com/rackspace/salus/telemetry/etcd/types/ResolvedZone.java#L23

harageth: Ahhh and since the getTenantId() function returns null or the tenantId

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -735,15 +695,698 @@ public void testDistributeNewMonitor_remote_emptyZone() {
...
+
+    @Test
+    public void testUpsertBindingToResource() {
+
+        final UUID m0 = UUID.fromString("00000000-0000-0000-0000-000000000000");
+        final UUID m1 = UUID.fromString("00000000-0000-0000-0001-000000000000");
+        final UUID m2 = UUID.fromString("00000000-0000-0000-0002-000000000000");
+        final UUID m3 = UUID.fromString("00000000-0000-0000-0003-000000000000");
+
+        // The following monitors hits the various scenarios we need to support when a
```
jjbuchan: when a.....

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -735,15 +695,698 @@ public void testDistributeNewMonitor_remote_emptyZone() {
...
+        List<Monitor> monitors = Arrays.asList(
+            // new local monitor
+            // --> 1 x BoundMonitor
+            new Monitor()
+            .setId(m0)
+            .setTenantId("t-1")
+            .setSelectorScope(ConfigSelectorScope.ALL_OF)
+            .setContent("new local domain=<<resource.labels.env>>"),
```
jjbuchan: I'm still not a huge fan of the `<< >>`.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -735,15 +695,698 @@ public void testDistributeNewMonitor_remote_emptyZone() {
...
+                .setZoneTenantId("t-1")
+                .setZoneId("z-2"),
+            new BoundMonitor()
+                .setMonitor(monitors.get(2))
+                .setResourceId("r-1")
+                .setEnvoyId("e-3")
+                .setRenderedContent("existing local domain=prod")
+                .setZoneTenantId("")
+                .setZoneId("")
+        ));
```
jjbuchan: Shouldn't there be an entry for the `"static content"` monitor?

src/main/java/com/rackspace/salus/monitor_management/repositories/BoundMonitorRepository.java
```diff
@@ -18,9 +18,19 @@
...
+  List<BoundMonitor> findOnesWithoutEnvoy(String zoneTenantId, String zoneId);
```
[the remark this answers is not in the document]
itzg: Good point. I'll change it, since Spring Data repositories use a convention of "findAll" prefixes.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -272,13 +281,12 @@ private void sendMonitorBoundEvents(OperationType operationType,
...
-            .setMonitorId(monitor.getId())
+            .setMonitor(monitor)
             .setResourceId(resource.getResourceId())
             .setEnvoyId(envoyId)
-            .setAgentType(monitor.getAgentType())
             .setRenderedContent(renderMonitorContent(monitor, resource))
-            .setTargetTenant("")
-            .setZone("");
+            .setZoneTenantId("")
+            .setZoneId("");
```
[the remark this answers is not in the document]
itzg: They have to be non-null or the SQL insertion will fail due to being part of primary key, etc.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -297,31 +305,100 @@ private BoundMonitor bindRemoteMonitor(Monitor monitor, Resource resource, Strin
...
+        final List<BoundMonitor> boundToPrev = boundMonitorRepository.findOnesWithEnvoy(
+            emptyStringForNull(tenantId),
+            zoneId,
+            fromEnvoyId
+        );
+
+        if (!boundToPrev.isEmpty()) {
+            log.debug("Re-assigning bound monitors={} to envoy={}", boundToPrev, toEnvoyId);
+            for (BoundMonitor boundMonitor : boundToPrev) {
+                boundMonitor.setEnvoyId(toEnvoyId);
```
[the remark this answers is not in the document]
itzg: My beliefs are inconsistent since I like Java 8 stream/lambda semantics, but with `forEach` I don't like paying the cost of lambda and scope capturing when a good old for-each construct works fine.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -454,36 +508,138 @@ public void removeMonitor(String tenantId, UUID id) {
...
+            selectedMonitors = Collections.emptyList();
+            // ...and monitorIdsToUnbind remains ALL of the currently bound
         }
-        if (event.getOldLabels() != null && !event.getOldLabels().equals(event.getResource().getLabels())) {
-            oldMonitors = getMonitorsFromLabels(event.getOldLabels(), event.getResource().getTenantId());
+
+        List<BoundMonitor> unbound = unbindByMonitorId(monitorIdsToUnbind);
+
+        final List<Monitor> monitorsToUpsert = selectedMonitors.stream()
+            .filter(monitor -> !monitorIdsToUnbind.contains(monitor.getId()))
```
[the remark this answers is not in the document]
itzg: Oh yeah, I think you're probably right. I'll change and see what unit tests reveal.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
itzg: it's a choose your own adventure :)

I'll finish the comment, but basically it was supposed to be "when a resource changes"

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
itzg: Me either. We are free to use any one or two letter delimiters other than `{{ ... }}`, which is what I really wanted to use.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
itzg: No. I'll add this inline comment in the code to note the purposeful absence:
```
            // NOTE binding of m3 did not need to be re-bound since its "static content" was
            // unaffected by the change in resource labels.
```

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
jjbuchan: Could we do `${...}` that's what ele alarms use.  Something involving `{...}` is definitely my preference.

jjbuchan: They might have been `#{...}` but either way, something like that would be nice.

itzg: I'll try that out. So far it looks like Insomnia doesn't get weird about `${...}`, but not sure if jmustache allows imbalanced delimiters. I'll find out...

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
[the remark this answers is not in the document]
itzg: Yep, that extra variable was redundant and helped make this complex method slightly simpler.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
itzg: Cool, `${...}` seems to work via unit tests. I'm just a little nervous since the javadoc don't offer the choice of "AB D":

http://samskivert.github.io/jmustache/apidocs/com/samskivert/mustache/Mustache.Compiler.html#withDelims-java.lang.String-

itzg: ...but seems to be fine since start1, start2 are maintained separately from end1,end2:

https://github.com/samskivert/jmustache/blob/master/src/main/java/com/samskivert/mustache/Mustache.java#L603-L633

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
jjbuchan: I realized my mistake.  I had missed `when(boundMonitorRepository.findAllByMonitor_IdAndResourceId(m3, "r-1"))` which means the test will see that this was already an existing bound monitor.  I was thinking there should be an event because it hadn't been created before.

PR #18, opened 2019-04-29
Adam9500/salus 330 add swagger documentation

harageth:
# Resolves

    SALUS-330

# What

    Adds internal and external documentation for Monitor-management

# How

    Generates internal documentation with Swagger then utilizes the SwaggerJsonConverter to export the internal documentation for external use

## How to test

    After Installing (Running the mvn install target) the SwaggerJsonConverter will be in the .m2 folder and then you can run the `test` target here that will generate the internal swagger and run the SwaggerJsonConverter

# TODO

    This should be complete

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -124,7 +124,12 @@
       <artifactId>swagger-hibernate-validations</artifactId>
       <version>1.5.6</version>
     </dependency>
-
+    <dependency>
+      <groupId>com.rackspace.salus</groupId>
+      <artifactId>salus-tools</artifactId>
+      <version>0.0.1-SNAPSHOT</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
@@ -165,8 +170,8 @@
             <apiSource>
               <springmvc>true</springmvc>
               <locations>
-                <location>com.rackspace.salus.resource_management.web.model</location>
-                <location>com.rackspace.salus.resource_management.web.controller.ResourceApi</location>
+                <location>com.rackspace.salus.monitor_management.web.model</location>
+                <location>com.rackspace.salus.monitor_management.web.controller.MonitorApiController</location>
               </locations>
               <templatePath>${basedir}/templates/strapdown.html.hbs</templatePath>
               <schemes>
--- a/pom.xml
+++ b/pom.xml
@@ -200,6 +200,28 @@
           </dependency>
         </dependencies>
       </plugin>
+
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>exec-maven-plugin</artifactId>
+        <version>1.2.1</version>
+
+        <executions>
+          <execution>
+            <phase>test-compile</phase>
+            <goals>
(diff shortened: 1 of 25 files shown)


pom.xml
```diff
@@ -108,6 +108,11 @@
       <version>7.2.1.RELEASE</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.swagger</groupId>
+      <artifactId>swagger-hibernate-validations</artifactId>
```
itzg: This only needs `swagger-core`

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
```diff
@@ -109,6 +120,7 @@ public SseEmitter getAllAsStream() {
     }
 
     @GetMapping("/tenant/{tenantId}/monitors/{uuid}")
+    @ApiOperation(value = "Gets specific Monitor for Tenant")
     public DetailedMonitorOutput getById(@PathVariable String tenantId,
```
itzg: How did swagger do with the heavily polymorphic types within `DetailedMonitorOutput`/`Input`?

pom.xml
```diff
@@ -139,6 +144,72 @@
...
+            <apiSource>
+              <springmvc>true</springmvc>
+              <locations>
+                <location>com.rackspace.salus.resource_management.web.model</location>
+                <location>com.rackspace.salus.resource_management.web.controller.ResourceApi</location>
+              </locations>
+              <templatePath>${basedir}/templates/strapdown.html.hbs</templatePath>
+              <schemes>
+                <scheme> https </scheme>
+                <scheme> http </scheme>
```
jjbuchan: Is there supposed to be white-space in these?

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
```diff
@@ -130,6 +143,8 @@ public DetailedMonitorOutput getById(@PathVariable String tenantId,
 
     @PostMapping("/tenant/{tenantId}/monitors")
     @ResponseStatus(HttpStatus.CREATED)
+    @ApiOperation(value = "Creates new Monitor for Tenant")
+    @ApiResponses(value = { @ApiResponse(code = 201, message = "Successfully Created Monitor")})
```
jjbuchan: Do the others default to a 200 code?

jjbuchan: Would be nice if this just used the `ResponseStatus` info so you didn't have to duplicate things.

harageth: It would be. We decided to just place the ApiResponses in locations where the traditional response code isn't expected.

itzg:
Next time git gives you an error that leads to this kind of branch-branch merge can you post that into Slack?

image

There's something about your workflow that's confusing git.

PR #20, opened 2019-05-01
Support tenant query of bound monitors

itzg:
# What

Add a REST API for a tenant to query their bound monitors. Later we can add query parameters to support further narrowing by resource and/or monitor ID.

## How to test

Added repository test.

# TODO

I'll update the public API to add a request proxy for the new API.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/repositories/BoundMonitorRepository.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/repositories/BoundMonitorRepository.java
@@ -19,6 +19,8 @@ package com.rackspace.salus.monitor_management.repositories;
 import com.rackspace.salus.monitor_management.entities.BoundMonitor;
 import java.util.List;
 import java.util.UUID;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
 import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.CrudRepository;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/rackspace/salus/monitor_management/repositories/BoundMonitorRepository.java
```diff
@@ -33,4 +35,6 @@
   List<BoundMonitor> findAllWithEnvoy(String zoneTenantId, String zoneId, String envoyId);
 
   List<BoundMonitor> findAllByMonitor_IdAndResourceId(UUID monitorId, String resourceId);
+
+  Page<BoundMonitor> findAllByMonitor_TenantId(String tenantId, Pageable pageable);
```
jjbuchan: I was gonna question this before but didn't... why have you started using `_` in these methods vs. `By`?

itzg: IntelliJ is helping me out with that, but here's the docs

https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.query-methods.query-property-expressions

It's because the query is traversing into the joined `Monitor` entity.

itzg: ...specifically, IntelliJ is preferring the recommended approach:
> To resolve this ambiguity you can use _ inside your method name to manually define traversal points.

jjbuchan: Interesting.

PR #22, opened 2019-05-02
SALUS-349: update parameters for salus-tools update

harageth:
# Resolves

SALUS-349

# What

Gives the string parameter that salus-tools needs for replacement

# How

Updates the parameter list in the pom

## How to test

After Installing (Running the mvn install target from salus-tools) the SwaggerJsonConverter will be in the .m2 folder and then you can run the `test` target here that will generate the internal swagger and run the SwaggerJsonConverter


harageth:
I just created this from a new branch.

PR #24, opened 2019-05-02
Remove CriteriaBuilder and use repository methods instead

jjbuchan:
# What

Removes all the `CriteriaBuilder` logic for building sql queries and replaces it by using standard repository options.

# How

The "how" is fairly obvious.  The only real decision was to start using `Optional` fields where we are getting at most one returned object.  I like this overall, but the downside is it makes the tests a little more tedious due to having to do various `isPresent()` and `get()` calls.

## How to test

Run all tests.  They still work!

# Why

Cleans up the code and standardizes things across all repos.

# TODO

Incoming PRs for model repo... I think that'll be it.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
@@ -17,10 +17,18 @@
 package com.rackspace.salus.monitor_management.repositories;
 
 import com.rackspace.salus.telemetry.model.Monitor;
+
+import java.util.Optional;
 import java.util.UUID;
+import java.util.stream.Stream;
+
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
 import org.springframework.data.repository.PagingAndSortingRepository;
 import org.springframework.stereotype.Repository;
 
 @Repository
 public interface MonitorRepository extends PagingAndSortingRepository<Monitor, UUID> {
+
+    Page<Monitor> findByTenantId(String tenantId, Pageable pageable);
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
@@ -18,10 +18,7 @@ package com.rackspace.salus.monitor_management.repositories;
 
 import com.rackspace.salus.telemetry.model.Monitor;
 
-import java.util.Optional;
 import java.util.UUID;
-import java.util.stream.Stream;
-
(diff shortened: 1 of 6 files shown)


src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
```diff
@@ -17,10 +17,18 @@
 package com.rackspace.salus.monitor_management.repositories;
 
 import com.rackspace.salus.telemetry.model.Monitor;
+
+import java.util.Optional;
 import java.util.UUID;
+import java.util.stream.Stream;
```
harageth: I dont think this or the java.util.Optional are being used here anymore

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -172,14 +146,9 @@ public Monitor getMonitor(String tenantId, UUID id) {
      * @return Stream of monitors.
      */
     public Stream<Monitor> getMonitorsAsStream() {
-        CriteriaBuilder cb = entityManager.getCriteriaBuilder();
-        CriteriaQuery<Monitor> cr = cb.createQuery(Monitor.class);
-        Root<Monitor> root = cr.from(Monitor.class);
-        cr.select(root);
-        return entityManager.createQuery(cr).getResultStream();
+        return Streams.stream(monitorRepository.findAll());
```
harageth: Considering the other stuff just wasnt working this looks a lot easier.

src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
jjbuchan: Correct.  Removed them now.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
itzg: Yeah, it's strange how going from an Iterable to a Stream is awkward even though they seem close in concept. I guess the non-Guava might be

```
StreamSupport.stream(
  monitorRepository.findAll().spliterator(),
  false
)
```

jjbuchan: Should we be doing it that way instead?  I know no reason to not use guava other than it saying it's unstable... but you've said before that it doesn't matter.

I'm happy to go either way.

itzg: I like guava, so I'd say keep it as you have it. What I found is just something we can try next time that's the reason we need to pull in guava.

PR #25, opened 2019-05-03
Update API Examples

jjbuchan:
When trying to get the whole bundle working I realized these API examples are out of date.  I figured out what payload I should be able to provide, but this `POST` example currently returns a `500`.  We need to figure out what's going on here, fix the problem, and then update these examples with whatever is correct.

FYI... the failure is due to `http://localhost:8085/api/tenant/aaaaaa/resourceLabels?agent.os=darwin` returning a `404`.  I'm unable to reproduce this via hitting the reosurce api directly.  It is only failing via the resourceApiClient being used by the monitor mgmt service.

Currently my only thought is that its somehow getting the `rootUri` parameters mixed up and querying the wrong service, but the logs don't back that up, and stepping through the request in debug mode also hasn't helped me identify the problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ HTTP POST to http://localhost:8089/api/tenant/aaaaa/monitors with body:
 }
 ```
 
-## Update an existing monitor to use wildcards
+## Update an existing monitor to use template placeholders
 
 http PUT localhost:8089/api/tenant/aaaaa/monitors/$MonitorId with body:
 ```
--- a/README.md
+++ b/README.md
@@ -21,9 +21,19 @@ You can trigger these events to be posted by utilizing some of the API operation
 Examples of a subset of the available API operations.
 
 ## Create a new monitor
+
+HTTP POST to http://localhost:8089/api/tenant/aaaaa/monitors with body:
 ```
-echo '{"monitorName":"mon1", "content":"content1", "agentType":"TELEGRAF"}' | http POST 'localhost:8089/api/tenant/aaaaa/monitors' | tee /tmp/newMonitor.txt; 
+{
+	"labelSelector": {"agent.os":"darwin"},
+	"details": {
+		"type": "local",
+		"agentType": "TELEGRAF",
+		"plugin": {"type":"mem"}
+	}
+}
 ```
+
 ## Save the monitor id in a env var
 export MonitorId=`jq -r .id /tmp/newMonitor.txt`
 
--- a/README.md
+++ b/README.md
@@ -20,26 +20,54 @@ You can trigger these events to be posted by utilizing some of the API operation
 # API Operations
 Examples of a subset of the available API operations.
 
-## Create a new monitor
+## Create a new local monitor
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,9 +21,19 @@ You can trigger these events to be posted by utilizing some of the API operation
...
 ## Create a new monitor
+
+HTTP POST to http://localhost:8089/api/tenant/aaaaa/monitors with body:
 ```
-echo '{"monitorName":"mon1", "content":"content1", "agentType":"TELEGRAF"}' | http POST 'localhost:8089/api/tenant/aaaaa/monitors' | tee /tmp/newMonitor.txt; 
+{
+	"labelSelector": {"agent.os":"darwin"},
+	"details": {
+		"type": "local",
+		"agentType": "TELEGRAF",
```
harageth: Are you sure that the agentType is in used? I dont see it in DetailedMonitorInput and its not anywhere to be found in the MonitorDetails or any of the sub classes.

itzg: @harageth is right this isn't explicitly needed anymore. Instead the details conversion code will pick off the agent type from the `@ApplicableAgentType` here

https://github.com/racker/salus-telemetry-monitor-management/blob/master/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Mem.java#L23

README.md
```diff
@@ -21,9 +21,19 @@ You can trigger these events to be posted by utilizing some of the API operation
 Examples of a subset of the available API operations.
 
 ## Create a new monitor
+
+HTTP POST to http://localhost:8089/api/tenant/aaaaa/monitors with body:
 ```
-echo '{"monitorName":"mon1", "content":"content1", "agentType":"TELEGRAF"}' | http POST 'localhost:8089/api/tenant/aaaaa/monitors' | tee /tmp/newMonitor.txt; 
+{
+	"labelSelector": {"agent.os":"darwin"},
```
itzg: Actually it's "agent.discovered.os"

jjbuchan: That's a good point... my db has it like this
image

I'll have to see why `discovered` isn't being used.

jjbuchan: Oh, that is being populated by the envoy and I guess I'm using an old version.

itzg: Hmm, double check your envoy code is indeed the latest. The envoy is the one pre-pending with `discovered.` with these decorated variables:

https://github.com/racker/salus-telemetry-envoy/blob/master/config/labels.go#L45

README.md
```diff
@@ -20,16 +20,54 @@ You can trigger these events to be posted by utilizing some of the API operation
...
+      "urls": ["127.0.0.1"]
+    }
+  }
+}
 ```
-## Save the monitor id in a env var
-export MonitorId=`jq -r .id /tmp/newMonitor.txt`
 
-## Update an existing monitor
+## Update an existing monitor to use wildcards
```
itzg: or how about
```suggestion
## Update an existing monitor to use template placeholders
```

jjbuchan: Thanks.  I didn't know what name to go for.

PR #28, opened 2019-05-09
add Procstat with custom validator

harageth:
# Resolves

SALUS-273 https://jira.rax.io/browse/SALUS-273

# What

Adds the Procstat configuration to Monitors
Procstat is a Telegraf plugin that allows monitoring specifics about certain processes. From their documentation
> The procstat plugin can be used to monitor the system resource usage of one or more processes. The procstat_lookup metric displays the query information, specifically the number of PIDs returned on a search

# How

Adds the Procstat class with a custom validator

## How to test

Right now you would have to send an insomnia request of the procstat type. 

# Why

This is a typical agent plugin that we will need for some process monitoring.

We could validate through a constructor but this ties into Spring on object creation from the request. We still have to write the validation code 

# TODO

Make validator actually validate that we have only one value set
Add tests


src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Procstat.java
```diff
@@ -0,0 +1,56 @@
...
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+import javax.validation.constraints.NotEmpty;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@ApplicableAgentType(AgentType.TELEGRAF)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@OneOfValidator.OneOf(groups = OneOfValidator.OneOf.class)
```
itzg: We discussed on the other PR, but just for notes, the `groups` probably won't be needed here for just the constraint enforcement.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Procstat.java
```diff
@@ -0,0 +1,56 @@
...
+public class Procstat extends LocalPlugin {
+    @NotEmpty(groups = OneOfValidator.OneOf.class)
+    String pidfile;
```
itzg: Needs to be `pid_file` to match the expected TOML field

https://github.com/influxdata/telegraf/blob/master/plugins/inputs/procstat/procstat.go#L26

But fun fact, can also use camelcase `pidFile` since I convert camelcase to snakecase here

https://github.com/racker/salus-telemetry-envoy/blob/master/agents/converters.go#L88

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Procstat.java
```diff
@@ -0,0 +1,56 @@
...
+    pgrep -f
+    pattern;
+    systemd unit name
+    systemd_unit;
+    cgroup name or path
+    cgroup;
+    windows service name
+    win_service
+
+    process_name
```
itzg: It looks like `process_name` is a field we want to allow. The others in this section make sense to not allow.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Procstat.java
```diff
@@ -0,0 +1,56 @@
...
+
+import javax.validation.constraints.NotEmpty;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@ApplicableAgentType(AgentType.TELEGRAF)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@OneOfValidator.OneOf(groups = OneOfValidator.OneOf.class)
+public class Procstat extends LocalPlugin {
+    @NotEmpty(groups = OneOfValidator.OneOf.class)
```
itzg: I don't think you'll need to include it on each field...unless you wanted to change

https://github.com/racker/salus-telemetry-monitor-management/pull/28/files#diff-a1e4572980aa1b6119b941d88b47b6d2R20

use reflection to find the annotated fields rather than hardcode them.

src/main/java/com/rackspace/salus/monitor_management/web/validator/OneOfValidator.java
```diff
@@ -0,0 +1,43 @@
...
+import javax.validation.Payload;
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+public class OneOfValidator implements ConstraintValidator <OneOfValidator.OneOf, Procstat> {
```
itzg: Since this is currently coded very specific to `Procstat` I'd say just call this `ProcstatValidator`.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Procstat.java
harageth: So if I leave off the groups annotation here then I dont need to annotate every field because it will just validate for everything?

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Procstat.java
harageth: I will camelCase it. Thought I had copied that over correctly. Oops

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Procstat.java
itzg: Correct since it should activate your validator on the class level annotation.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Procstat.java
itzg: I was double checking since the telegraf plugins are not totally consistent with word breaks on field names.

src/main/java/com/rackspace/salus/monitor_management/web/validator/OneOfValidator.java
harageth: Yeah that makes more sense. The fields just dont really allow for us to put it at a higher level. This might cause some issues with some of the plugins where we have an explosion of these ConstraintValidators.

itzg: At that point we can implement a general purpose one that uses reflection to find the annotated fields.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Procstat.java
```diff
@@ -3,34 +3,27 @@
...
 import lombok.EqualsAndHashCode;
 
-import javax.validation.constraints.NotEmpty;
-
 @Data
 @EqualsAndHashCode(callSuper = true)
 @ApplicableAgentType(AgentType.TELEGRAF)
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@OneOfValidator.OneOf(groups = OneOfValidator.OneOf.class)
+@ProcstatValidator.OneOf(groups = ProcstatValidator.OneOf.class)
```
itzg: You don't need to set `groups` here.

jjbuchan:
This is a WIP so I won't hold it against you too much, but the PR description doesn't say anything about why we're adding procstat.  As far as I can tell the _why_ only relates to the _how_ and not the _what_.  I had to look at jira to see what the why really was, and no one likes jira!

src/main/java/com/rackspace/salus/monitor_management/web/model/LocalPlugin.java
```diff
@@ -20,17 +20,15 @@
...
+import com.rackspace.salus.monitor_management.web.model.telegraf.*;
 
 @JsonTypeInfo(use = Id.NAME, property = "type")
 @JsonSubTypes({
     @Type(name = "cpu", value = Cpu.class),
     @Type(name = "disk", value = Disk.class),
     @Type(name = "diskio", value = DiskIo.class),
-    @Type(name = "mem", value = Mem.class)
+    @Type(name = "mem", value = Mem.class),
+    @Type(name = "procstat", value = Procstat.class)
```
itzg: Thanks, I had missed that previously.

README.md
```diff
@@ -32,9 +32,9 @@ export MonitorId=`jq -r .id /tmp/newMonitor.txt`
 echo '{"content":"content1xxxxx"}' | http PUT localhost:8089/api/tenant/aaaaa/monitors/$MonitorId
 ```
 
-## Get a stream of all monitors
+## Get all monitors
 ```
-curl localhost:8089/api/monitorsAsStream
+curl localhost:8089/api/monitors?page=0&size=100
```
itzg: Strange. The git merge I suggested is showing the changes from master even when looking at all commits. Can't win either way I guess. Let's just stop using branches :)

harageth: that seems reasonable. We can go back to communicating with Smoke Signals... Which I dont think has an RFC yet.

src/main/java/com/rackspace/salus/monitor_management/web/validator/ProcstatValidator.java
```diff
@@ -0,0 +1,63 @@
...
+        if(!isNullOrEmpty(monitor.getSystemdUnit())) {
+            count++;
+        }
+        if(!isNullOrEmpty(monitor.getCgroup())) {
+            count++;
+        }
+        if(!isNullOrEmpty(monitor.getWinService())) {
+            count++;
+        }
+        if(count != 1) {
```
itzg: Strange, the REST API let me create the monitor using the following, but telegraf isn't happy without one of those set.

```
{
  "labelSelector": {
    "agent_environment": "localdev"
  },
  "details": {
    "type": "local",
    "plugin": {
      "type": "procstat"
    }
  }
}
```

harageth: I could have sworn I had it in my validator to error out if nothing was set

harageth: And yeah... this should error out if nothing gets set

PR #29, opened 2019-05-09
Add zone management services

jjbuchan:
# What

This felt like a good place to push a PR without it getting too big.

 * Bumps to Java 12
 * Adds zone repository, entity, models, and management service
 * Adds checks to monitor create to verify the provided zone can be used

# How
Mostly similar to every other service we've created.  A lot of filling in all the generic methods etc.

## How to test

Run included tests or try things manually.

# Why

This allows us to do more validation around zones for monitor creation and envoy attachment.  It is also the first step towards having users create their own zones and implementing the redistribution of monitors if an envoy/poller disconnects for too long.

Cyclical dependencies become a bigger problem if zone management is in a separate repo.

# TODO

 * There are a couple `// TBD` comments that will be removed/replaced when the use of timeouts is implemented.
 * More javadoc will be added in another pr too
 * Look into why `findByTenantIdAndZonesContains` only worked in tests?
 * Add public zone creation api

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
   <description>Service for managing monitors</description>
 
   <properties>
-    <java.version>12</java.version>
+    <java.version>1.8</java.version>
   </properties>
 
   <dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
   <description>Service for managing monitors</description>
 
   <properties>
-    <java.version>1.8</java.version>
+    <java.version>12</java.version>
   </properties>
 
   <dependencies>
--- a/src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
@@ -1,21 +1,41 @@
+/*
+ * Copyright 2019 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.rackspace.salus.monitor_management.entities;
 
+import com.rackspace.salus.monitor_management.types.ZoneState;
 import com.rackspace.salus.monitor_management.web.model.ZoneDTO;
 import lombok.Data;
 import org.hibernate.annotations.Type;
 import org.springframework.boot.convert.DurationUnit;
 
 import javax.persistence.*;
+import javax.validation.constraints.NotNull;
+
 import org.hibernate.validator.constraints.NotBlank;
 
 import java.io.Serializable;
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
+import java.util.List;
 import java.util.UUID;
 
 @Entity
 @Table(name = "zones", uniqueConstraints = {
         @UniqueConstraint(columnNames = {"tenant_id", "name"})})
+
 @Data
 public class Zone implements Serializable {
     @Id
@@ -23,22 +43,80 @@ public class Zone implements Serializable {
     @Type(type="uuid-char")
     UUID id;
 
+    /**
+     * Contains the unique name/label for the zone.
+     * Public zones should have a "public/" prefix and then a trailing region descriptor.
+     * e.g. "public/us-central-1"
+     */
     @NotBlank
     @Column(name="name")
     String name;
 
+    /**
+     * Contains an optional hosting provider of the zone.
+     * e.g. Rackspace, Google, Amazon
+     */
+    @Column(name="provider")
+    String provider;
+
+    /**
+     * Contains an optional region to more precisely define where the zone is running.
+     * e.g. for Rackspace, dfw3 may be used; for Google, europe-west3.
+     */
+    @Column(name="provider_region")
+    String providerRegion;
+
+    /**
+     * Defines whether a zone is public or private.
+     */
+    @NotNull
+    @Column(name="is_public")
+    boolean isPublic;
+
+    /**
+     * Contains an optional list of ipv4 and ipv6 ranges that the pollers in this zone reside in.
+     * This can be used to whitelist ranges to allow for remote polling.
+     */
+    @ElementCollection(fetch = FetchType.EAGER)
+    @Column(name="source_ips")
+    List<String> sourceIpRanges;
+
+    /**
+     * Defines whether the zone is available or not.
+     * This helps determine whether new monitors can be added to the zone.
+     */
+    @Enumerated(EnumType.STRING)
+    @Column(name="state")
+    ZoneState state;
+
+    /**
+     * Contains the tenant that owns the private zone or "_PUBLIC_" for public zones.
+     */
     @NotBlank
     @Column(name="tenant_id")
     String tenantId;
 
+    /**
+     * Contains a timeout value which begins to countdown after an envoy/poller has disconnected.
+     * If the timeout is met, the monitors bound to it will be distributed to other available pollers.
+     */
     @DurationUnit(ChronoUnit.SECONDS)
     @Column(name="envoy_timeout")
     Duration envoyTimeout = Duration.ofSeconds(120);
 
+    /**
+     * Converts the Zone object to a ZoneDTO which is a stripped down version to be used
+     * as output in the APIs.
+     *
+     * @return A ZoneDTO with fields mapped from the Zone.
+     */
     public ZoneDTO toDTO() {
         return new ZoneDTO()
                 .setName(name)
-                .setTenantId(tenantId)
-                .setEnvoyTimeout(envoyTimeout.getSeconds());
+                .setEnvoyTimeout(envoyTimeout.getSeconds())
+                .setProvider(provider)
+                .setProviderRegion(providerRegion)
+                .setPublic(isPublic)
+                .setSourceIpRanges(sourceIpRanges);
     }
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
@@ -1,0 +1,44 @@
+package com.rackspace.salus.monitor_management.entities;
+
+import com.rackspace.salus.monitor_management.web.model.ZoneDTO;
+import lombok.Data;
+import org.hibernate.annotations.Type;
+import org.springframework.boot.convert.DurationUnit;
+
+import javax.persistence.*;
+import org.hibernate.validator.constraints.NotBlank;
+
+import java.io.Serializable;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.util.UUID;
+
+@Entity
+@Table(name = "zones", uniqueConstraints = {
+        @UniqueConstraint(columnNames = {"tenant_id", "name"})})
+@Data
+public class Zone implements Serializable {
+    @Id
+    @GeneratedValue
+    @Type(type="uuid-char")
+    UUID id;
+
+    @NotBlank
+    @Column(name="name")
+    String name;
+
+    @NotBlank
+    @Column(name="tenant_id")
+    String tenantId;
+
+    @DurationUnit(ChronoUnit.SECONDS)
+    @Column(name="envoy_timeout")
+    Duration envoyTimeout = Duration.ofSeconds(120);
+
+    public ZoneDTO toDTO() {
+        return new ZoneDTO()
+                .setName(name)
+                .setTenantId(tenantId)
+                .setEnvoyTimeout(envoyTimeout.getSeconds());
+    }
+}
--- a/src/main/java/com/rackspace/salus/monitor_management/errors/ZoneAlreadyExists.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/errors/ZoneAlreadyExists.java
@@ -18,7 +18,7 @@ package com.rackspace.salus.monitor_management.errors;
 import org.springframework.http.HttpStatus;
 import org.springframework.web.bind.annotation.ResponseStatus;
 
-@ResponseStatus(HttpStatus.BAD_REQUEST)
+@ResponseStatus(HttpStatus.UNPROCESSABLE_ENTITY)
 public class ZoneAlreadyExists extends RuntimeException {
     public ZoneAlreadyExists(String message) { super(message);}
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/errors/ZoneAlreadyExists.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/errors/ZoneAlreadyExists.java
@@ -13,16 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+package com.rackspace.salus.monitor_management.errors;
 
-package com.rackspace.salus.monitor_management.config;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
 
-import lombok.Data;
-import org.springframework.boot.context.properties.ConfigurationProperties;
-import org.springframework.stereotype.Component;
-
-@ConfigurationProperties("services")
-@Component
-@Data
-public class ServicesProperties {
-    String resourceManagementUrl;
+@ResponseStatus(HttpStatus.BAD_REQUEST)
+public class ZoneAlreadyExists extends RuntimeException {
+    public ZoneAlreadyExists(String message) { super(message);}
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
@@ -18,9 +18,11 @@ package com.rackspace.salus.monitor_management.repositories;
 
 import com.rackspace.salus.telemetry.model.Monitor;
 
+import java.util.List;
 import java.util.UUID;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
+import org.springframework.data.jpa.repository.Query;
 import org.springframework.data.repository.PagingAndSortingRepository;
 import org.springframework.stereotype.Repository;
 
@@ -28,4 +30,10 @@ import org.springframework.stereotype.Repository;
 public interface MonitorRepository extends PagingAndSortingRepository<Monitor, UUID> {
 
     Page<Monitor> findByTenantId(String tenantId, Pageable pageable);
+
+    // this works in tests but not in reality.
+    List<Monitor> findByTenantIdAndZonesContains(String tenantId, String zone);
+
+    @Query("select m from Monitor m where m.tenantId = :tenantId and :zone member of m.zones")
+    List<Monitor> customFindByTenantIdAndZonesContains(String tenantId, String zone);
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -18,10 +18,12 @@ package com.rackspace.salus.monitor_management.services;
 
 import static com.rackspace.salus.telemetry.etcd.types.ResolvedZone.createPrivateZone;
 import static com.rackspace.salus.telemetry.etcd.types.ResolvedZone.createPublicZone;
+import static java.util.function.Predicate.not;
 
 import com.google.common.collect.Streams;
 import com.rackspace.salus.monitor_management.config.ZonesProperties;
 import com.rackspace.salus.monitor_management.entities.BoundMonitor;
+import com.rackspace.salus.monitor_management.entities.Zone;
 import com.rackspace.salus.monitor_management.repositories.BoundMonitorRepository;
 import com.rackspace.salus.monitor_management.repositories.MonitorRepository;
 import com.rackspace.salus.monitor_management.web.model.MonitorCU;
@@ -49,12 +51,15 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import javax.persistence.EntityManager;
 import javax.persistence.PersistenceContext;
 import javax.validation.Valid;
+
+import com.rackspace.salus.monitor_management.web.model.ZoneDTO;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.context.properties.PropertyMapper;
@@ -75,6 +80,7 @@ public class MonitorManagement {
     private final MonitorEventProducer monitorEventProducer;
     private final MonitorContentRenderer monitorContentRenderer;
     private final ResourceApi resourceApi;
+    private final ZoneManagement zoneManagement;
     private final ZonesProperties zonesProperties;
 
     private final MonitorRepository monitorRepository;
@@ -94,7 +100,7 @@ public class MonitorManagement {
                              MonitorEventProducer monitorEventProducer,
                              MonitorContentRenderer monitorContentRenderer,
                              ResourceApi resourceApi,
-                             ZonesProperties zonesProperties,
+                             ZoneManagement zoneManagement, ZonesProperties zonesProperties,
                              JdbcTemplate jdbcTemplate) {
         this.monitorRepository = monitorRepository;
         this.entityManager = entityManager;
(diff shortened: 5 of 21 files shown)


src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
```diff
@@ -18,14 +18,22 @@
...
 import org.springframework.data.repository.PagingAndSortingRepository;
 import org.springframework.stereotype.Repository;
 
 @Repository
 public interface MonitorRepository extends PagingAndSortingRepository<Monitor, UUID> {
 
     Page<Monitor> findByTenantId(String tenantId, Pageable pageable);
+
+    // this works in tests but not in reality.
+    List<Monitor> findByTenantIdAndZonesContains(String tenantId, String zone);
```
jjbuchan: Need to either look into this or just remove it and rename the custom one.

src/main/java/com/rackspace/salus/monitor_management/services/ZoneManagement.java
```diff
@@ -0,0 +1,137 @@
...
+
+    public Zone updateZone(String tenantId, String name, @Valid ZoneUpdate updatedZone) {
+        Zone zone = getZone(tenantId, name).orElseThrow(() ->
+                new NotFoundException(String.format("No zone found named %s on tenant %s",
+                        name, tenantId)));
+
+        zone.setEnvoyTimeout(Duration.ofSeconds(updatedZone.getPollerTimeout()));
+        zoneRepository.save(zone);
+
+        // TBD: need to restart watchers?
```
jjbuchan: I know about this.  Will be handled in the next set of PRs.

src/main/java/com/rackspace/salus/monitor_management/services/ZoneManagement.java
```diff
@@ -0,0 +1,137 @@
...
+        long activeEnvoys = getActiveEnvoyCountForZone(zone);
+        log.debug("Found {} active envoys for zone {}", activeEnvoys, name);
+        if (activeEnvoys > 0) {
+            throw new IllegalArgumentException(
+                    String.format("Cannot remove zone with connected pollers. Found %s.", activeEnvoys));
+        }
+
+        zoneRepository.deleteById(zone.getId());
+
+        // TBD: remove expected entries in etcd?
```
jjbuchan: I know about this.  Will be handled in the next set of PRs.

src/main/java/com/rackspace/salus/monitor_management/web/model/ZoneCreate.java
```diff
@@ -0,0 +1,20 @@
...
+import javax.validation.constraints.Min;
+import javax.validation.constraints.Pattern;
+import org.hibernate.validator.constraints.NotBlank;
+import java.io.Serializable;
+
+@Data
+public class ZoneCreate implements Serializable {
+
+    @NotBlank
+    @Pattern(regexp = "^[\\p{Alnum}]+$", message = "Only alphanumeric characters can be used")
```
jjbuchan: Annoyingly `podam` can't handle this and ends up causing it to leave all `name` fields blank.

src/main/java/com/rackspace/salus/monitor_management/services/ZoneManagement.java
```diff
@@ -0,0 +1,137 @@
...
+@Service
+public class ZoneManagement {
+    private final ZoneRepository zoneRepository;
+    private final ZoneStorage zoneStorage;
+    private final MonitorManagement monitorManagement;
+
+    static final String PUBLIC = "_PUBLIC_";
+
+    @Autowired
+    public ZoneManagement(ZoneRepository zoneRepository, ZoneStorage zoneStorage, @Lazy MonitorManagement monitorManagement) {
```
jjbuchan: Thoughts on `@Lazy`?

src/main/java/com/rackspace/salus/monitor_management/services/ZoneManagement.java
itzg: I'm hopeful this will have minimal impact since the timeout value probably just needs to be obtained when an envoy leaves the zone. At that point a timer is started with the currently stored timeout of the zone. If the user changes the timeout mid-timer, then I'd say it sticks the previous timeout duration and the next envoy departure will use the new timeout value.

src/main/java/com/rackspace/salus/monitor_management/services/ZoneManagement.java
```diff
@@ -0,0 +1,137 @@
...
+
+        return zone;
+    }
+
+    public void removeZone(String tenantId, String name) {
+        Zone zone = getZone(tenantId, name).orElseThrow(() ->
+                new NotFoundException(String.format("No zone found named %s on tenant %s",
+                        name, tenantId)));
+
+        int monitors = monitorManagement.getMonitorsForZone(tenantId, name).size();
```
itzg: It looks you can declare `count*` repository methods (from example 5 of https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.core-concepts).

src/main/java/com/rackspace/salus/monitor_management/services/ZoneManagement.java
```diff
@@ -0,0 +1,137 @@
...
+        if(monitors > 0) {
+            throw new IllegalArgumentException(
+                    String.format("Cannot remove zone with configured monitors. Found %s.", monitors));
+        }
+
+        long activeEnvoys = getActiveEnvoyCountForZone(zone);
+        log.debug("Found {} active envoys for zone {}", activeEnvoys, name);
+        if (activeEnvoys > 0) {
+            throw new IllegalArgumentException(
+                    String.format("Cannot remove zone with connected pollers. Found %s.", activeEnvoys));
```
itzg:
```suggestion
                    String.format("Cannot remove zone with connected pollers. Found %d.", activeEnvoys));
```

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
```diff
@@ -0,0 +1,64 @@
...
+    @GetMapping("/tenant/{tenantId}/zones/{name}")
+    public ZoneDTO getByZoneName(@PathVariable String tenantId, @PathVariable String name) {
+        Optional<Zone> zone = zoneManagement.getZone(tenantId, name);
+        return zone.orElseThrow(() -> new NotFoundException(String.format("No zone found for %s on tenant %s",
+                name, tenantId)))
+                .toDTO();
+    }
+
+    @PostMapping("/tenant/{tenantId}/zones")
+    @ResponseStatus(HttpStatus.CREATED)
```
itzg: Nice, we should do this on the others.

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
```diff
@@ -0,0 +1,64 @@
...
+
+import javax.validation.Valid;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+@Slf4j
+@RestController
+@RequestMapping("/api")
+public class ZoneApiController implements ZoneApi {
```
itzg: It looks like this needs some admin (non-{tenantId}) API requests for management of our public zones.

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @DeleteMapping("/tenant/{tenantId}/zones/{name}")
+    @ResponseStatus(HttpStatus.NO_CONTENT)
+    public void delete(@PathVariable String tenantId, @PathVariable String name) {
+        zoneManagement.removeZone(tenantId, name);
+    }
+
+    @GetMapping("/tenant/{tenantId}/zones")
+    public List<ZoneDTO> getAvailableZones(@PathVariable String tenantId) {
```
itzg: It looks like `ZoneDTO`'s `tenantId` will end up with "_PUBLIC_", which is meant to be an internal token. It also creates an awkward usability since the expectation of remote monitors is that the zone references are prefixed with the configure public designation, such as "public/", which is processed here

https://github.com/racker/salus-telemetry-monitor-management/blob/master/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java#L376

If I were an end user, I'd want to get the exact list of zone strings back that can be passed into monitor create/update calls.

`getByZoneName` is also awkward as is, since they'll get public zones back via this `getAvailableZones` but they can never get a public zone via `getByZoneName` since the `{tenantId}` gets populated by our public API proxy code as obtained from their Identity authentication.

src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApi.java
```diff
@@ -17,9 +17,13 @@
...
 import com.rackspace.salus.monitor_management.web.model.BoundMonitorDTO;
+import com.rackspace.salus.telemetry.model.Monitor;
+
 import java.util.List;
 
 public interface MonitorApi {
 
   List<BoundMonitorDTO> getBoundMonitors(String envoyId);
+
+  List<Monitor> getMonitorsForZone(String tenantId, String zone);
```
itzg: Do other microservices need this yet? Also, if going with my suggestion of moving the service method into `ZoneManagement` it would seem logical to move this to `ZoneApi`; however, that creates an inconsistency in response types, so I'd say leave here regardless of the other change.

src/main/java/com/rackspace/salus/monitor_management/web/client/ZoneApiClient.java
```diff
@@ -0,0 +1,79 @@
...
+ *   configured with the appropriate root URI for locating the zone management service.
+ *   The following is an example of a configuration bean that does that:
+ * </p>
+ *
+ * <pre>
+ {@literal @}Configuration
+ public class RestClientsConfig {
+
+   {@literal @}Bean
+   public ZoneApi resourceApi(RestTemplateBuilder restTemplateBuilder) {
```
itzg:
```suggestion
   public ZoneApi zoneApi(RestTemplateBuilder restTemplateBuilder) {
```

src/main/java/com/rackspace/salus/monitor_management/web/model/ZoneCreate.java
itzg: Yeah, that would have been impressive if it could interpret the regex and create a compliant string.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -229,8 +235,16 @@ public void testGetMonitorForUnauthorizedTenant() {
     public void testCreateNewMonitor() {
         MonitorCU create = podamFactory.manufacturePojo(MonitorCU.class);
         create.setSelectorScope(ConfigSelectorScope.ALL_OF);
+
+        @SuppressWarnings("unchecked")
+        List<Zone> zones = podamFactory.manufacturePojo(ArrayList.class, Zone.class);
```
itzg: Hmm, the use of podam here is masking some concerns I have about the prefixed-public-zone convention. I would like to see some test cases specifically verifying zones like "public/west" vs "somePrivateZone".

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -248,6 +262,20 @@ public void testCreateNewMonitor() {
         assertTrue(Maps.difference(returned.getLabelSelector(), retrieved.get().getLabelSelector()).areEqual());
     }
 
+    @Test(expected = IllegalArgumentException.class)
+    public void testCreateNewMonitorInvalidZone() {
+        MonitorCU create = podamFactory.manufacturePojo(MonitorCU.class);
+        create.setSelectorScope(ConfigSelectorScope.ALL_OF);
+        String tenantId = RandomStringUtils.randomAlphanumeric(10);
+
+        assertThat(create.getZones().size(), greaterThan(0));
```
itzg: What is this asserting exactly? That podam populated something? Set the zones explicitly rather than counting on podam in that case.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -1416,4 +1453,17 @@ public void testhandleResourceEvent_removedResource() {
         verifyNoMoreInteractions(boundMonitorRepository, envoyResourceManagement,
             zoneStorage, monitorEventProducer, resourceApi);
     }
+
+    @Test
+    public void testGetMonitorsForZone() {
```
itzg: Please add a variant for public zones specifically.

src/test/java/com/rackspace/salus/monitor_management/services/ZoneManagementTest.java
```diff
@@ -0,0 +1,157 @@
...
+public class ZoneManagementTest {
+
+    @MockBean
+    MonitorManagement monitorManagement;
+
+    @MockBean
+    ZoneStorage zoneStorage;
+
+    @MockBean
+    AgentsCatalogService agentsCatalogService;
```
itzg: Doesn't seem to need this

src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
```diff
@@ -0,0 +1,44 @@
...
+@Data
+public class Zone implements Serializable {
+    @Id
+    @GeneratedValue
+    @Type(type="uuid-char")
+    UUID id;
+
+    @NotBlank
+    @Column(name="name")
+    String name;
```
itzg: It appears that public zones will be stored with the "public/" prefix; however, that will then drift from the configuration property for that prefix

https://github.com/racker/salus-telemetry-monitor-management/blob/master/src/main/java/com/rackspace/salus/monitor_management/config/ZonesProperties.java#L31

src/test/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiControllerTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Test
+    public void testGetByZoneName() throws Exception {
+        final Zone expectedZone = podamFactory.manufacturePojo(Zone.class);
+        when(zoneManagement.getZone(any(), any()))
+                .thenReturn(Optional.of(expectedZone));
+
+        mvc.perform(get(
+                "/api/tenant/{tenantId}/zones/{name}",
+                "t-1", "r-1")
```
itzg: "r-1" should be a zone name, so the "r" is misleading. I was hoping to see some uses of public zones to answer my concerns.

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
jjbuchan: It does.  I realize I'm running behind so was thinking of leaving that until later... but failed to mention it in the PR.

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
jjbuchan: I thought we already did...

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
jjbuchan: > which is meant to be an internal token

I thought it was going to be returned via the API vs. those `zoneTenant` fields being blank?

jjbuchan: We might need to chat more about this because I think you're trying to say more than what I'm reading.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
jjbuchan: Yeah, I was a little torn on this one too.  I'll switch.

jjbuchan: It's the fact it can't handle all our annotations that made me do this.  I initially also wasn't sure how it would do it... but it ends up creating a list of 5 entries each time.  It's a trust problem.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -656,4 +684,8 @@ void handleResourceEvent(ResourceEvent event) {
         }
         return monitors;
     }
-}
+
+    public List<Monitor> getMonitorsForZone(String tenantId, String zone) {
```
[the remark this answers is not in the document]
jjbuchan: Yeah, the fact this is called from the ZM delete method is the remaining circular problem.  I'll try move things around.

src/main/java/com/rackspace/salus/monitor_management/services/ZoneManagement.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.Optional;
+
+@Slf4j
+@Service
+public class ZoneManagement {
+    private final ZoneRepository zoneRepository;
+    private final ZoneStorage zoneStorage;
+    private final MonitorManagement monitorManagement;
+
+    static final String PUBLIC = "_PUBLIC_";
```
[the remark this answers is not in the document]
jjbuchan: Agree.  When I first added this it was because `ResolvedZone` wasn't pulled in and it seemed overkill to do it for this one thing... but as I now look at ZM and MM I'm not sure if that was ever the case, so looks like an oversight on my end.

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
itzg: No, you're right, the bound monitor API does return a `_PUBLIC_` in the `zoneTenantId`...but then I'm now realizing that's why I wanted to use a special blank value there. To follow my own advice the BoundMonitorDTO should really transform the two fields into just a `zone` which is either prefixed with the public designation or left as is for private.

jjbuchan: >  chat more about this

For that, I did mean your whole comment and not just that one point I commented on already.

src/main/java/com/rackspace/salus/monitor_management/services/ZoneManagement.java
itzg: The javadocs don't help clarify

https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/context/annotation/Lazy.html

but I believe it's meant for delaying the initialization of a bean so that resource intensive operations it abstract can be put off until the bean is actually retrieved and used.

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
itzg: Maybe so, I didn't look yet. I just know I haven't been using it :)

jjbuchan: Just looked and I had been using it in the controllers.  The new place I now use it is in on the error messages.

src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
itzg: As a note so I don't forget: I'm liking the idea of the boolean `isPublic` field as I'm looking at combining zone tenant and ID in BoundMonitorDTO into a `zone` field. The remaining `zoneTenantId` could be replaced by an `isPublic` boolean since a private zone's tenant probably won't ever be different from the associated monitor's tenant.

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
jjbuchan: Note: Add query parameters to this to lookup individual zones.  Returns empty list or list of 1.

Remove `getByZoneName`

pom.xml
```diff
@@ -32,7 +32,7 @@
   <description>Service for managing monitors</description>
 
   <properties>
-    <java.version>1.8</java.version>
+    <java.version>12</java.version>
```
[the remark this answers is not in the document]
itzg: I'm now thinking we should save this for a follow-up PR since the base image needs to be changed first in salus-app-base.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -157,9 +164,11 @@ public MonitorManagement(MonitorRepository monitorRepository, EntityManager enti
      * @return The newly created monitor.
      */
     public Monitor createMonitor(String tenantId, @Valid MonitorCU newMonitor) throws IllegalArgumentException, AlreadyExistsException {
-      log.debug("Creating monitor={} for tenant={}", newMonitor, tenantId);
+        log.debug("Creating monitor={} for tenant={}", newMonitor, tenantId);
+
+        validateMonitoringZones(tenantId, newMonitor.getZones());
```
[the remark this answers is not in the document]
itzg: Sorry, `determineMonitoringZones` was the method I meant; however, I think your PR's call location is correct after we talked about the public zone prefixing.

pom.xml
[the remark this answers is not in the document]
jjbuchan: Yeah, probably should.  I'll have to tweak my one use of java11/12 features first.

src/test/java/com/rackspace/salus/monitor_management/services/ZoneManagementTest.java
jjbuchan: Oh yeah... thought I removed that but this was confusing me yesterday.  Tests kept failing to start up due to numerous etcd-adapter things.  Importing `ObjectMapper` solved it in the end, but I wasn't too sure why it was even trying to start those tihngs anyway.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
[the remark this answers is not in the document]
jjbuchan: Hmm, I wasn't thinking about that use-case/location.  I think we should probably trust what is returned by `determineMonitoringZones` without doing any extra validation.  Since that happens behind the scenes and we are preventing users from removing zones that have monitors assigned, we should be safe.

src/main/java/com/rackspace/salus/monitor_management/services/ZoneManagement.java
jjbuchan: I was starting to think that could cause problems if all envoys disconnected, but even then it should only be a small amount that are configured as pollers and end up hitting the zone api.

itzg: We can always use Spring Cache to minimize Envoy stampedes hitting APIs and/or DB.

src/main/java/com/rackspace/salus/monitor_management/errors/ZoneAlreadyExists.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.rackspace.salus.monitor_management.errors;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(HttpStatus.BAD_REQUEST)
```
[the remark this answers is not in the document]
jjbuchan: After reading up I'm going with 422 Unprocessable Entity - https://tools.ietf.org/html/rfc4918#section-11.2

I was starting to lean that way and then read.  Since we like to use github as a good example, I'll stick with it.

> For the record, 422 is also the status code GitHub uses when you try to create a repository by a name already in use.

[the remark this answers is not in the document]
itzg: Good to know. I like it.

src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
harageth: I approve of the boolean field idea

src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
itzg: Any more findings on this one?

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
itzg: BTW, I'm now in process of removing `zoneTenantId` from `BoundMonitor` ...and renaming its `zoneId` to `zoneName` to line up with `name` of `Zone`.

src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
jjbuchan: Not had a chance to look at it yet, and will likely leave it to last due to low priority.

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
jjbuchan: and relating to that previous note, I was second guessing the idea to remove `getByZoneName`.  i'm going to leave it for now and we can chat later if needed.

PR #31, opened 2019-05-14
Public Zone Additions

jjbuchan:
# What

Mostly tweaks and tests involving public monitoring zones.
Has some other indentation changes... my IntelliJ went weird after undoing it's auto-indentation and wouldn't let me apply it again, so after this is merged I'll try do a sweep of files to get all the indentation correct.

# How

 * Changed `envoyTimeout` in the Zone to be `pollerTimeout` to make it consistent with customer facing fields.
 * Allow underscores in zone names so that customer can use that to namespace things if they wish.
 * Added some `count` repository methods to limit the amount of data passed between services where possible
 * Separate methods/objects for dealing with public/private zones when it comes to both creation and deletion.
 * Admin api endpoints start with `/admin`, `public` ones start with `/tenant`
 * Public zones default to `INACTIVE`
 * Private zones default to `ACTIVE`
 * Public zones cannot have null provider etc. fields
 * Various tests

## How to test

Run all tests

# Why

Allows us to interact with public zones correctly via the service api.  Things are handled differently for public zones vs. private zones.

# TODO

1. An indentation PR sweep
1. The admin api should be able to see the `state` of a zone, but the public api should not.  We need to figure out how we should "filter" certain fields based on what API requests the info.  I looked into `JsonFilter`, `JsonViews`, and some other things along those lines but none of them worked out well.  I really want to avoid creating a different DTO for each api response.
1. Public API PR will be created shortly.
1. Admin API is still using GraphQL so I ignored it for now.
1. The difference between the two methods below still baffles me.  They both log the same SQL but the first returns 0 results.  I realized my previous test was incorrect which allowed the first one to pass.  It now fails the test if used.  It happens with other `Contains` methods too.
```
List<Monitor> findByTenantIdAndZonesContains(String tenantId, String zone);

@Query("select m from Monitor m where m.tenantId = :tenantId and :zone member of m.zones")
List<Monitor> findByTenantIdAndZonesContains(String tenantId, String zone);
```

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
@@ -34,89 +34,89 @@ import java.util.UUID;
 
 @Entity
 @Table(name = "zones", uniqueConstraints = {
-        @UniqueConstraint(columnNames = {"tenant_id", "name"})})
+    @UniqueConstraint(columnNames = {"tenant_id", "name"})})
 
 @Data
 public class Zone implements Serializable {
-    @Id
-    @GeneratedValue
-    @Type(type="uuid-char")
-    UUID id;
+  @Id
+  @GeneratedValue
+  @Type(type="uuid-char")
+  UUID id;
 
-    /**
-     * Contains the unique name/label for the zone.
-     * Public zones should have a "public/" prefix and then a trailing region descriptor.
-     * e.g. "public/us-central-1"
-     */
-    @NotBlank
-    @Column(name="name")
-    String name;
+  /**
+   * Contains the tenant that owns the private zone or "_PUBLIC_" for public zones.
+   */
+  @NotBlank
+  @Column(name="tenant_id")
+  String tenantId;
 
-    /**
-     * Contains an optional hosting provider of the zone.
-     * e.g. Rackspace, Google, Amazon
-     */
-    @Column(name="provider")
-    String provider;
+  /**
+   * Contains the unique name/label for the zone.
+   * Public zones should have a "public/" prefix and then a trailing region descriptor.
+   * e.g. "public/us-central-1"
+   */
+  @NotBlank
+  @Column(name="name")
+  String name;
 
-    /**
-     * Contains an optional region to more precisely define where the zone is running.
-     * e.g. for Rackspace, dfw3 may be used; for Google, europe-west3.
-     */
-    @Column(name="provider_region")
-    String providerRegion;
+  /**
+   * Contains an optional hosting provider of the zone.
+   * e.g. Rackspace, Google, Amazon
+   */
+  @Column(name="provider")
+  String provider;
 
-    /**
-     * Defines whether a zone is public or private.
-     */
-    @NotNull
-    @Column(name="is_public")
-    boolean isPublic;
+  /**
+   * Contains an optional region to more precisely define where the zone is running.
+   * e.g. for Rackspace, dfw3 may be used; for Google, europe-west3.
+   */
+  @Column(name="provider_region")
+  String providerRegion;
 
-    /**
-     * Contains an optional list of ipv4 and ipv6 ranges that the pollers in this zone reside in.
-     * This can be used to whitelist ranges to allow for remote polling.
-     */
-    @ElementCollection(fetch = FetchType.EAGER)
-    @Column(name="source_ips")
-    List<String> sourceIpRanges;
+  /**
+   * Defines whether a zone is public or private.
+   */
+  @NotNull
+  @Column(name="is_public")
+  boolean isPublic;
 
-    /**
-     * Defines whether the zone is available or not.
-     * This helps determine whether new monitors can be added to the zone.
-     */
-    @Enumerated(EnumType.STRING)
-    @Column(name="state")
-    ZoneState state;
+  /**
+   * Contains an optional list of ipv4 and ipv6 ranges that the pollers in this zone reside in.
+   * This can be used to whitelist ranges to allow for remote polling.
+   */
+  @ElementCollection(fetch = FetchType.EAGER)
+  @Column(name="source_ips")
+  List<String> sourceIpAddresses;
 
-    /**
-     * Contains the tenant that owns the private zone or "_PUBLIC_" for public zones.
-     */
-    @NotBlank
-    @Column(name="tenant_id")
-    String tenantId;
+  /**
+   * Defines whether the zone is available or not.
+   * This helps determine whether new monitors can be added to the zone.
+   */
+  @Enumerated(EnumType.STRING)
+  @Column(name="state")
+  ZoneState state;
 
-    /**
-     * Contains a timeout value which begins to countdown after an envoy/poller has disconnected.
-     * If the timeout is met, the monitors bound to it will be distributed to other available pollers.
-     */
-    @DurationUnit(ChronoUnit.SECONDS)
-    @Column(name="envoy_timeout")
-    Duration envoyTimeout = Duration.ofSeconds(120);
+  /**
+   * Contains a timeout value which begins to countdown after an poller has disconnected.
+   * If the timeout is met, the monitors bound to it will be distributed to other available pollers.
+   */
+  @DurationUnit(ChronoUnit.SECONDS)
+  @Column(name="poller_timeout")
+  Duration pollerTimeout = Duration.ofSeconds(120);
 
-    /**
-     * Converts the Zone object to a ZoneDTO which is a stripped down version to be used
-     * as output in the APIs.
-     *
-     * @return A ZoneDTO with fields mapped from the Zone.
-     */
-    public ZoneDTO toDTO() {
-        return new ZoneDTO()
-                .setName(name)
-                .setEnvoyTimeout(envoyTimeout.getSeconds())
-                .setProvider(provider)
-                .setProviderRegion(providerRegion)
-                .setPublic(isPublic)
-                .setSourceIpRanges(sourceIpRanges);
-    }
+  /**
+   * Converts the Zone object to a ZoneDTO which is a stripped down version to be used
+   * as output in the APIs.
+   *
+   * @return A ZoneDTO with fields mapped from the Zone.
+   */
+  public ZoneDTO toDTO() {
+    return new ZoneDTO()
+        .setName(name)
+        .setPollerTimeout(pollerTimeout.getSeconds())
+        .setProvider(provider)
+        .setProviderRegion(providerRegion)
+        .setPublic(isPublic)
+        .setSourceIpAddresses(sourceIpAddresses);
+  }
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
@@ -97,7 +97,7 @@ public class Zone implements Serializable {
   ZoneState state;
 
   /**
-   * Contains a timeout value which begins to countdown after an poller has disconnected.
+   * Contains a timeout value which begins to countdown after a poller has disconnected.
    * If the timeout is met, the monitors bound to it will be distributed to other available pollers.
    */
   @DurationUnit(ChronoUnit.SECONDS)
--- a/src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
@@ -83,6 +83,7 @@ public class Zone implements Serializable {
   /**
    * Contains an optional list of ipv4 and ipv6 ranges that the pollers in this zone reside in.
    * This can be used to whitelist ranges to allow for remote polling.
+   * Entries must be valid CIDR notation.
    */
   @ElementCollection(fetch = FetchType.EAGER)
   @Column(name="source_ips")
--- a/src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
@@ -30,9 +30,6 @@ public interface MonitorRepository extends PagingAndSortingRepository<Monitor, U
 
     Page<Monitor> findByTenantId(String tenantId, Pageable pageable);
 
-    // this doesn't work in tests but it logs the exact same query as the one below.
-    //List<Monitor> findByTenantIdAndZonesContains(String tenantId, String zone);
-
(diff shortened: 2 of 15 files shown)


src/test/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiControllerTest.java
```diff
@@ -108,11 +156,49 @@ public void testCreateZoneInvalidName() throws Exception {
...
+            .when(zoneManagement).removePrivateZone(any(), any());
+
+        ZoneCreatePrivate create = newZoneCreatePrivate();
+
+        mvc.perform(delete(
+            "/api/tenant/{tenantId}/zones/{name}", "t-1", "z-1")
+            .content(objectMapper.writeValueAsString(create))
+            .contentType(MediaType.APPLICATION_JSON)
+            .characterEncoding(StandardCharsets.UTF_8.name()))
+            .andExpect(status().isBadRequest()) // This should maybe be a 409 conflict.
```
jjbuchan: Note to look at this

src/test/java/com/rackspace/salus/monitor_management/services/ZoneManagementTest.java
```diff
@@ -60,86 +71,151 @@ public void setUp() {
...
+
+        Zone zone = zoneManagement.createPublicZone(create);
+        assertThat(zone.getId(), notNullValue());
+        assertThat(zone.getTenantId(), equalTo(ResolvedZone.PUBLIC));
+        assertThat(zone.getName(), equalTo(create.getName()));
+        assertThat(zone.getProvider(), equalTo(create.getProvider()));
+        assertThat(zone.getProviderRegion(), equalTo(create.getProviderRegion()));
+        assertThat(zone.getState(), equalTo(ZoneState.INACTIVE));
+        assertThat(zone.getSourceIpAddresses(), hasSize(create.getSourceIpAddresses().size()));
+        //assertThat(zone.getSourceIpAddresses(), equalTo(create.getSourceIpAddresses()));// not sure why this isn't working. Looks fine when inspecting.
```
jjbuchan: This is an odd one.  It looks like it should match, but doesn't.

src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
```diff
@@ -30,9 +30,12 @@
 
     Page<Monitor> findByTenantId(String tenantId, Pageable pageable);
 
-    // this works in tests but not in reality.
+    @Query("select m from Monitor m where m.tenantId = :tenantId and :zone member of m.zones")
```
jjbuchan: I removed the auto-generated query version, but still don't know why it fails.  It outputs the exact same sql in the logs as this one does.

itzg:
I definitely like the idea of an indentation-only PR sweep. I keep being tempted to fix a particular file I'm editing, but the mass-whitespace would dilute the real diff.

src/test/java/com/rackspace/salus/monitor_management/services/ZoneManagementTest.java
itzg: Double check that it's not just an ordering difference. I'm still thinking that `equalTo` is order sensitive with collections even though the javadoc are a little ambiguous with description of **array** handling:

http://hamcrest.org/JavaHamcrest/javadoc/1.3/org/hamcrest/Matchers.html#equalTo(T)

src/test/java/com/rackspace/salus/monitor_management/services/ZoneManagementTest.java
```diff
@@ -154,11 +230,11 @@ public void testGetAvailableZonesForTenant() {
         assertThat(zoneManagement.getAvailableZonesForTenant(tenant), hasSize(1));
 
         // any new private zone for the tenant should be visible
-        createZonesForTenant(privateCount, tenant);
+        createPrivateZonesForTenant(privateCount, tenant);
         assertThat(zoneManagement.getAvailableZonesForTenant(tenant), hasSize(1 + privateCount));
 
         // new public zones should be visible too
-        createZonesForTenant(publicCount, ResolvedZone.PUBLIC);
+        createPrivateZonesForTenant(publicCount, ResolvedZone.PUBLIC);
```
itzg: Helper method name gets a little awkward here :)

src/test/java/com/rackspace/salus/monitor_management/services/ZoneManagementTest.java
```diff
@@ -177,4 +253,83 @@ public void testGetMonitorsForZone() {
...
+        int publicCount = 17;
+        String tenant = RandomStringUtils.randomAlphabetic(10);
+        String privateZone = RandomStringUtils.randomAlphabetic(10);
+        String publicZone = ResolvedZone.PUBLIC_PREFIX + RandomStringUtils.randomAlphabetic(6);
+
+        assertThat(zoneManagement.getMonitorCountForPrivateZone(tenant, privateZone), equalTo(0));
+        assertThat(zoneManagement.getMonitorCountForPublicZone(publicZone), equalTo(0));
+
+        createRemoteMonitorsForTenant(privateCount, tenant, privateZone);
+        createRemoteMonitorsForTenant(privateCount, tenant, "anotherPrivateZone");
```
itzg: Just to avoid accidentally picking up "anotherPrivateZone" and asserting `privateCount`, perhaps create `privateCount+1` of these.

src/main/java/com/rackspace/salus/monitor_management/web/model/ZoneDTO.java
```diff
@@ -15,16 +15,19 @@
...
     String name;
-    long envoyTimeout;
+    long pollerTimeout;
     String provider;
     String providerRegion;
     boolean isPublic;
-    List<String> sourceIpRanges;
+    List<String> sourceIpAddresses;
+
+    ZoneState state; // we want to filter this from public api responses
```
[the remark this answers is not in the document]
harageth: I just realized that whatever we do here is almost certainly going to require some changes to my documentation converter.

src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
```diff
@@ -34,89 +34,89 @@
...
-     * Contains the tenant that owns the private zone or "_PUBLIC_" for public zones.
-     */
-    @NotBlank
-    @Column(name="tenant_id")
-    String tenantId;
+  /**
+   * Defines whether the zone is available or not.
+   * This helps determine whether new monitors can be added to the zone.
+   */
+  @Enumerated(EnumType.STRING)
```
harageth: This annotation right here is super handy since database enum's are a huge PITA.

itzg: Yeah, it's annoying the default JPA behavior is to encode the enum's ordinal value.

harageth: I mean its more space efficient and is gonna be a faster lookup. Plus its closer to what the Database does behind the scenes (basically a numeric mapping). But yeah it would be much nicer to have them stored as Strings by default.

src/main/java/com/rackspace/salus/monitor_management/repositories/MonitorRepository.java
harageth: If it's the same SQL then yeah that doesnt make sense why it was failing.

src/test/java/com/rackspace/salus/monitor_management/services/ZoneManagementTest.java
jjbuchan: Ordering looked fine as well.  Maybe I'll throw these in a loop to see if that brings out more.

I don't remember those docs stating `in the same indexes` before.  I was pretty sure order was irrelevant.

src/main/java/com/rackspace/salus/monitor_management/entities/Zone.java
```diff
@@ -34,89 +34,89 @@
...
-    @Enumerated(EnumType.STRING)
-    @Column(name="state")
-    ZoneState state;
+  /**
+   * Contains an optional list of ipv4 and ipv6 ranges that the pollers in this zone reside in.
+   * This can be used to whitelist ranges to allow for remote polling.
+   */
+  @ElementCollection(fetch = FetchType.EAGER)
+  @Column(name="source_ips")
+  List<String> sourceIpAddresses;
```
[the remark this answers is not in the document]
jjbuchan: Having that validation crossed my mind before... then I forgot.  I'll see how we can add that on here.

src/main/java/com/rackspace/salus/monitor_management/services/ZoneManagement.java
```diff
@@ -43,70 +46,206 @@
...
-                .setName(newZone.getName())
-                .setEnvoyTimeout(Duration.ofSeconds(newZone.getPollerTimeout()));
+            .setTenantId(tenantId)
+            .setName(newZone.getName())
+            .setProvider(newZone.getProvider())
+            .setProviderRegion(newZone.getProviderRegion())
+            .setSourceIpAddresses(newZone.getSourceIpAddresses())
+            .setPollerTimeout(Duration.ofSeconds(newZone.getPollerTimeout()))
+            .setState(newZone.getState())
+            .setPublic(false);
```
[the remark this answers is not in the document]
jjbuchan: In the api responses it also trims of the `is` and the field just comes back as `public: true`

src/main/java/com/rackspace/salus/monitor_management/services/ZoneManagement.java
```diff
@@ -43,70 +46,206 @@
...
-                        name, tenantId)));
+  /**
+   * Helper method to update a private zone's details.
+   * @param tenantId The tenant of the zone.
+   * @param name The name of the zone.
+   * @param updatedZone The zone parameters to update.
+   * @return The newly updated zone.
+   */
+  public Zone updatePrivateZone(String tenantId, String name, @Valid ZoneUpdate updatedZone) {
+    Zone zone = getZone(tenantId, name).orElseThrow(() ->
```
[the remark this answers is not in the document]
jjbuchan: Ah yes.  Good idea.

src/main/java/com/rackspace/salus/monitor_management/web/model/ZoneCreatePublic.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.validation.constraints.Max;
+import javax.validation.constraints.Min;
+import org.hibernate.validator.constraints.NotBlank;
+import java.io.Serializable;
+
+@Data
+public class ZoneCreatePublic implements Serializable {
+
+  @NotBlank
+  String name;
```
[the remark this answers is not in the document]
jjbuchan: That crossed my mind too.  I'll add it.

src/main/java/com/rackspace/salus/monitor_management/web/model/ZoneDTO.java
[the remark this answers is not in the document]
jjbuchan: > Could pass a flag to toDTO that indicates public/admin, set state to null if public, and then add a @JsonInclude(NON_NULL) here.

I had similar ideas.  Using this method might work for the `state` but we'd hit different scenarios in other objects that it wouldn't work for.  I was hoping for a solution that would handle them all well.

src/main/java/com/rackspace/salus/monitor_management/services/ZoneManagement.java
```diff
@@ -43,70 +46,206 @@
...
+      int monitors = getMonitorCountForPrivateZone(tenantId, name);
+      if(monitors > 0) {
+        throw new IllegalArgumentException(
+            String.format("Cannot remove zone with configured monitors. Found %s.", monitors));
+      }
+      removeZone(zone);
+    }
+
+    /**
+     * Helper method to public private zones by tenant id and zone name.
```
harageth: Either the function name needs to change or this description needs to change. But the function appears to do what it says.

harageth:
```suggestion
     * Helper method to remove public zones by zone name.
```

src/main/java/com/rackspace/salus/monitor_management/web/model/ZoneCreatePublic.java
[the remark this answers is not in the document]
harageth: It would force us to consistency rather than consistency through policy.

jjbuchan:
New changes pushed.  There will be a PR for salus-common that pulls in https://github.com/edazdarevic/CIDRUtils/blob/master/CIDRUtils.java

src/test/java/com/rackspace/salus/monitor_management/services/ZoneManagementTest.java
jjbuchan: Didn't get to this... I'll check again after standup.

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
```diff
@@ -54,28 +55,54 @@ public ZoneApiController(ZoneManagement zoneManagement) {
...
     public ZoneDTO getByZoneName(@PathVariable String tenantId, @PathVariable String name) {
-        Optional<Zone> zone = zoneManagement.getZone(tenantId, name);
-        return zone.orElseThrow(() -> new NotFoundException(String.format("No zone found for %s on tenant %s",
+        Optional<Zone> zone = zoneManagement.getPrivateZone(tenantId, name);
+        return zone.orElseThrow(() -> new NotFoundException(String.format("No zone found named %s on tenant %s",
                 name, tenantId)))
                 .toDTO();
     }
 
+    @GetMapping("/admin/zones/{name}")
```
harageth: Would you mind adding the documentation annotations on these so that we have the documentation already?

You would basically need something like this at the top https://github.com/racker/salus-telemetry-monitor-management/blob/master/src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java#L62-L69

And something like this on each endpoint https://github.com/racker/salus-telemetry-monitor-management/blob/master/src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java#L86

jjbuchan: I think I'll do this in a PR directly after this gets merged.  I don't wanna slow @itzg down anymore for getting the demo ready.  I did feel guilty leaving it out though.

harageth: I am good with that

PR #32, opened 2019-05-14
Add distribution management to pom

dgvigil:
# Resolves
SALUS-334

# What

Correcting the repos that are in Artifactory.
Updated cloudbuild files to ensure only deploying in specific cases, otherwise just doing a `mvn verify`

Also ensured there are 3 cloudbuild triggers (tags, master = cloudbuild-deploy.yaml, any branch = cloudbuild.yaml)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,10 +151,6 @@
         </executions>
       </plugin>
       <plugin>
-        <groupId>pl.project13.maven</groupId>
-        <artifactId>git-commit-id-plugin</artifactId>
-      </plugin>
-      <plugin>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -245,5 +241,47 @@
     </plugins>
   </build>
 
+  <distributionManagement>
```
itzg: I'm actually not sure if these app poms need this called out since they use `salus-app-base` as their parent and probably inherit from that.

dgvigil: While it should inherit, I made it explicit in the pom.xml to avoid any hint of ambiguity.

PR #33, opened 2019-05-14
Add doc annotations to zone api

jjbuchan:
Adds the basic swagger annotation to the zone api.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -188,6 +188,7 @@
               <locations>
                 <location>com.rackspace.salus.monitor_management.web.model</location>
                 <location>com.rackspace.salus.monitor_management.web.controller.MonitorApiController</location>
+                <location>com.rackspace.salus.monitor_management.web.controller.ZoneApiController</location>
               </locations>
               <templatePath>${basedir}/templates/strapdown.html.hbs</templatePath>
               <schemes>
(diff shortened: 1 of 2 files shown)


harageth:
I lied you do need to make a pom change to tell it to also look at the ZoneApiController

https://github.com/racker/salus-telemetry-monitor-management/blob/master/pom.xml#L190

jjbuchan:
@harageth added.

I wonder how things like https://github.com/racker/salus-telemetry-monitor-management/tree/master/src/main/java/com/rackspace/salus/monitor_management/web/model/validator will show up since it's under `model`.  Maybe that's something we'll have to filter out later.

harageth:
> @harageth added.
> 
> I wonder how things like https://github.com/racker/salus-telemetry-monitor-management/tree/master/src/main/java/com/rackspace/salus/monitor_management/web/model/validator will show up since it's under `model`. Maybe that's something we'll have to filter out later.

It didn't show up when I just generated it. 

Having this as part of the same project does interpolate the zone endpoints in the middle of the monitor endpoints. Probably not a huge deal right now, but it would be nice to have them all grouped together at some point in the future.

harageth:
Oh and now that both your and my PR's are merged we have two different locations for the validators. I didn't put mine in the `model` package because I didn't feel it fit in there.

PR #36, opened 2019-05-15
Add API model filtering

jjbuchan:
# What

Adds the ability to filter what fields of a model are returned via each API type (public, internal, admin).

Currently it only takes effect on `ZoneDTO` objects.

# How

Using JsonViews this is fairly simple.

By adding `spring.jackson.mapper.default-view-inclusion=true` it means any model field without a view set with automatically be included.  This means we do not have to set the `Public` view on every field we want to be in there, we can just leave them blank.

For any other field we can restrict access by adding the annotation for the relevant view.  The Internal view automatically pulls in any field that is Public in addition to those tagged as Internal and the Admin view pulls in any field that is in the Public or Internal view as well as any Admin specific fields.

We do however have to annotate all the api endpoint mappings with a view otherwise everything will be returned regardless of the field annotations.

## How to test

I added tests specific for this (the ZoneDTOTest should help make things clear) and manually tested.

# Why

Some fields should be customer visibile, some should not.  This allows us to filter what people see via annotations vs. having to create different objects for every scenario.

# TODO

 * Maybe `View.java` should go in the salus model repo?  Put it in here so you could see the full picture first.
 * Do we want to consider changing how MonitorDTO/DetailedMonitorOutput works and try to remove one of them?


itzg:
To answer the question about `MonitorDTO`, I don't think we really need it since the only usage

image

is `getMonitorsForZone` in `ZoneApi` and nothing is even using that method:

image

src/main/java/com/rackspace/salus/monitor_management/web/model/View.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.rackspace.salus.monitor_management.web.model;
+
+public class View {
```
itzg: I agree with your comment that we can later move this over to the model repo.

src/test/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiControllerTest.java
```diff
@@ -82,7 +83,9 @@ public void testGetByZoneName() throws Exception {
                 .andExpect(status().isOk())
                 .andExpect(content()
                         .contentTypeCompatibleWith(MediaType.APPLICATION_JSON))
-                .andExpect(content().json(objectMapper.writeValueAsString(expectedZone.toDTO())));
+                .andExpect(content().json(objectMapper
```
itzg: In this case it might be better to be comparing against a json test data file. Exercising the same view option is good to know is an option, but might mask any problems if we incorrectly annotate something.

FYI here's my cheat sheet for loading expected text content into a string:
https://gist.github.com/itzg/f2863492b85a7673b892cc48771a2a12

src/test/java/com/rackspace/salus/monitor_management/web/model/ZoneDTOTest.java
```diff
@@ -49,5 +53,21 @@ public void testFieldsCovered() {
     assertThat(dto.getSourceIpAddresses(), equalTo(zone.getSourceIpAddresses()));
     assertThat(dto.getState(), equalTo(zone.getState()));
     assertThat(dto.isPublic(), equalTo(zone.isPublic()));
+
+
+    String objectAsString;
+    ZoneDTO convertedDto;
+
+    objectAsString = objectMapper.writerWithView(View.Public.class).writeValueAsString(dto);
```
itzg: Here the use of `writerWithView` makes sense.

src/main/java/com/rackspace/salus/monitor_management/web/model/View.java
jjbuchan: https://github.com/racker/salus-telemetry-model/pull/51

src/test/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiControllerTest.java
jjbuchan: Agreed.  I've updated them.

PR #37, opened 2019-05-17
Handle public zones in GET zone endpoint

jjbuchan:
# What

Adds the ability to handle requests for public zones by a customer, such as:
```
GET http://localhost:8089/api/tenant/aaaaaa/zones/public/us-east
```

# How

Rather than using `@PathVariable` to extract the zone, which won't work with values containing `/` we can match on whatever string is after `.../zones/` and use that.

## How to test

Run tests. Try manually.

# Why

I don't think we will be adding any other endpoints after `.../zones/{id}` so we should be good to do that.

it means we can keep that endpoint and don't have to worry about it failing for customers who try to follow the same method they do for private zones.  It also means if we want to use it ourselves we don't have the url encode the zone name.

# TODO

I wont merge this until after the `JsonViews` changes as I think I might have to move things around since `getPublicZone` would likely return the `state` which we do not want to do.  Just wanted to put this out there for now.

We should probably also modify the name of `getByZoneName` but since that's an interface change I didn't want to do it here.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
@@ -72,6 +72,7 @@ public class ZoneApiController implements ZoneApi {
 
     @GetMapping("/tenant/{tenantId}/zones/**")
     @ApiOperation(value = "Gets specific zone by tenant id and zone name")
+    @JsonView(View.Public.class)
     public ZoneDTO getAvailableZone(@PathVariable String tenantId, HttpServletRequest request) {
       String name = extractZoneNameFromUri(request);
       if (name.startsWith(ResolvedZone.PUBLIC_PREFIX)) {
@@ -129,6 +130,7 @@ public class ZoneApiController implements ZoneApi {
     @PostMapping("/admin/zones")
     @ResponseStatus(HttpStatus.CREATED)
     @ApiOperation(value = "Creates a new public zone")
+    @JsonView(View.Admin.class)
     public ZoneDTO create(@Valid @RequestBody ZoneCreatePublic zone)
         throws ZoneAlreadyExists {
         return zoneManagement.createPublicZone(zone).toDTO();
--- a/src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
@@ -76,7 +76,8 @@ public class ZoneApiController implements ZoneApi {
     public ZoneDTO getAvailableZone(@PathVariable String tenantId, HttpServletRequest request) {
       String name = extractZoneNameFromUri(request);
       if (name.startsWith(ResolvedZone.PUBLIC_PREFIX)) {
-        return getPublicZone(request); // probably have to change this when json views are merged.
+        // The JsonView on getPublicZone is ignored
+        return getPublicZone(request);
       }
       return getByZoneName(tenantId, name);
     }
--- a/src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
@@ -64,125 +64,123 @@ import org.springframework.web.servlet.HandlerMapping;
         })
 })
 public class ZoneApiController implements ZoneApi {
-    private ZoneManagement zoneManagement;
+  private ZoneManagement zoneManagement;
 
-    public ZoneApiController(ZoneManagement zoneManagement) {
-        this.zoneManagement = zoneManagement;
-    }
+  public ZoneApiController(ZoneManagement zoneManagement) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
```diff
@@ -64,23 +68,48 @@ public ZoneApiController(ZoneManagement zoneManagement) {
...
+      String name = extractZoneNameFromUri(request);
+      if (name.startsWith(ResolvedZone.PUBLIC_PREFIX)) {
+        return getPublicZone(request); // probably have to change this when json views are merged.
+      }
+      return getByZoneName(tenantId, name);
+    }
+
+    @Override
+    public ZoneDTO getByZoneName(String tenantId, String name) {
+      Optional<Zone> zone = zoneManagement.getPrivateZone(tenantId, name);
```
itzg: If you moved the `if (name.startsWith(ResolvedZone.PUBLIC_PREFIX)) {` it might give this API overriding method a little more purpose on the controller side. Then the ant-matching `getAvailableZone` is just a thin wrapper around `getByZoneName`

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
```diff
@@ -64,23 +68,48 @@ public ZoneApiController(ZoneManagement zoneManagement) {
...
+    }
+
+    /**
+     * Discovers the zone name within the uri.
+     * Handles both public zones containing slashes as well as more simple private zone.
+     * Using @PathVariable does not work for public zones.
+     * @param request The incoming http request.
+     * @return The zone name provided within the uri.
+     */
+    private String extractZoneNameFromUri(HttpServletRequest request) {
```
itzg: Very cool

src/test/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiControllerTest.java
```diff
@@ -85,6 +85,22 @@ public void testGetByZoneName() throws Exception {
...
     }
 
+    @Test
+    public void testGetAvailablePublicZoneForTenant() throws Exception {
+        final Zone expectedZone = podamFactory.manufacturePojo(Zone.class);
+        when(zoneManagement.getPublicZone(any()))
+            .thenReturn(Optional.of(expectedZone));
+
+        mvc.perform(get(
+            "/api/admin/zones/{name}", "public/z-1")
```
itzg: I'm wondering if it's going to url-encode this path part. I think a Mockito `verify` of the `zoneManagement.getPublicZone` could assert that.

itzg:
Oh, I guess the same kind of logic will be needed in the `api` module's public proxy stuff. SImilar to my unit test comment, we'll need to be careful that it doesn't try to URL encode the zone name part.

itzg:
@jjbuchan I was about to make some API changes and noticed this code was "missing"...but then found the PR is still in flight. Do we still need this for the "public/..." style naming of public zones? I'm now wondering if we should just prepend "public/" onto the zone name given in the admin API calls.

jjbuchan:
> Do we still need this for the "public/..." style naming of public zones?

Yep

> I'm now wondering if we should just prepend "public/" onto the zone name given in the admin API calls

This is mainly so customers could do a GET on the zones for their private one or a public one and they'd all act the same.

> the PR is still in flight

I can try get to this later today.  The json views stuff got merged, so this just needs to be rebased and likely move some things around a little bit.

jjbuchan:
> I'm now wondering if we should just prepend "public/" onto the zone name given in the admin API calls

I think adding validation that the provided zone starts with `public/` might be good.  If we auto prepend public we might end up creating `public/public/zoneName`.

We could validate and then prepend if it's not there... but I like the consistency of it creating whatever you specify

jjbuchan:
@itzg I've adjusted some things.  Ended up removing the different error messages for public vs. private zone so we could consolidate more into `getByZoneName`.  Indentation was bugging me so I ended up doing it here for one final - I'll still aim to follow up with another PR for all the others.

Thought it might be best to push this as is and I'll worry about the encoding questions you had after lunch.  I didn't want to hold your other PR up too much as I dig into that.

src/test/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiControllerTest.java
```diff
@@ -100,11 +100,38 @@ public void testGetAvailablePrivateZoneForTenant() throws Exception {
...
+
+        mvc.perform(get(
+            "/api/tenant/{tenantId}/zones/{name}", "t-1", "public/zone-1")
+            .contentType(MediaType.APPLICATION_JSON))
+            .andDo(print())
+            .andExpect(status().isOk())
+            .andExpect(content()
+                .contentTypeCompatibleWith(MediaType.APPLICATION_JSON))
+            .andExpect(content().json(
+                // the STATE field should not be included
```
itzg: ah, definitely calls for the use of strict json checking

PR #38, opened 2019-05-21
Handle ReattachEnvoyResourceEvent and re-bind local monitors

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-360

# What

Consumes `ReattachedEnvoyResourceEvent` event introduced by https://github.com/racker/salus-telemetry-model/pull/52 , locates **local** monitors bound to the tenant's resource, and re-binds those to the newly attached Envoy. It then sends out monitor bind change events for the new envoy ID and just in case the old envoys are lingering, it sends change events for those too.

Design background is provided in https://github.com/racker/salus-telemetry-bundle/pull/50

## How to test

Unit tests updated.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,6 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>21.0</version>
     </dependency>
     <dependency>
       <groupId>com.rackspace.salus</groupId>
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -33,6 +33,7 @@ import com.rackspace.salus.telemetry.etcd.services.EnvoyResourceManagement;
 import com.rackspace.salus.telemetry.etcd.services.ZoneStorage;
 import com.rackspace.salus.telemetry.etcd.types.ResolvedZone;
 import com.rackspace.salus.telemetry.messaging.MonitorBoundEvent;
+import com.rackspace.salus.telemetry.messaging.ReattachedEnvoyResourceEvent;
 import com.rackspace.salus.telemetry.messaging.ResourceEvent;
 import com.rackspace.salus.telemetry.model.ConfigSelectorScope;
 import com.rackspace.salus.telemetry.model.NotFoundException;
@@ -639,7 +640,7 @@ public class MonitorManagement {
      *
      * @param event the new resource event.
      */
-    void handleResourceEvent(ResourceEvent event) {
+    void handleResourceChangeEvent(ResourceEvent event) {
         final String tenantId = event.getTenantId();
         final String resourceId = event.getResourceId();
 
@@ -682,6 +683,37 @@ public class MonitorManagement {
         }
 
         sendMonitorBoundEvents(changes);
+    }
+
+    /**
+     * Finds all the local monitors bound to the resource and re-bind them to the newly attached
+     * Envoy
+     * @param resourceEvent the event indicating the re-attached envoy-resource
+     */
+    void handleReattachedEnvoy(ReattachedEnvoyResourceEvent resourceEvent) {
+        final List<BoundMonitor> bound = boundMonitorRepository
+            .findAllLocalByTenantResource(
+                resourceEvent.getTenantId(),
+                resourceEvent.getResourceId()
+            );
+
+        final List<String> previousEnvoyIds = bound.stream()
+            .map(BoundMonitor::getEnvoyId)
+            .filter(Objects::nonNull)
+            .distinct()
+            .collect(Collectors.toList());
+
+        bound.forEach(boundMonitor ->
+            boundMonitor.setEnvoyId(resourceEvent.getEnvoyId())
+        );
+
+        boundMonitorRepository.saveAll(bound);
+
+        // now that the re-binding is saved
+        // ...tell any previous envoys about loss of binding
+        previousEnvoyIds.forEach(this::sendMonitorBoundEvent);
+        // ...and tell the attached envoy about the re-bindings
+        sendMonitorBoundEvent(resourceEvent.getEnvoyId());
     }
 
     List<BoundMonitor> upsertBindingToResource(List<Monitor> monitors,
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -717,11 +717,7 @@ public class MonitorManagement {
                 resourceId
             );
 
-        final List<String> previousEnvoyIds = bound.stream()
-            .map(BoundMonitor::getEnvoyId)
-            .filter(Objects::nonNull)
-            .distinct()
-            .collect(Collectors.toList());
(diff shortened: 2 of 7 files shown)


jjbuchan:
Going to hold off any any comments here until we chat about https://github.com/racker/salus-telemetry-resource-management/pull/31

itzg:
@jjbuchan ready for re-review

pom.xml
```diff
@@ -69,7 +69,6 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <version>21.0</version>
```
itzg: Thanks to https://github.com/racker/salus-app-base/pull/12

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -717,33 +778,50 @@ void handleResourceEvent(ResourceEvent event) {
...
+                        existingBind.setRenderedContent(newRenderedContent);
+                        boundMonitors.add(existingBind);
+                    } else if (
+                        reattachedEnvoyId != null &&
+                            monitor.getSelectorScope() == ConfigSelectorScope.LOCAL &&
+                            existingBind.getEnvoyId() != null
+                    ) {
+                        // need to send an event to old Envoy just in case it's around, but
+                        // probably won't be due to the re-attachment
+                        affectedEnvoys.add(existingBind.getEnvoyId());
```
itzg: FYI, this was the line that induces the refactoring from methods returning changed `BoundMonitors` to affected envoy IDs. It turned out the return values were only being used to compute the distinct envoy IDs to notify.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -655,6 +671,11 @@ void handleResourceEvent(ResourceEvent event) {
         if (resource != null) {
             // resource created or updated
 
+            if (event.isDeleted()) {
```
jjbuchan: It looks like you added this field for this one case?  Is there something I'm missing?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -717,33 +778,50 @@ void handleResourceEvent(ResourceEvent event) {
...
-                        .collect(Collectors.toList())
-                );
+                for (BoundMonitor existingBind : existing) {
+                    if (!existingBind.getRenderedContent().equals(newRenderedContent)) {
+                        existingBind.setRenderedContent(newRenderedContent);
+                        boundMonitors.add(existingBind);
+                    } else if (
+                        reattachedEnvoyId != null &&
+                            monitor.getSelectorScope() == ConfigSelectorScope.LOCAL &&
+                            existingBind.getEnvoyId() != null
```
jjbuchan: `existingBind.getEnvoyId() != null`

I'm not sure why this matters?

I also might be getting confused... remote monitors will have a `null` envoyId if there is no envoy connected / the monitor is not actively being run.  Is that the same case for local monitors?

I would expect it to still be `null` for local monitors if it had been assigned to an envoy in the past but that envoy has since been disconnected for a while.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -717,33 +778,50 @@ void handleResourceEvent(ResourceEvent event) {
...
-                        // for those that changed, modify entity
-                        .peek(existingBind -> existingBind.setRenderedContent(newRenderedContent))
-                        // and add all of these to list to save and return
-                        .collect(Collectors.toList())
-                );
+                for (BoundMonitor existingBind : existing) {
+                    if (!existingBind.getRenderedContent().equals(newRenderedContent)) {
+                        existingBind.setRenderedContent(newRenderedContent);
+                        boundMonitors.add(existingBind);
+                    } else if (
```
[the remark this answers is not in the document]
jjbuchan: A test case for this would be good.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
itzg: Really I just added the `deleted` field for completeness since it was one of three reason the `ResourceEvent` was getting produced.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
itzg: The re-attachment logic is only applicable to local/agent monitors because, as you're finding/implementing right now we've got remote/zone oriented covered.

The code doesn't currently null-out the `envoyId` of `BoundMonitor's when **non-zone** related Envoys go away. We could eventually tap into presence monitor, but I didn't want to add even more complexity at this point.

Perhaps I don't need the `existingBind.getEnvoyId() != null` since local monitors are always assigned to an Envoy, unlike remote monitors that can be in an "unassigned" state when the zone is empty.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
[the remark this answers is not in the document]
itzg: Either way, the affected envoy makes it into the set due to the code further down

```
        affectedEnvoys.addAll(
            extractEnvoyIds(boundMonitors)
        );
```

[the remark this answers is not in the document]
itzg: ...I'll add a unit test since you're right that the two tests near there doesn't exercise exactly that combination.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -671,28 +692,68 @@ void handleResourceEvent(ResourceEvent event) {
...
+            .findAllLocalByTenantResource(
+                tenantId,
+                resourceId
+            );
+
+        final List<String> previousEnvoyIds = bound.stream()
+            .map(BoundMonitor::getEnvoyId)
+            .filter(Objects::nonNull)
+            .distinct()
+            .collect(Collectors.toList());
```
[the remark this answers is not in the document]
itzg: Ah yes, I can now.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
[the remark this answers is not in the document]
jjbuchan: > Either way, the affected envoy makes it into the set due to the code further down

but both `existingBind.set` operations wont happen; only the first?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
jjbuchan: Ok.  I'm kinda indifferent to it.  Just seemed unnecessary.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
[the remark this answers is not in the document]
itzg: Good catch. Adding the unit test revealed the problem you were pointing out.

PR #40, opened 2019-05-27
adds the x509 and http remote monitors

GeorgeJahad:
# Resolves
https://jira.rax.io/browse/SALUS-364
https://jira.rax.io/browse/SALUS-363

# What
adds the x509 and http remote monitors

## How to test

unit tests

# TODO

I wanted to do an end to end test to confirm they correctly get passed from the ambassador to the envoy,  and tried to update the insomnia workspace to config them, but I'm confused by the state of what is configured with graphql and what is configured with insomnia.  No longer sure how to init/config the agents on the envoy.  I'll try to set that up soon.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2019 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.monitor_management.web.model.telegraf;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.rackspace.salus.monitor_management.web.model.ApplicableAgentType;
+import com.rackspace.salus.monitor_management.web.model.RemotePlugin;
+import com.rackspace.salus.telemetry.model.AgentType;
+import java.util.List;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data @EqualsAndHashCode(callSuper = true)
+@ApplicableAgentType(AgentType.TELEGRAF)
+@JsonInclude(Include.NON_NULL)
+public class HttpResponse extends RemotePlugin {
+  List<String> sources;
+  String timeout;
+  String tlsCa;
+  String tlsCert;
+  String tlsKey;
+  boolean insecureSkipVerify;
+}
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
@@ -21,18 +21,25 @@ import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.rackspace.salus.monitor_management.web.model.ApplicableAgentType;
 import com.rackspace.salus.monitor_management.web.model.RemotePlugin;
 import com.rackspace.salus.telemetry.model.AgentType;
-import java.util.List;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
+
+import java.util.Map;
 
 @Data @EqualsAndHashCode(callSuper = true)
 @ApplicableAgentType(AgentType.TELEGRAF)
 @JsonInclude(Include.NON_NULL)
 public class HttpResponse extends RemotePlugin {
-  List<String> sources;
-  String timeout;
+  String address;
+  String httpProxy;
+  String responseTimeout;
+  String method;
+  boolean followRedirects;
+  String body;
+  String responseStringMatch;
   String tlsCa;
   String tlsCert;
   String tlsKey;
   boolean insecureSkipVerify;
+  Map<String, String> headers;
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
@@ -35,7 +35,7 @@ public class HttpResponse extends RemotePlugin {
   @NotEmpty
   String address;
   String httpProxy;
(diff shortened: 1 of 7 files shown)


itzg:
@GeorgeJahad I'm sorry about that -- my environment apparently works because I have an old telegraf install still present. I am working on cleaning up the Insomnia examples and some required code changes in the admin API

https://jira.rax.io/browse/SALUS-370

itzg:
This PR is part of the fix https://github.com/racker/salus-telemetry-api/pull/31

itzg:
This change is helpful https://github.com/racker/salus-telemetry-etcd-adapter/pull/37

itzg:
...and updated shared Insomnia workspace with corrected operations

https://github.com/racker/salus-telemetry-bundle/pull/50/commits/4572dad13fe6406ec6f77965e5e966996618dfbd

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+import java.util.Map;
+
+@Data @EqualsAndHashCode(callSuper = true)
+@ApplicableAgentType(AgentType.TELEGRAF)
+@JsonInclude(Include.NON_NULL)
+public class HttpResponse extends RemotePlugin {
+  String address;
```
itzg: Could add `@NotEmpty` to this since telegraf's default of "http://localhost" isn't good for us.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
```diff
@@ -0,0 +1,45 @@
...
+  String httpProxy;
+  String responseTimeout;
```
itzg: For future enhancement, a custom validator/constraint would be good for Go duration validation here.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
```diff
@@ -0,0 +1,45 @@
...
+  String method;
```
itzg: I wonder if you could use https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/http/HttpMethod.html to constrain the values end users provide.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/X509Cert.java
```diff
@@ -0,0 +1,38 @@
...
+import com.rackspace.salus.telemetry.model.AgentType;
+import java.util.List;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data @EqualsAndHashCode(callSuper = true)
+@ApplicableAgentType(AgentType.TELEGRAF)
+@JsonInclude(Include.NON_NULL)
+public class X509Cert extends RemotePlugin {
+  List<String> sources;
```
itzg: `@NotEmpty` would be good here

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
itzg: Oh, a regex constraint would suffice since the format isn't too complex:

https://golang.org/pkg/time/#ParseDuration

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
jjbuchan: This comment led me to see how this plugin handles hitting the same host on multiple different IPs.  Looks like this field can contain the target ip address and then the host headers specified in the Headers field.  I like that the host field is optional so for most cases only the address needs to be set.

I never liked how ele handled all these fields.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
```diff
@@ -0,0 +1,45 @@
...
+  boolean followRedirects;
+  String body;
+  String responseStringMatch;
+  String tlsCa;
+  String tlsCert;
+  String tlsKey;
+  boolean insecureSkipVerify;
+  Map<String, String> headers;
```
jjbuchan: > A value which is not one of: content-length, user-agent, host, connection, keep-alive, transfer-encoding, upgrade.

This is what ele restricts.  `host` should not be restricted anymore, but maybe the others should be?

jjbuchan: It's easier to restrict them earlier than later, so might be best to start here.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/X509Cert.java
```diff
@@ -0,0 +1,38 @@
...
+import com.rackspace.salus.monitor_management.web.model.RemotePlugin;
+import com.rackspace.salus.telemetry.model.AgentType;
+import java.util.List;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data @EqualsAndHashCode(callSuper = true)
+@ApplicableAgentType(AgentType.TELEGRAF)
+@JsonInclude(Include.NON_NULL)
+public class X509Cert extends RemotePlugin {
```
jjbuchan: This plugin seems like it could be local or remote depending on how you configure it.

We'd probably need to prevent customers from specifying local `sources` if they used it on public pollers?

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
GeorgeJahad: is something like this acceptable:
```
  @Pattern(regexp = "GET|PUT|POST|DELETE|HEAD|OPTIONS|PATCH|TRACE", message = "invalid http method")
```

itzg: Works for me

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/X509Cert.java
itzg: Good point. I guess we could also declare an `X509CertLocal` that extends `LocalPlugin`, but that assumes we'd also have an appropriate constraint on each of their `sources`.

jjbuchan: This will be added via a new jira issue.  Customer's will be able to dictate whether they use `local` or `remote` via the existing api operations and depending on which they choose we will have different validations act on the `sources`.

GeorgeJahad:
here are the changes requested by @itzg:

https://github.com/racker/salus-telemetry-monitor-management/compare/3d6d038...2af9803

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/X509Cert.java
GeorgeJahad: here is the jira issue:
https://jira.rax.io/browse/SALUS-371

PR #41, opened 2019-05-29
Provide zone rebalancing operation

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-294

# What

Active Envoys within a zone are assigned new bound monitors in a least-loaded way; however, if new Envoys are added later then they will be potentially left unutilized or underutilized.

# How

The rebalancing operation acts upon a given public/private zone. It will gather the assignment counts of all the envoy-resources in the zone and calculate the statistics over those counts. Specifically, the average (aka mean) and standard deviation calculations are of primary interest.

The algorithm looks for envoys that are outliers by computing a threshold that is

    threshold = avg + rebalanceStandardDeviations * stddev

Any envoys that have more assigned that `threshold` will have enough bound monitors unassigned such that it would end up with `avg` bound monitors assigned. After all of the un-assignments are executed, the pre-existing `handleNewEnvoyInZone` is used to trigger the re-assignment of those newly unassigned bindings to the least-loaded Envoys. In the case where new Envoys are active in the zone, the idea is that those Envoys are the ones that get the new assignments. 

**NOTE** more analysis might be needed to ensure that we don't end up with two clumps of Envoys: non-zero, but balanced and zero-assigned Envoys.

## How to test

Unit tests were updated and added.

## Requires

https://github.com/racker/salus-telemetry-etcd-adapter/pull/38

## TODO

- [x] After https://github.com/racker/salus-telemetry-monitor-management/pull/37 is merged, the APIs need to be updated to use the ant-pattern logic to allow for zone names with slashes.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -993,20 +993,27 @@ public class MonitorManagement {
                     .collect(Collectors.toList()));
     }
 
-    public CompletableFuture<Void> rebalanceZone(@Nullable String zoneTenantId, String zoneName) {
+  /**
+   * Rebalances a zone by re-assigning bound monitors from envoys that are over-assigned
+   * to the least loaded envoys.
+   * @param zoneTenantId for private zones, the tenant owning the zone, or null for public zones
+   * @param zoneName the zone name
+   * @return the number of bound monitors that were re-assigned
+   */
+    public CompletableFuture<Integer> rebalanceZone(@Nullable String zoneTenantId, String zoneName) {
         final ResolvedZone zone = resolveZone(zoneTenantId, zoneName);
 
         return zoneStorage.getZoneBindingCounts(zone)
-            .thenAccept(bindingCounts ->
+            .thenApply(bindingCounts ->
                 rebalanceWithZoneBindingCounts(zone, bindingCounts)
             );
     }
 
-    private void rebalanceWithZoneBindingCounts(ResolvedZone zone,
+    private int rebalanceWithZoneBindingCounts(ResolvedZone zone,
                                                 Map<EnvoyResourcePair, Integer> bindingCounts) {
         if (bindingCounts.size() <= 1) {
             // nothing to rebalance if only one or none envoys in zone
-            return;
+            return 0;
         }
 
         log.debug("Rebalancing zone={} given bindingCounts={}", zone, bindingCounts);
@@ -1058,6 +1065,8 @@ public class MonitorManagement {
         sendMonitorBoundEvents(overassignedEnvoyIds);
         // ...and then this will "re-assign" the bound monitors and send out new bound events
         handleNewEnvoyInZone(zone.getTenantId(), zone.getName());
+
+        return overAssigned.size();
     }
 
     private List<BoundMonitor> findBoundMonitorsWithEnvoy(ResolvedZone zone, String envoyId,
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -384,22 +384,8 @@ public class MonitorManagement {
 
         final ResolvedZone resolvedZone = resolveZone(tenantId, zoneName);
 
-        final List<BoundMonitor> boundToPrev;
-        if ( resolvedZone.isPublicZone()) {
-            boundToPrev = boundMonitorRepository.findWithEnvoyInPublicZone(
-                zoneName,
-                fromEnvoyId,
-                null
-            );
-        }
-        else {
-            boundToPrev = boundMonitorRepository.findWithEnvoyInPrivateZone(
-                tenantId,
-                zoneName,
-                fromEnvoyId,
-                null
-            );
-        }
+        final List<BoundMonitor> boundToPrev = findBoundMonitorsWithEnvoy(
+            resolvedZone, fromEnvoyId, null);
 
         if (!boundToPrev.isEmpty()) {
             log.debug("Re-assigning bound monitors={} to envoy={}", boundToPrev, toEnvoyId);
@@ -1048,17 +1034,8 @@ public class MonitorManagement {
                 final int amountToUnassign = entry.getValue() - avgInt;
                 final PageRequest limit = PageRequest.of(0, amountToUnassign);
 
-                final List<BoundMonitor> boundMonitors;
-                if (zone.isPublicZone()) {
-                    boundMonitors = boundMonitorRepository.findWithEnvoyInPublicZone(
-                        zone.getName(), entry.getKey().getEnvoyId(), limit
-                    );
-                }
-                else {
-                    boundMonitors = boundMonitorRepository.findWithEnvoyInPrivateZone(
-                        zone.getTenantId(), zone.getName(), entry.getKey().getEnvoyId(), limit
-                    );
-                }
+                final List<BoundMonitor> boundMonitors = findBoundMonitorsWithEnvoy(
+                    zone, entry.getKey().getEnvoyId(), limit);
 
                 overAssigned.addAll(boundMonitors);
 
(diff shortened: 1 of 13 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -967,4 +990,95 @@ public void handleExpiredEnvoy(String zoneTenantId, String zoneName, String envo
...
+
+                final List<BoundMonitor> boundMonitors;
+                if (zone.isPublicZone()) {
+                    boundMonitors = boundMonitorRepository.findWithEnvoyInPublicZone(
+                        zone.getName(), entry.getKey().getEnvoyId(), limit
+                    );
+                }
+                else {
+                    boundMonitors = boundMonitorRepository.findWithEnvoyInPrivateZone(
+                        zone.getTenantId(), zone.getName(), entry.getKey().getEnvoyId(), limit
```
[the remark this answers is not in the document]
itzg: Unfortunately each one of them has been a little different, but let me see what I can do.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -967,4 +990,95 @@ public void handleExpiredEnvoy(String zoneTenantId, String zoneName, String envo
...
+                zoneStorage.changeBoundCount(
+                    zone, entry.getKey().getResourceId(), -amountToUnassign
+                );
+            }
+        }
+
+        log.debug("Rebalancing boundMonitors={} in zone={}", overAssigned, zone);
+
+        // tell previous envoys to stop unassigned monitors
+        sendMonitorBoundEvents(extractEnvoyIds(overAssigned));
```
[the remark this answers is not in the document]
itzg: I actually had it that way at first, but the envoyId's are all null at that point. I'll extract the IDs here and then move the send events below the saveAll.

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
```diff
@@ -87,6 +93,22 @@ public ZoneDTO getPublicZone(@PathVariable String name) {
             .toDTO();
     }
 
+    @GetMapping("/tenant/{tenantId}/zoneAssignmentCounts/{name}")
```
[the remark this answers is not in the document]
itzg: Obviously I'm never quite sure the proper convention. Glancing at Github's REST API, hyphenated or "kebab case" seems to be the convention. I'll change it here, but yeah I'm quite sure there's other to fix up.

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
```diff
@@ -87,6 +93,22 @@ public ZoneDTO getPublicZone(@PathVariable String name) {
...
+    @ApiOperation(value = "Gets assignment counts of monitors to poller-envoys in the private zone")
+    public CompletableFuture<List<ZoneAssignmentCount>> getPrivateZoneAssignmentCounts(
```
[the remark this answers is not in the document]
itzg: @jjbuchan It quietly succeeds since it's the range query in etcd that drives the remaining computation. Should I make `ZoneManagement.exists` public and use that in each of these new API controller methods to first check the zone exists?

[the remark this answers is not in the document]
jjbuchan: Yeah, that might be a good idea.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -967,4 +990,95 @@ public void handleExpiredEnvoy(String zoneTenantId, String zoneName, String envo
...
+        @SuppressWarnings("UnstableApiUsage")
+        final Stats stats = Stats.of(values);
+
+        final double stddev = stats.populationStandardDeviation();
+        final double avg = stats.mean();
+        final double threshold = avg + stddev * zonesProperties.getRebalanceStandardDeviations();
+        // round up to be lean towards slightly less reassignments
+        final int avgInt = (int) Math.ceil(avg);
+
+        final List<BoundMonitor> overAssigned = new ArrayList<>();
```
GeorgeJahad: should 'overAssigned' be reset each time through the loop?

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
```diff
@@ -118,6 +140,29 @@ public ZoneDTO update(@PathVariable String name, @Valid @RequestBody ZoneUpdate
         return zoneManagement.updatePublicZone(name, zone).toDTO();
     }
 
+    @PostMapping("/tenant/{tenantId}/rebalance-zone/{name}")
+    @ApiOperation(value = "Rebalances a private zone")
+    @ResponseStatus(HttpStatus.NO_CONTENT)
+    public void rebalancePrivateZone(@PathVariable String tenantId,
+                                     @PathVariable String name) {
+        if (name.startsWith(ResolvedZone.PUBLIC_PREFIX)) {
```
itzg: @jjbuchan rather than inline these checks, I implemented new constraint annotations 

https://github.com/racker/salus-telemetry-etcd-adapter/pull/38/commits/4a9b5eaa65883075171acca2793691868a6bcbed

and am going to use them on the zone name parameters

jjbuchan: Good idea

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
```diff
@@ -142,26 +190,54 @@ public ZoneDTO update(@PathVariable String tenantId, @PathVariable String name,
...
   @ApiOperation(value = "Updates a specific public zone")
   @JsonView(View.Admin.class)
-  public ZoneDTO update(@PathVariable String name, @Valid @RequestBody ZoneUpdate zone) {
+  public ZoneDTO update(HttpServletRequest request, @Valid @RequestBody ZoneUpdate zone) {
+    String name = extractPublicZoneNameFromUri(request);
     return zoneManagement.updatePublicZone(name, zone).toDTO();
   }
 
+  @PostMapping("/tenant/{tenantId}/rebalance-zone/{name}")
+  @ApiOperation(value = "Rebalances a private zone")
```
jjbuchan: I was going to suggest you add a `JsonView` to this endpoint.  Even though the result doesn't need to be filtered now I like having that stuff defined earlier rather than later.

I wonder if we should instead separate the admin and public api endpoints into different files.  That way we could probably set `/api/tenant` in the request mapping for one and `/api/admin` for the other.  That way we might be able to specify `JsonViews` at the top level instead of on the individual endpoints.

This comment is more for something we could do later... not in this PR.

src/main/java/com/rackspace/salus/monitor_management/web/model/ZoneCreatePrivate.java
```diff
@@ -16,19 +16,21 @@
...
+import lombok.Data;
 import org.hibernate.validator.constraints.NotBlank;
-import java.io.Serializable;
 
 @Data
 public class ZoneCreatePrivate implements Serializable {
 
     @NotBlank
     @Pattern(regexp = "^[A-Za-z0-9_]+$", message = "Only alphanumeric and underscore characters can be used")
+    @PrivateZoneName
```
jjbuchan: Maybe we should add the `@Pattern` validation into this one too?

src/main/java/com/rackspace/salus/monitor_management/web/model/ZoneCreatePublic.java
```diff
@@ -16,20 +16,22 @@
...
-import javax.validation.constraints.Min;
 import org.hibernate.validator.constraints.NotBlank;
-import java.io.Serializable;
 
 @Data
 public class ZoneCreatePublic implements Serializable {
 
   @NotBlank
   @Pattern(regexp = "^[A-Za-z0-9_/]+$", message = "Only alphanumeric, underscores, and slashes can be used")
+  @PublicZoneName
```
jjbuchan: Ditto.  Add Pattern into PublicZoneName?

src/test/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiControllerTest.java
```diff
@@ -396,6 +410,222 @@ public void testDeletePrivateZoneWithMonitors() throws Exception {
...
+        );
+
+        when(monitorManagement.getZoneAssignmentCounts(any(), any()))
+            .thenReturn(CompletableFuture.completedFuture(expected));
+
+        final MvcResult result = mvc.perform(
+            get("/api/tenant/{tenantId}/zone-assignment-counts/{name}",
+                "t-1", "z-1"))
+            // CompletableFuture return value, so the request is asynchronous
+            .andExpect(request().asyncStarted())
```
jjbuchan: Forgot about this.

src/main/java/com/rackspace/salus/monitor_management/web/model/ZoneCreatePrivate.java
itzg: Oh, do you mean rather than call out the separate `@Pattern` on L32 compose it onto the `@PrivateZoneName` leaving only that one needed here? I might leave that for a separate PR since I'm not totally sure you can compose constraint annotations.

src/test/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiControllerTest.java
itzg: and I didn't know it existed until these async request handlers were failing on the content type :)

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
itzg: Will plan on doing this in follow up sub-task https://jira.rax.io/browse/SALUS-381

itzg: I started working on splitting the admin and public endpoints into two controller classes, but quickly found that it got tricky with some of the supporting methods like `extractZoneNameFromUri` and `getByZoneName`. I could probably moved shared methods into a common `@Component`, called something like "ZoneControllerServices"...but then I realized it's actually kind of nice to see the pairings of admin/tenant APIs next to each other which are generally public/private zone variants.

I'll add just the `JsonView` for now to get consistency across the APIs in that controller.

PR #43, opened 2019-06-03
Add telegraf net_response and net plugins

itzg:
# Resolves

- https://jira.rax.io/browse/SALUS-373
- https://jira.rax.io/browse/SALUS-372

# What

Adds support for the telegraf plugins
- https://github.com/influxdata/telegraf/tree/master/plugins/inputs/net_response
- https://github.com/influxdata/telegraf/blob/master/plugins/inputs/net/NET_README.md

# How

While I was at it I added a `@ValidGoDuration` and retrofitted the http and x509 plugin schemas with that.

I also added `@ValidHostAndPort` which uses a bit of guava and a bit of Apache Commons validator library. Initially I was going to add these new validators into the "model" repo; however, that would have forced these new dependencies on practically everything. 

If this feels like too much validator bloat in monitor-management, then we can always create a repo for our validators specifically.

## How to test

Added new unit tests.

Speaking of which, rather than keep adding to `MonitorConversionServiceTest` I realized that it'll be better to create a new unit test class for each of these plugin "schemas". If this approach is agreeable then I can split out the existing unit tests into the same structure.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/LocalPlugin.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/LocalPlugin.java
@@ -20,7 +20,12 @@ import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonSubTypes.Type;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeInfo.Id;
-import com.rackspace.salus.monitor_management.web.model.telegraf.*;
+import com.rackspace.salus.monitor_management.web.model.telegraf.Cpu;
+import com.rackspace.salus.monitor_management.web.model.telegraf.Disk;
+import com.rackspace.salus.monitor_management.web.model.telegraf.DiskIo;
+import com.rackspace.salus.monitor_management.web.model.telegraf.Mem;
+import com.rackspace.salus.monitor_management.web.model.telegraf.Net;
+import com.rackspace.salus.monitor_management.web.model.telegraf.Procstat;
 
 @JsonTypeInfo(use = Id.NAME, property = "type")
 @JsonSubTypes({
@@ -28,6 +33,7 @@ import com.rackspace.salus.monitor_management.web.model.telegraf.*;
     @Type(name = "disk", value = Disk.class),
     @Type(name = "diskio", value = DiskIo.class),
     @Type(name = "mem", value = Mem.class),
+    @Type(name = "net", value = Net.class),
     @Type(name = "procstat", value = Procstat.class)
 })
 public abstract class LocalPlugin {
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/NetResponse.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/NetResponse.java
@@ -21,22 +21,35 @@ import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.rackspace.salus.monitor_management.web.model.ApplicableAgentType;
 import com.rackspace.salus.monitor_management.web.model.RemotePlugin;
 import com.rackspace.salus.monitor_management.web.model.validator.ValidGoDuration;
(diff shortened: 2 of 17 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/model/LocalPlugin.java
```diff
@@ -20,14 +20,20 @@
 import com.fasterxml.jackson.annotation.JsonSubTypes.Type;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeInfo.Id;
-import com.rackspace.salus.monitor_management.web.model.telegraf.*;
+import com.rackspace.salus.monitor_management.web.model.telegraf.Cpu;
```
GeorgeJahad: Are "*"'s on imports bad?

itzg: I hadn't heard this explanation before, but also makes sense

http://checkstyle.sourceforge.net/config_imports.html#AvoidStarImport

src/main/java/com/rackspace/salus/monitor_management/web/model/validator/ValidGoDuration.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Indicates that the string is a simplified Go duration formed as an integer value followed by a
+ * time unit suffix of "ms", "s", "m", or "h".
+ */
+@Pattern(regexp = "([0-9]+)(h|m|s|ms)", message = "invalid duration")
+@Target(ElementType.FIELD)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Constraint(validatedBy = {}/*composite*/)
```
GeorgeJahad: interesting, so because you have the Pattern validator added here in the constraint definition, you don't need to use a validator class.  nice!

itzg: Yep. Had to dig in the Hibernate docs to learn the `{}` trick

https://docs.jboss.org/hibernate/stable/validator/reference/en-US/html_single/#section-constraint-composition

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/NetResponse.java
```diff
@@ -0,0 +1,55 @@
...
+
+  @NotNull
+  Protocol protocol;
+
+  @NotNull
+  @ValidHostAndPort
+  String address;
+
+  @ValidGoDuration
+  String timeout;
```
jjbuchan: Looks like this is an initial connect timeout vs. an overall timeout (including read time)?  Can we rename this to something more descriptive?

I feel like this could end up like the agent plugin timeouts where we have two of them and always forget why.

jjbuchan: or does it have to match the same as the plugin field?

itzg: We're stuck with the telegraf naming of the field

https://github.com/influxdata/telegraf/tree/master/plugins/inputs/net_response

jjbuchan: When I was making that comment I was thinking we could use one of the `@Json...` annotations to hide this from the customer, but then I realized us changing it may add a different kind of confusion for support when debugging / comparing to telegraf.

itzg: Yeah, I was pondering some kind of post-processing, but yeah the consistency in field names between "ours" and telegraf feels better than the lack of self-descriptive-ness.

harageth: Some java doc around it might help alleviate some of the issues james was bringing up

itzg: I pondered that too, but essentially we're copying and pasting the telegraf docs at that point.

PR #46, opened 2019-06-07
Use extra attributes in REST error responses and use salus-test

itzg:
# Resolves

Part of
- https://jira.rax.io/browse/SALUS-413
- https://jira.rax.io/browse/SALUS-383

# What

The initial goal was to get "app" and "host" to be included in the REST error JSON responses.

...however, to do that required leaning more on the Spring Boot's `BasicErrorController` and its use of `ErrorAttributes` to build the REST response content. 

That in turn required our `@ControllerAdvice` for exception handling to not override the handling of `MethodArgumentNotValidException`, which is thrown when validation fails.

...however, the standard failures like that are [known to not integrate totally as you'd expected with MockMvc](https://github.com/spring-projects/spring-boot/issues/7321), so I had to update some of the unit tests that assert the validation failures to do those assertions more "MockMvc friendly".

So, that all became a natural reason to integrate with the new salus-test module (https://github.com/racker/salus-test/pull/1) and use the extended error attributes (https://github.com/racker/salus-common/pull/19)

## How to test

Unit tests updated


src/main/java/com/rackspace/salus/monitor_management/web/error/RestExceptionHandler.java
```diff
@@ -18,60 +18,55 @@
...
+  @ExceptionHandler({AlreadyExistsException.class})
+  public ResponseEntity<?> handleAlreadyExists(
+      HttpServletRequest request) {
+    return respondWith(request, HttpStatus.UNPROCESSABLE_ENTITY);
+  }
+
+  private ResponseEntity<?> respondWith(HttpServletRequest request,
+                                        HttpStatus status) {
+    Map<String, Object> body = getErrorAttributes(request);
+    body.put("status", status.value());
```
jjbuchan: How come we need to pull the status from there but not `getReasonPhrase()`?

itzg: Good question. I could include the `getReasonPhrase` also, but "status" itself is normally the numerical value of the HTTP status. I think Spring Boot usually puts the textual version of the status in "message", so I'll check that.

itzg: ...and this mirrors exactly what Spring Boot is doing for "status":

https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/error/BasicErrorController.java#L89

itzg: Good catch. As is, the "error" of "None" is not useful:

```json
{
  "timestamp": "2019-06-07T17:25:23.300+0000",
  "status": 422,
  "error": "None",
  "message": "Zone already exists with name dev on tenant aaaaaa",
  "host": "MS90HCG8WL",
  "reason": "Unprocessable Entity"
}
```

So, I'm going to put the reason phrase in there.

itzg: ...with pushed fix, it is now:

```json
{
  "timestamp": "2019-06-07T17:26:59.248+0000",
  "status": 422,
  "error": "Unprocessable Entity",
  "message": "Zone already exists with name dev on tenant aaaaaa",
  "host": "MS90HCG8WL"
}
```

BUT, I'm going to create an abstract base class situated in salus-common for us to share this logic across modules.

src/main/java/com/rackspace/salus/monitor_management/web/error/RestExceptionHandler.java
```diff
@@ -18,60 +18,41 @@
...
-import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
+import org.springframework.web.bind.annotation.ResponseBody;
 
-@ControllerAdvice
-public class RestExceptionHandler
-        extends ResponseEntityExceptionHandler {
+@ControllerAdvice(basePackages = "com.rackspace.salus.monitor_management.web")
+@ResponseBody
+public class RestExceptionHandler extends
+    com.rackspace.salus.common.web.AbstractRestExceptionHandler {
```
itzg: Uses https://github.com/racker/salus-common/pull/19/files#diff-001d96d78e4a25c047bf15f06e7806c2

itzg:
@jjbuchan , can I get a re-review of the `GenericError` removal?

PR #48, opened 2019-06-11
Fix response parsing in api client

jjbuchan:
# What

Updates the client to handle PagedContent response instead of List.

# How

To avoid getting multiple pages back we now make the request by providing the maximum size possible.

## How to test

Run tests.
Try to connect and envoy to the ambassador.  It now works.

I added a test to verify that there isnt a 1000 object limit in responses, but also tested this manually in case the test case wasn't hitting the right layers.
```
  "number": 0,
  "totalPages": 1,
  "totalElements": 1351,
  "last": true,
  "first": true
}
```
# Why

To fix inter-service communication.

# TODO

Need to double check resource api.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
@@ -17,10 +17,14 @@
 package com.rackspace.salus.monitor_management.web.client;
 
 import com.rackspace.salus.monitor_management.web.model.BoundMonitorDTO;
+import com.rackspace.salus.telemetry.model.PagedContent;
+import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import org.springframework.core.ParameterizedTypeReference;
 import org.springframework.http.HttpMethod;
 import org.springframework.web.client.RestTemplate;
+import org.springframework.web.util.UriComponentsBuilder;
 
 /**
  * This client component provides a small subset of Monitor Management REST operations that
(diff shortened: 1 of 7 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
```diff
@@ -60,12 +64,17 @@ public MonitorApiClient(RestTemplate restTemplate) {
...
   public List<BoundMonitorDTO> getBoundMonitors(String envoyId) {
-    return restTemplate.exchange(
-        "/api/boundMonitors/{envoyId}",
+    final String uri = UriComponentsBuilder
+        .fromPath("/api/admin/bound-monitors/{envoyId}")
+        .queryParam("size", Integer.MAX_VALUE)
+        .build(envoyId)
+        .toString();
+
+    return Objects.requireNonNull(restTemplate.exchange(
```
itzg: That's good to know about `requireNonNull`

jjbuchan: I was debating whether we should catch the exception here as well but I don't think we should.  I think if null is returned it means we're doing something wrong... unless a network issue occurs?

Either way, I think the exception handling would be in the service calling the api vs. in this quiet.

itzg: Yeah, I think letting the NPE bubble up is the best strategy. It'll result in a 500 status code, which is appropriate. Technically in this spot, the `.getContent()` access would have thrown the NPE anyway...however, I like the self-documented aspect of the `requireNonNull` here.

PR #54, opened 2019-06-27
Provide nicer defaults for the disk and cpu plugins

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-464

# What

The existing defaults (false for everything) resulted in a telegraf cpu plugin config that gathered nothing, which is not a nice end user experience...and confused me during ele node testing.

The disk plugin defaults were decent, but excluding the various temp filesystem types is probably common enough that we should provide that as the default. I copied the filesystem types from the [example the plugin provides](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/disk#configuration).

## How to test

Existing unit tests were expanded for cpu and disk to include a "defaults" verification. I also continued in the pattern of the new plugin conversion test organization and migrated those two to a unit test class each.


src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Cpu.java
```diff
@@ -26,7 +27,8 @@
 @ApplicableAgentType(AgentType.TELEGRAF)
 public class Cpu extends LocalPlugin {
   boolean percpu;
-  boolean totalcpu;
+  @JsonProperty(defaultValue = "true")
+  boolean totalcpu = true;
   boolean collectCpuTime;
   boolean reportActive;
```
jjbuchan: When looking at this plugin earlier, I noted down these
```
# Can probably disable this?
percpu = false
# Probably want this?
totalcpu = true
# Shouldn't need this
collect_cpu_time = false
```

So looks like we're on the same page.

PR #64, opened 2019-07-23
WIP/ introducing chained transactations, kafka and jpa

GeorgeJahad:
# Resolves

https://jira.rax.io/browse/SALUS-535

# What

Introducing chained transactions/ kafka and jpa using the same Transactional annotation.  The goal is to maintain the integrity of our system state across all db tables, and kafka.

This code works in my hand tests.  When I force an exception at the end of the methods so marked, the data is not committed to either jpa or kafka.

As you can see, from the code below, it is relatively straightforward, although as always with Spring, it took a long time to figure out how to do it the easy way.

This is a good intro to kafka transactions if you want more info:
https://www.confluent.io/blog/transactions-apache-kafka/

## How to test

I've confirmed that all the old tests still work, but they don't really excercise the chainedTransaction.  I'm writing new tests for that but they aren't working yet which is why this is a draft PR.

# Why

It just seemed like a much more appropriate use of off the shelf tech, rather than trying to implement retries.

# TODO

Still need to finish the new tests.

Also, a couple of things that took me a while to figure out:
1. The brokers need additional config, like so, in the infrastructure docker compose yml:
KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR: 1
KAFKA_TRANSACTION_STATE_LOG_MIN_ISR: 1

2. The consumer has to be inited like so:
spring.kafka.consumer.properties.isolation.level=read_committed

I'll create separate bundle and ambassador pr's for those as well.


GeorgeJahad:
closing for now until we see if this is something worth doing

PR #65, opened 2019-07-23
Avoid binding monitors when the rendered content is not valid

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-495

# What

Previously, a "relaxed" setting of jmustache was being used for rendering monitor content where it would replace unresolvable template variables with an empty string. That leads to a misleading result where bound monitors would still get created, but with incorrect/invalid agent configuration blobs. It is better to avoid creating those bound monitors in the first place and reply upon a resource or monitor change to re-evaluate the monitor content.

# How

Remove the relaxed jmustache setting and handle the newly thrown `InvalidTemplateException`. In most cases it is as simple as not creating the bound monitoring and logging. The hardest case was when a resource change event comes across existing bindings that would become invalid due to a change in resource labels, for example.

## How to test

Existing unit tests and new unit tests were added.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -25,6 +25,7 @@ import com.rackspace.salus.monitor_management.config.ZonesProperties;
 import com.rackspace.salus.monitor_management.entities.BoundMonitor;
 import com.rackspace.salus.monitor_management.entities.Monitor;
 import com.rackspace.salus.monitor_management.entities.Zone;
+import com.rackspace.salus.monitor_management.errors.InvalidTemplateException;
 import com.rackspace.salus.monitor_management.repositories.BoundMonitorRepository;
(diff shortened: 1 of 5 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -177,6 +178,11 @@ ZonesProperties getZonesProperties() {
    * @return The newly created monitor.
    */
   public Monitor createMonitor(String tenantId, @Valid MonitorCU newMonitor) throws IllegalArgumentException, AlreadyExistsException {
+    if (newMonitor.getSelectorScope() == ConfigSelectorScope.LOCAL &&
```
itzg: This change was not related to the initial goal, but fixing a missing validation I notice while unit testing invalid templates.

itzg:
@jjbuchan yeah and also makes me think it would be nice to have an API to query a "what if": given label selectors, what resources would be matched.

PR #67, opened 2019-07-30
Add endpoint to get policy monitors by id

jjbuchan:
# What

Adds an api endpoint to retrieve a monitor under the `_POLICY_` tenant by its id.

# How

Use the existing `getById` method but hardcode the special tenant id.
Add all endpoints and tests.

## How to test

Run tests!

# Why

Policy Management will call this whenever a new monitor policy is created to verify that the monitor actually exists before the policy can be created.

# TODO

More changes will come for this service.  This is just the part that policy mgmt needs to interact with directly (and is used in tests).

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -96,6 +96,8 @@ public class MonitorManagement {
 
   private final EnvoyResourceManagement envoyResourceManagement;
 
+  public static final String POLICY_TENANT = "_POLICY_";
+
   JdbcTemplate jdbcTemplate;
 
   @Autowired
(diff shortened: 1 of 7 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -96,6 +96,8 @@
 
   private final EnvoyResourceManagement envoyResourceManagement;
 
+  public static final String POLICY_TENANT = "_POLICY_";
```
jjbuchan: This might not be the best place for this, but I didn't know where would be better.

itzg: This seems fine. The only other place I can think is in the `Monitor` entity class.

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
```diff
@@ -106,6 +106,13 @@ public DetailedMonitorOutput getById(@PathVariable String tenantId,
                         uuid, tenantId)));
         return monitorConversionService.convertToOutput(monitor);
     }
+    @GetMapping("/admin/policy-monitors/{uuid}")
```
itzg: Regarding this comment

https://github.com/racker/salus-policy-management/pull/1#discussion_r309273832

you could have this be a `@RequestMapping(method=HEAD)` to allow for simply retrieving existence of the policy. Then is would be 200=exists, 404=doesn't based on the HEAD response.

jjbuchan: I considered adding an `exists` endpoint, but figured there will eventually be cases where we do want the full thing, so left it as is.  I think I'd prefer adding that vs. changing this to a HEAD, since having that restriction in the client seems too much, but do agree it would be a good option if we were just hitting the API directly.

I'll add the `exists in a new PR if that's what we decide we want to use.  I'd likely leave this one in here too though.

PR #68, opened 2019-08-07
Use shared entities

jjbuchan:
Up to you whether you review now or wait until the salus-model PR is up as well.

# What

Mostly minimal changes needed after moving entities/repositories to a shared repo.

# How

1. `@EntityScan` and `@EnableJpaRepositories` are needed since those classes are in a different package now
1. Added a new `DatabaseConfig` to hold the jpa annotations.
    * EnableJpaRepositories affects tests if it is on the main App class
1 . Uses entities from salus-model instead of storing them in here
1. Added a constructor to DTOs since I have removed the toDTO methods from the entities.
    * This means the DTOs can stay in each individual repo as they should not need to be shared
1. Removed direct repository usage from controllers to simplify tests

## How to test

Run tests. They all pass.

# Why

Rather than having our services perform api calls to communicate with eachother they will just read from the shared database. This lowers the risk of hitting circular dependency problems and should reduce latency.

# TODO

More PRs are coming for the other repos and for salus-model.

Then I'll loop back around and start removing api calls and replace them with repository lookups.


src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
```diff
@@ -67,15 +65,12 @@
 public class MonitorApiController {
 
     private MonitorManagement monitorManagement;
-    private final BoundMonitorRepository boundMonitorRepository;
     private MonitorConversionService monitorConversionService;
 
     @Autowired
     public MonitorApiController(MonitorManagement monitorManagement,
-                                BoundMonitorRepository boundMonitorRepository,
```
itzg: Thank you for fixing this. I was being lazy at the time by injecting the repository into the controller.

PR #69, opened 2019-08-09
Handle policy management events

jjbuchan:
# What

Adds functionality to handle new monitor policies being created, existing ones being deleted, or an existing policy monitor being updated.

Also handles changes in Tenant Metadata, such as an account type being set leading to the relevant policies changing.

# How

* Adds PolicyMgmt API dependency
* Add producer to send `PolicyMonitorUpdateEvent`s (to be consumed by PolicyMgmt before coming back to MonitorMgmt.
* Adds CRUD methods for policy monitors
* Adds API methods for those CRUD operations
* Updates binding/unbinding methods to require a tenant id, since the monitor tenant could not be `_POLICY_`
* Bound Monitors will use the resource tenant for their tenantId.
* When policy monitors are updated the new monitor will be saved, an event will be sent to policy mgmt which will figure out which tenants are relevant, then send an event per tenant for monitor mgmt to consume and update all the relevant bound monitors.

## How to test

Run tests.

# Why

When we create policies we need them to take effect.

# TODO

More tests (in this PR)
Handle policy metadata within all monitors (in another PR)
Update design doc to fully explain what is going on.


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -587,13 +578,106 @@ else if (monitor.getZones() != null){
...
+    final Set<String> affectedEnvoys = new HashSet<>();
+    Optional<Monitor> monitor = monitorRepository.findById(monitorId);
+
+    if (monitor.isEmpty()) {
+      // This should never happen.
+      log.error("Failed to bind policy to tenant={} due to non-existent monitor={}", tenantId, monitorId);
+      return;
+    }
+    try {
+      validateMonitoringZones(tenantId, monitor.get(),
```
itzg: I'm curious how could the existing monitor end up with invalid zones?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1053,7 +1189,7 @@ private void decrementBoundCounts(List<BoundMonitor> needToDelete) {
     for (BoundMonitor boundMonitor : needToDelete) {
       if (boundMonitor.getEnvoyId() != null &&
           boundMonitor.getMonitor().getSelectorScope() == ConfigSelectorScope.REMOTE) {
-        ResolvedZone zone = getResolvedZoneOfBoundMonitor(boundMonitor);
+        ResolvedZone zone = resolveZone(boundMonitor.getMonitor().getTenantId(), boundMonitor.getZoneName());
```
itzg: Shouldn't this get the tenant ID of the bound monitor rather than the monitor's tenant ID?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1286,7 +1422,36 @@ private int rebalanceWithZoneBindingCounts(ResolvedZone zone,
...
+   * @param resource The resource to get the policy monitors for.
+   * @return A list of monitors.
+   */
+  private List<Monitor> getPolicyMonitorsForResource(Resource resource) {
+    String tenantId = resource.getTenantId();
+    List<UUID> policyMonitorIds = policyApi.getEffectivePolicyMonitorIdsForTenant(tenantId);
+
+    List<Monitor> resourcePolicies = policyMonitorIds
+        .stream()
+        .map(monitorRepository::findById)
```
itzg: I'm not sure if the optimization is worth it, but perhaps could add a `findByIdIn` where you can pass the list `policyMonitorIds`.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1286,7 +1422,36 @@ private int rebalanceWithZoneBindingCounts(ResolvedZone zone,
...
+        .filter(Optional::isPresent).map(Optional::get)
+        .filter(m -> resource.getLabels().entrySet().containsAll(m.getLabelSelector().entrySet()))
```
itzg: Don't you want to turn this into a SQL query? :)

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
```diff
@@ -109,6 +109,41 @@ public DetailedMonitorOutput getPolicyMonitorById(@PathVariable UUID uuid) throw
         return monitorConversionService.convertToOutput(monitor);
     }
 
+    @PutMapping("/admin/policy-monitors/{uuid}")
+    @ApiOperation(value = "Updates specific Policy Monitor")
+    @JsonView(View.Admin.class)
+    public DetailedMonitorOutput update(@PathVariable UUID uuid,
```
itzg: This method should probably be named `updatePolicyMonitor`

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
jjbuchan: When the monitor is created, the zones are only validated against the POLICY_TENANT.  When it comes to actually binding the monitors it will then validate against the customer tenant.

I think this should be fine.  Since policies would be using public zones the POLICY_TENANT should always have access to it.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -284,19 +286,19 @@ private void validateMonitoringZones(String tenantId, Monitor monitor, MonitorCU
...
    * Performs label selection of the given monitor to locate resources for bindings.
    * For remote monitors, this will only perform binding within the given zones.
    * @return affected envoy IDs
    */
-  Set<String> bindMonitor(Monitor monitor,
+  Set<String> bindMonitor(String tenantId, Monitor monitor,
       List<String> zones) {
     final List<ResourceDTO> resources = resourceApi.getResourcesWithLabels(
-        monitor.getTenantId(), monitor.getLabelSelector());
+        tenantId, monitor.getLabelSelector());
```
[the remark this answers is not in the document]
jjbuchan: correct

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
jjbuchan: Yep, it should.  Looks like I accidentally overwrote the real fix with the temporary fix from https://github.com/racker/salus-telemetry-monitor-management/pull/72

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
jjbuchan: I'm choosing to ignore this comment!

src/main/java/com/rackspace/salus/monitor_management/errors/DeletionNotAllowed.java
```diff
@@ -4,6 +4,6 @@
 import org.springframework.web.bind.annotation.ResponseStatus;
 
 @ResponseStatus(HttpStatus.CONFLICT)
-public class ZoneDeletionNotAllowed extends RuntimeException {
-  public ZoneDeletionNotAllowed(String message) { super(message);}
+public class DeletionNotAllowed extends RuntimeException {
```
harageth: Shouldn't this be DeletionNotAllowedException?

src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApi.java
```diff
@@ -19,6 +19,7 @@
 import com.rackspace.salus.monitor_management.web.model.BoundMonitorDTO;
 import com.rackspace.salus.monitor_management.web.model.DetailedMonitorOutput;
 import java.util.List;
+import java.util.UUID;
```
harageth: Looks like this doesn't need to be used here

src/main/java/com/rackspace/salus/monitor_management/services/MonitorEventProducer.java
```diff
@@ -1,7 +1,12 @@
 package com.rackspace.salus.monitor_management.services;
 
+import static com.rackspace.salus.common.messaging.KafkaMessageKeyBuilder.buildMessageKey;
+
 import com.rackspace.salus.common.messaging.KafkaTopicProperties;
 import com.rackspace.salus.telemetry.messaging.MonitorBoundEvent;
+import com.rackspace.salus.telemetry.messaging.PolicyMonitorUpdateEvent;
+import java.util.List;
+import java.util.UUID;
```
harageth: It looks like some of these imports can be removed

jjbuchan: done

src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApi.java
jjbuchan: done

src/main/java/com/rackspace/salus/monitor_management/errors/DeletionNotAllowed.java
jjbuchan: works for me.  changed.

PR #71, opened 2019-08-16
Add support for local test-monitor given a specific resource ID

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-564

# What

Implements the test-monitor feature for the specific case when
- the user provides a resource ID to be evaluated
- monitor to be tested is a "local" type (remote monitors will be supported in follow-up story)

Design proposal is currently here
https://github.com/racker/salus-docs/blob/master/proposals/test-monitor.md

# How

- Adds a new REST endpoint for test-monitor operations
- Adds a new Spring service to process the requests, track the pending requests, and "route" the results back via the pending `CompletableFuture`
- Adds companion Spring services to interface with Kafka

## How to test

Unit tests have been added for the new service beans


src/main/java/com/rackspace/salus/monitor_management/web/controller/RestExceptionHandler.java
```diff
@@ -49,7 +50,7 @@ public RestExceptionHandler(ErrorAttributes errorAttributes) {
     return respondWith(request, HttpStatus.NOT_FOUND);
   }
 
-  @ExceptionHandler({AlreadyExistsException.class})
+  @ExceptionHandler({AlreadyExistsException.class, MissingRequirementException.class})
   public ResponseEntity<?> handleAlreadyExists(
```
GeorgeJahad: should this now be called handleUnprocessable() ?

src/main/java/com/rackspace/salus/monitor_management/web/controller/TestMonitorApiController.java
```diff
@@ -0,0 +1,79 @@
...
+      } else {
+        return HttpStatus.UNPROCESSABLE_ENTITY;
+      }
+    } else {
+      return HttpStatus.OK;
+    }
+  }
+
+  private boolean hasErrors(TestMonitorOutput testMonitorOutput) {
+    return testMonitorOutput.getErrors() != null && !testMonitorOutput.getErrors().isEmpty();
```
GeorgeJahad: CollectionUtils.isNotEmpty()?

src/main/java/com/rackspace/salus/monitor_management/services/TestMonitorService.java
```diff
@@ -0,0 +1,204 @@
...
+    if (prev == null) {
+      log.warn(
+          "Test-monitor with correlationId={} was unexpected absent during removal", correlationId);
+    }
+  }
+
+  private TestMonitorOutput buildTimedOutResult() {
+    return new TestMonitorOutput()
+        .setErrors(
+            singletonList(String.format(
```
GeorgeJahad: out of curiosity, when do you use singletonList(), vs List.of()?

GeorgeJahad:
thanks for the latest commit.  i was just about to ask about the compiler errors.

src/main/java/com/rackspace/salus/monitor_management/services/TestMonitorService.java
itzg: I use `List.of` when I happen to remember we're using Java 11 now :)

src/main/java/com/rackspace/salus/monitor_management/web/controller/TestMonitorApiController.java
itzg: Thanks, that'll be much cleaner.

src/main/java/com/rackspace/salus/monitor_management/web/controller/RestExceptionHandler.java
itzg: Good point. Will do.

src/test/java/com/rackspace/salus/monitor_management/services/TestMonitorServiceTest.java
```diff
@@ -0,0 +1,258 @@
...
+        .setMetrics(expectedMetrics);
+    testMonitorService.handleTestMonitorResultsEvent(resultsEvent);
+
+    assertThat(future.isDone()).isTrue();
+    final TestMonitorOutput output = future.get();
+    assertThat(output).isNotNull();
+    assertThat(output.getErrors()).containsExactly("error-1");
+    assertThat(output.getMetrics()).isEqualTo(expectedMetrics);
+
+    assertThat(testMonitorService.containsCorrelationId(correlationId)).isFalse();
```
GeorgeJahad: is it worth confirming that it was once true?

itzg: Technically the successful `handleTestMonitorResultsEvent` indirectly verifies this; however, I like that idea of explicitly verifying to balance the test code.

itzg:
You can ignore that final re-review request @GeorgeJahad

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -776,19 +776,21 @@ private void handleReattachedEnvoy(String tenantId, String resourceId, String en
             resourceId
         );
 
-    final Set<String> previousEnvoyIds = extractEnvoyIds(bound);
+    if (!bound.isEmpty()) {
```
itzg: This was an unrelated issue that I noticed during end to end testing. I didn't have any monitors created, yet a bound monitor change event was going out for my reattached envoy.

src/main/java/com/rackspace/salus/monitor_management/services/TestMonitorEventProducer.java
```diff
@@ -0,0 +1,48 @@
...
+  public TestMonitorEventProducer(KafkaTemplate kafkaTemplate,
+                                  KafkaTopicProperties kafkaTopicProperties) {
+    this.kafkaTemplate = kafkaTemplate;
+    this.kafkaTopicProperties = kafkaTopicProperties;
+  }
+
+  public void send(TestMonitorRequestEvent event) {
+    final String topic = kafkaTopicProperties.getTestMonitorRequests();
+
+    log.debug("Sending test-monitor request event={} on topic={}", event, topic);
```
jjbuchan: I'm starting to go in the opposite direction from you now.  I wonder if this should be an info?  It's something that shouldn't happen too frequently to just be noise and i already don't think we have enough infos.

src/main/java/com/rackspace/salus/monitor_management/services/TestMonitorService.java
```diff
@@ -0,0 +1,202 @@
...
+
+    final String correlationId = UUID.randomUUID().toString();
+
+    final TestMonitorRequestEvent event = new TestMonitorRequestEvent()
+        .setCorrelationId(correlationId)
+        .setAgentType(monitorCU.getAgentType())
+        .setTenantId(tenantId)
+        .setResourceId(resourceId);
+
+    final ResourceDTO resource = resourceApi.getByResourceId(tenantId, resourceId);
```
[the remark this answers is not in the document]
itzg: Ah yes, thanks for the reminder. I mean to circle around to this especially as you said, it should be a fairly easy replacement.

src/main/resources/application.yml
```diff
@@ -1,5 +1,7 @@
 salus:
   environment: local
+  test-monitor:
+    results-timeout: 60s
```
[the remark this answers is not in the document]
itzg: I definitely agree with you...just thinking I/we can save this for the very next jira issue where remote test-monitors are addressed. With that use case the timeout makes a much bigger difference.

src/main/java/com/rackspace/salus/monitor_management/services/TestMonitorEventProducer.java
itzg: You know I was pondering the same thing after suggesting the info->debug change : we have ended up with very few info logs in our code.

I like your logic that for low cardinality operations an info log is appropriate.

src/main/resources/application.yml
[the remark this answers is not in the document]
jjbuchan: im good with that

src/test/java/com/rackspace/salus/monitor_management/services/TestMonitorServiceTest.java
```diff
@@ -192,11 +245,13 @@ public void testPerformTestMonitorOnResource_timeout()
...
 
-    ResourceDTO resource = new ResourceDTO()
+    Resource resource = new Resource()
         .setResourceId("r-1")
-        .setLabels(Map.of("key-1", "value-1"));
-    when(resourceApi.getByResourceId(any(), any()))
-        .thenReturn(resource);
+        .setLabels(Map.of("key-1", "value-1"))
+        .setCreatedTimestamp(Instant.EPOCH)
+        .setUpdatedTimestamp(Instant.EPOCH);
```
jjbuchan: Ooh, that's a better way than what I've been doing.

itzg: I was curious to see what auto-complete provided and was quite pleased.

itzg:
Re-review needed due to merge from master

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1047,19 +1050,21 @@ private void handleReattachedEnvoy(String tenantId, String resourceId, String en
...
+      final Set<String> previousEnvoyIds = extractEnvoyIds(bound);
 
-    bound.forEach(boundMonitor ->
-        boundMonitor.setEnvoyId(envoyId)
-    );
+      bound.forEach(boundMonitor ->
+          boundMonitor.setEnvoyId(envoyId)
```
harageth: Doesn't Java complain about setting values in one of these forEach loops?

itzg: No but I'm trying to think of what scenario you might be thinking.

itzg: ...btw this was existing code, I just wrapped it up in a condition to see if any bindings even need processing.

PR #76, opened 2019-08-22
Add label selector method & some small changes

jjbuchan:
## You'll want to enable the setting to ignore whitespace changes when reviewing!

# What

* Adds ability to set labelSelectorMethod.
* Adds basic handling of labelSelectorMethod in `getPolicyMonitorsForResource`
* Updates the models used by the api
* Prevents Policy Monitors from being created with a `resourceId`
* Adds tests (that assume the resource API has already been updated)
* Fixes indentation in test file

## How to test

Run tests.
Since not all services are updated yet, this option does not work in reality, but only in tests when we can mock out some of the other requests.

# Why

This adds more flexibility to the Monitor configurations and allows for scenarios such as

```
monitorType: SSH
labelSelector: [{"os":"linux"}, {"agent_discovered_os": "linux"}]
labelSelectorMethod: OR
```

With this an SSH monitor will be applied to resources whether there is a user (or automation) specified label on the resource, or if there is an OS label discovered by the envoy that matches.

# TODO

Adam will be working on these.

* Update resource management and its api to understand the label selector method and act differently whether it is AND or OR.
* Update monitor management label selector query to understand the label selector method.

Other thought:
* We should probably get rid of MonitorDTO and/or rename the detailed output to MonitorDTO

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -32,15 +32,13 @@ import com.rackspace.salus.telemetry.entities.Zone;
 import com.rackspace.salus.monitor_management.errors.InvalidTemplateException;
 import com.rackspace.salus.telemetry.messaging.PolicyMonitorUpdateEvent;
 import com.rackspace.salus.telemetry.messaging.TenantPolicyChangeEvent;
+import com.rackspace.salus.telemetry.model.LabelSelectorMethod;
 import com.rackspace.salus.telemetry.repositories.BoundMonitorRepository;
 import com.rackspace.salus.telemetry.repositories.MonitorRepository;
 import com.rackspace.salus.monitor_management.web.model.MonitorCU;
 import com.rackspace.salus.monitor_management.web.model.ZoneAssignmentCount;
 import com.rackspace.salus.resource_management.web.client.ResourceApi;
 import com.rackspace.salus.resource_management.web.model.ResourceDTO;
-import com.rackspace.salus.telemetry.entities.BoundMonitor;
-import com.rackspace.salus.telemetry.entities.Monitor;
-import com.rackspace.salus.telemetry.entities.Zone;
 import com.rackspace.salus.telemetry.errors.AlreadyExistsException;
 import com.rackspace.salus.telemetry.etcd.services.EnvoyResourceManagement;
 import com.rackspace.salus.telemetry.etcd.services.ZoneStorage;
@@ -230,6 +228,7 @@ public class MonitorManagement {
         .setTenantId(tenantId)
         .setMonitorName(newMonitor.getMonitorName())
         .setLabelSelector(newMonitor.getLabelSelector())
+        .setLabelSelectorMethod(newMonitor.getLabelSelectorMethod())
         .setResourceId(newMonitor.getResourceId())
         .setContent(newMonitor.getContent())
         .setAgentType(newMonitor.getAgentType())
@@ -253,10 +252,16 @@ public class MonitorManagement {
 
     validateMonitoringZones(POLICY_TENANT, null, newMonitor);
 
+    if (!StringUtils.isBlank(newMonitor.getResourceId())) {
+      throw new IllegalArgumentException(
+          "Policy Monitors must use label selectors and not a resourceId");
+    }
+
     Monitor monitor = new Monitor()
         .setTenantId(POLICY_TENANT)
         .setMonitorName(newMonitor.getMonitorName())
         .setLabelSelector(newMonitor.getLabelSelector())
+        .setLabelSelectorMethod(newMonitor.getLabelSelectorMethod())
         .setContent(newMonitor.getContent())
         .setAgentType(newMonitor.getAgentType())
         .setSelectorScope(newMonitor.getSelectorScope())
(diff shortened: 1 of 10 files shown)


itzg:
> We should probably get rid of MonitorDTO and/or rename the detailed output to MonitorDTO

I'm good with renaming of detailed output to become `MonitorDTO`. Some of the current namings were left overs of the GraphQL days.

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorDTO.java
```diff
@@ -35,6 +36,8 @@
 
   Map<String,String> labelSelector;
 
+  LabelSelectorMethod labelSelectorMethod;
```
itzg: Hmm, `MonitorDTO` only seems to be returned by `ZoneApiController#getMonitorsForZone` and it seems like that should actually return what is currently `DetailedMonitorOutput` since it's a public API. So it definitely looks like `MonitorDTO` can be removed and its name re-purposed for `DetailedMonitorOutput`

src/main/java/com/rackspace/salus/monitor_management/web/model/DetailedMonitorInput.java
```diff
@@ -38,6 +39,8 @@
   @ValidLabelKeys
   Map<String,String> labelSelector;
 
+  LabelSelectorMethod labelSelectorMethod;
```
itzg: I think it would be good to default the field to `AND` and also annotate with `@JsonDefault("AND")`. Also adding a `@NotNull` would alleviate any service code from needing to check for null and "default" to AND.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -124,3160 +125,3282 @@
...
+  @Test
+  public void testUpdateExistingMonitor_labelSelectorMethodChanged() {
+    reset(envoyResourceManagement, resourceApi);
+
+    // The labels to set on the monitor
+    Map<String, String> labels = Map.of("os", "linux", "env", "dev");
+
+    final ResourceDTO windowsResource = new ResourceDTO()
+        .setLabels(Map.of("os", "windows", "env", "dev"))
+        .setResourceId("r-1")
```
itzg: nit: if you make this resource id `"r-windows"` it'll be easier to follow the test validation logic further down.

harageth:
Holy large diff batman

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -253,6 +252,11 @@ public Monitor createPolicyMonitor(@Valid MonitorCU newMonitor) {
 
     validateMonitoringZones(POLICY_TENANT, null, newMonitor);
 
+    if (!StringUtils.isBlank(newMonitor.getResourceId())) {
```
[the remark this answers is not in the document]
jjbuchan: Yeah, that crossed my mind but it felt more trouble than it was worth.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -561,16 +569,35 @@ public Monitor updateMonitor(String tenantId, UUID id, @Valid MonitorCU updatedV
...
+      }
+
       affectedEnvoys.addAll(
-          processMonitorLabelSelectorModified(tenantId, monitor, updatedValues.getLabelSelector())
+          processMonitorLabelSelectorModified(tenantId, monitor, labels)
       );
 
-      monitor.setLabelSelector(updatedValues.getLabelSelector());
+      // Finally, update the monitors labels
+      if (labelSelectorChanged) {
```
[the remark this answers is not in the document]
jjbuchan: Good point.  Will change this.  I had also wondered whether we could remove this down here entirely and set it on the monitor before passing through to `processMonitorLabelSelectorModified`, but figured I'd dig into that afterwards

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1549,13 +1594,25 @@ public Monitor getPolicyMonitorForTenant(String tenantId, UUID monitorId) {
...
         .stream()
-        .filter(m -> resource.getLabels().entrySet().containsAll(m.getLabelSelector().entrySet()))
+        .filter(m -> {
+          if (m.getLabelSelector().isEmpty()) {
+            // If no labels are set the monitor applies to all resources
+            return true;
+          }
+          else if (m.getLabelSelectorMethod().equals(LabelSelectorMethod.OR)) {
+            return m.getLabelSelector().entrySet().stream().anyMatch(labels -> {
+              return resource.getLabels().entrySet().contains(labels);
```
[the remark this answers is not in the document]
jjbuchan: If I used that I'd have to do the same for the key.  I'm not sure if I see the optimization.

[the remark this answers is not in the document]
itzg: Yeah, the reduced readability and ripple effect definitely makes it not worth it. The possible optimization (maybe) is avoiding the "creation" of the `entrySet()` just to check if a specific label is included in the map.

src/main/java/com/rackspace/salus/monitor_management/web/model/DetailedMonitorInput.java
jjbuchan: I started to do that but that adds problems for updates.  If this is not provided during an update we do not want to default to AND, we want to know to leave the value as is.  If this was only used for Creates I'd have set a default.

src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorDTO.java
jjbuchan: Yep, I noticed the same.  I'll do some PRs for that.

jjbuchan:
@itzg I pushed one small commit.  If you think I'm misunderstanding your other comments we can chat.

src/main/java/com/rackspace/salus/monitor_management/web/model/DetailedMonitorInput.java
jjbuchan: I'll add that I did set the default on the Monitor over here - https://github.com/racker/salus-telemetry-model/pull/68/files#diff-a4fa15450600ec828ee6939067478edfR75

PR #80, opened 2019-08-28
 Store index of policy metadata used in monitor's contents 

jjbuchan:
# What

Whenever a monitor is created or updated we will store any of the template variables within its content that corresponds to policy metadata.

# How

A regex is used to look for anything that fits the format `${rackspace.metadata.*}` and we will retrieve the value in place of the `*`.

## How to test

Runt tests.
Try manually.

# Why

Whenever policy metadata (the default configuration settings for monitors/tasks) is changed we will be able to easily look up any monitor that is using that field and therefore affected by the change.

# TODO / Incoming PRs

* Add the handling of metadata change events that will utilize this lookup.
   * Update `getRenderedContent` to retrieve the relevant policy metadata fields for the tenant.
* Policy Management PR to introduce the concept of policy metadata and the sending of events that will be consumed by monitor management.
* model changes to allow for this

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/config/MonitorContentProperties.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/config/MonitorContentProperties.java
@@ -16,6 +16,7 @@
 
 package com.rackspace.salus.monitor_management.config;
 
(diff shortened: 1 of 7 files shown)


jjbuchan:
This is all irrelevant now :(

PR #81, opened 2019-08-29
Allowing monitoring interval to be configurable per monitor

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-592

# What

Propagate an optional monitoring interval from the users create and update requests into the stored monitor.

The conversion logic will apply a default per local/remote type and also enforce a configurable minimum bound on the interval.

The monitor update logic needed to be tweaked to ensure that if the user only changed the interval then simply an event went out to the affected envoys. The existing bound monitors don't need to be touched since the interval is not actually stored within those entities.

## How to test

Unit tests update

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/config/MonitorConversionProperties.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/config/MonitorConversionProperties.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2019 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.monitor_management.config;
+
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.convert.DurationUnit;
+
+@ConfigurationProperties("salus.monitor-conversion")
+@Data
+public class MonitorConversionProperties {
+
+  /**
+   * Monitors are not allowed to be created/updated with an interval less than this value.
+   */
+  @DurationUnit(ChronoUnit.SECONDS)
+  Duration minimumAllowedInterval = Duration.ofSeconds(30);
+
+  /**
+   * This is the default value used if a create or update API call provides a local monitor
+   * without <code>interval</code> set.
+   */
+  @DurationUnit(ChronoUnit.SECONDS)
+  Duration defaultLocalInterval = Duration.ofSeconds(60);
+
+  /**
+   * This is the default value used if a create or update API call provides a remote monitor
+   * without <code>interval</code> set.
+   */
+  @DurationUnit(ChronoUnit.SECONDS)
+  Duration defaultRemoteInterval = Duration.ofSeconds(60);
+}
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
@@ -18,7 +18,7 @@ package com.rackspace.salus.monitor_management.services;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.rackspace.salus.telemetry.entities.Monitor;
+import com.rackspace.salus.monitor_management.config.MonitorConversionProperties;
 import com.rackspace.salus.monitor_management.web.model.ApplicableAgentType;
 import com.rackspace.salus.monitor_management.web.model.DetailedMonitorInput;
 import com.rackspace.salus.monitor_management.web.model.DetailedMonitorOutput;
@@ -28,11 +28,14 @@ import com.rackspace.salus.monitor_management.web.model.MonitorCU;
 import com.rackspace.salus.monitor_management.web.model.MonitorDetails;
 import com.rackspace.salus.monitor_management.web.model.RemoteMonitorDetails;
 import com.rackspace.salus.monitor_management.web.model.RemotePlugin;
+import com.rackspace.salus.telemetry.entities.Monitor;
 import com.rackspace.salus.telemetry.model.ConfigSelectorScope;
 import java.io.IOException;
+import java.time.Duration;
 import java.time.format.DateTimeFormatter;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.stereotype.Service;
 
 /**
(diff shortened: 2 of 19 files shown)


src/main/java/com/rackspace/salus/monitor_management/config/MonitorConversionProperties.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.convert.DurationUnit;
+
+@ConfigurationProperties("salus.monitor-conversion")
+@Data
+public class MonitorConversionProperties {
```
harageth: Is this the right name for this? Or was this named such because of the MonitorConversion Epic coming up?

itzg: It's used by the `MonitorConversionService` which has been there a while

https://github.com/racker/salus-telemetry-monitor-management/blob/master/src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java

Maybe the upcoming epic should be called the "even more monitor conversion" :)

harageth: Approved! (And I wont be here for the next couple days... So I am super tempted to rename it)

src/main/java/com/rackspace/salus/monitor_management/config/MonitorConversionProperties.java
```diff
@@ -0,0 +1,48 @@
...
+   */
+  @DurationUnit(ChronoUnit.SECONDS)
+  Duration minimumAllowedInterval = Duration.ofSeconds(30);
+
+  /**
+   * This is the default value used if a create or update API call provides a local monitor
+   * without <code>interval</code> set.
+   */
+  @DurationUnit(ChronoUnit.SECONDS)
+  Duration defaultLocalInterval = Duration.ofSeconds(60);
```
itzg: This and `defaultRemoteInterval` are just a placeholder approach until the policy management changes are in place.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
```diff
@@ -28,25 +28,31 @@
...
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.stereotype.Service;
 
 /**
  * This service enables the public API to interface with the backend REST services for monitor management.
  */
 @Service
 @Slf4j
+@EnableConfigurationProperties(MonitorConversionProperties.class)
```
jjbuchan: Why don't we set `MonitorConversionProperties` as a `@Component`?

I read some docs on this but don't quite understand the use-case.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
```diff
@@ -126,9 +137,26 @@ public MonitorCU convertFromInput(DetailedMonitorInput input) {
...
+          monitor.getSelectorScope() == ConfigSelectorScope.LOCAL ?
+              properties.getDefaultLocalInterval() :
+              properties.getDefaultRemoteInterval()
+      );
+    }
+
     return monitor;
   }
 
+  private void validateInterval(Duration interval) {
```
jjbuchan: Seeing this makes me wonder if `validateMonitoringZones` should be over here too.

src/main/java/com/rackspace/salus/monitor_management/web/model/BoundMonitorDTO.java
```diff
@@ -54,6 +56,7 @@ public BoundMonitorDTO(BoundMonitor boundMonitor) {
     this.tenantId = boundMonitor.getTenantId();
     this.zoneName = boundMonitor.getZoneName();
     this.resourceId = boundMonitor.getResourceId();
+    this.interval = boundMonitor.getMonitor().getInterval();
```
jjbuchan: Aha... above, I was thinking the bound monitor should have the interval within the contents, but this makes sense now.

Also might change my thought in your docs draft pr for monitor conversion.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorConversionServiceTest.java
```diff
@@ -559,4 +572,73 @@ public void testConvertFrom_LabelSelectorMethod() {
...
+
+  @Test
+  public void testConvertTo_Interval() {
+    final UUID monitorId = UUID.randomUUID();
+
+    Monitor monitor = new Monitor()
+        .setLabelSelectorMethod(LabelSelectorMethod.AND)
+        .setId(monitorId)
+        .setInterval(Duration.ofSeconds(60))
+        .setCreatedTimestamp(DEFAULT_TIMESTAMP)
```
jjbuchan: I was starting to change things to `Instant.EPOCH` as I saw them.

A nit would be for you to change it here, but I don't mind if it gets done later.

src/test/resources/BoundMonitorDTOJsonTest/testAllPopulated.json
```diff
@@ -7,6 +7,7 @@
   "renderedContent": "{}",
   "envoyId": "e-1",
   "zoneName": "z-1",
+  "interval": "PT30S",
```
jjbuchan: Just noting that I'm OK with this format since Intelligence will be able to read it and convert to something more human friendly.

src/test/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiControllerTest.java
```diff
@@ -621,6 +626,49 @@ public void testUpdateMonitor_BothLabelsAndResourceId() throws Exception {
...
+  public void testCreateMonitor_intervalParsing_invalidWithNumber() throws Exception {
+    final String content = readContent("MonitorApiControllerTest/create_monitor_duration_number.json");
+
+    mockMvc.perform(post("/api/tenant/t-1/monitors")
+        .content(content)
+        .contentType(MediaType.APPLICATION_JSON)
+        .characterEncoding(StandardCharsets.UTF_8.name()))
+        .andExpect(status().isBadRequest())
+        .andExpect(httpMessageNotReadable(
+            "Cannot deserialize value of type `java.time.Duration` from String \"30\""));
```
jjbuchan: Should we catch this and return a better error that references the format that should be input vs. relating to java.Duration?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
itzg: Good question and I probably should have put a comment since not declaring as `@Component` was unusual here. As I was tweaking the unit tests I found it was cleaner and easier to import the service and then let the service's `@EnableConfigurationProperties` drive the inclusion of this properties class.

itzg: FYI looking through Spring Boot's source code, the approach I used here is actually more idiomatic, but I have tended to like making the property classes `@Component`s instead for clarity.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorConversionServiceTest.java
itzg: Yep, I noticed that while reviewing your PR. I'll do that here.

src/test/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiControllerTest.java
itzg: I was about to say that I don't think we can catch this exception since it is early in the MVC request processing, but maybe `RestExceptionHandler` can post-process the message. The exception handlers don't get invoked for these MVC mocked unit tests, hence these weird result matchers are needed.

I still haven't done end to end testing, so I'll try out the exception handler approach.

itzg: I added an exception handler

https://github.com/racker/salus-telemetry-monitor-management/pull/81/files#diff-03f139be4207edb3b9d87c275c560960R50

but the unit test will have to remain using the uglier message check since MockMvc unit tests don't seem to invoke these exception handlers.

PR #82, opened 2019-08-29
migrated the sql label matching query into a sql file to more easily read it

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-480

# What

This is meant to simplify reading and debugging of any of the actual SQL queries we need to write for Resource Management

# How

This migrates the StringBuilder approach into just a static .sql file. Most of the query is static with only a few locations needing parameters being set. The static query does utilize a String.Format tag to allow us to inject the complex label matching portion of the query into the middle of this template.

## How to test

Run the unit tests. We weren't testing the query in the first place so the tests dont need to change. We just need to make sure they still return the same.

# Why

I considered removing all of the string building from java and building the whole thing with a jmustache template. Ultimately this is for readability of the query and ease of debugging. I didn't think that jmustache would help with the readability and it actually would add a bit of weirdness because we still need to make sure we have cleansed our inputs by passing everything in through parameters to jdbc.

### DIFF
--- a/src/test/java/com/rackspace/salus/monitor_management/services/MonitorConversionServiceTest.java
+++ b/src/test/java/com/rackspace/salus/monitor_management/services/MonitorConversionServiceTest.java
@@ -16,11 +16,11 @@
 
 package com.rackspace.salus.monitor_management.services;
 
-import static com.rackspace.salus.test.JsonTestUtils.readContent;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.junit.Assert.assertEquals;
 
+import com.rackspace.salus.common.util.SpringResourceUtils;
 import com.rackspace.salus.telemetry.entities.Monitor;
 import com.rackspace.salus.monitor_management.web.model.DetailedMonitorInput;
 import com.rackspace.salus.monitor_management.web.model.DetailedMonitorOutput;
@@ -80,7 +80,7 @@ public class MonitorConversionServiceTest {
   public void convertToOutput_diskio() throws IOException {
     // NOTE: this unit test is purposely abbreviated compared convertToOutput
 
-    final String content = readContent("/MonitorConversionServiceTest_diskio.json");
+    final String content = SpringResourceUtils.readContent("/MonitorConversionServiceTest_diskio.json");
 
     Monitor monitor = new Monitor()
         .setId(UUID.randomUUID())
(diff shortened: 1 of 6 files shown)


src/test/java/com/rackspace/salus/monitor_management/services/MonitorConversionServiceTest.java
```diff
@@ -16,11 +16,11 @@
 
 package com.rackspace.salus.monitor_management.services;
 
-import static com.rackspace.salus.test.JsonTestUtils.readContent;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.junit.Assert.assertEquals;
 
+import com.rackspace.salus.common.util.SpringResourceUtils;
```
itzg: If you instead do an `import static` of `com.rackspace.salus.common.util.SpringResourceUtils.readContent` then all the line changes below can be avoided. It's one of the benefits of static imports.

GeorgeJahad:
> We weren't testing the query in the first place so the tests dont need to change. 

Are we really not testing the query?  That seems bad.

itzg:
Thanks @GeorgeJahad , I had missed that statement.

These two unit tests aren't exhaustive, but do seem to exercise the query directly:

https://github.com/racker/salus-telemetry-monitor-management/blob/6d8c5a14cf12103675bef08892f76f67a7259043/src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java#L1409-L1452

The "handleResourceEvent" unit tests probably exercise the query indirectly.

I re-ran the GCB job for this branch and the unit tests are passing.

GeorgeJahad:
i notice that the jira ticket says that this is in review, but I don't see the corresponding PR's  agent-catalog management sql changes.  have I missed those?

https://github.com/racker/salus-telemetry-agent-catalog-management/blob/f726f026b99daae4519699bde9b5520be8aacbda/src/main/java/com/rackspace/salus/acm/services/AgentInstallService.java#L176

harageth:
> > We weren't testing the query in the first place so the tests dont need to change.
> 
> Are we really not testing the query? That seems bad.

I suppose I should have stated that we aren't testing the String that is the query. We are testing the results of the query, which imo is the more important part. If we wanted to test the output String that gets fed into JDBC we would need to rejigger functions and that just seems like a bad idea.

PR #83, opened 2019-09-03
Set monitorType on new Monitor entities

jjbuchan:
# What

When a monitor is created/updated, set the `monitorType` field.

# How

Added an annotation to each plugin that maps it to a `MonitorType` enum value, similar to what is done with AgentTypes.

When we see the DetailedMonitorInput, we can retrieve the monitor type from the plugin class and set it on the MonitorCU object, which will then be used to set it on the Monitor entity.

# Why

For Policy Mgmt we will want to query metadata by monitor type. Having this as a field on the monitor makes life a lot easier. Otherwise we'd have to parse the string content field.

# TODO

Approve https://github.com/racker/salus-telemetry-model/pull/73

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
@@ -161,14 +161,14 @@ public class MonitorConversionService {
     }
   }
 
-  private MonitorType populateMonitorType(MonitorCU monitor, Object plugin) {
+  private void populateMonitorType(MonitorCU monitor, Object plugin) {
     final ApplicableMonitorType applicableMonitorType = plugin.getClass()
         .getAnnotation(ApplicableMonitorType.class);
     if (applicableMonitorType == null) {
       log.warn("monitorClass={} is missing ApplicableMonitorType", plugin.getClass());
       throw new IllegalStateException("Missing ApplicableMonitorType");
     }
-    return applicableMonitorType.value();
+    monitor.setMonitorType(applicableMonitorType.value());
   }
 
   private void populateAgentConfigContent(DetailedMonitorInput input, MonitorCU monitor,
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.rackspace.salus.monitor_management.config.MonitorConversionProperties;
 import com.rackspace.salus.monitor_management.web.model.ApplicableAgentType;
+import com.rackspace.salus.monitor_management.web.model.ApplicableMonitorType;
 import com.rackspace.salus.monitor_management.web.model.DetailedMonitorInput;
 import com.rackspace.salus.monitor_management.web.model.DetailedMonitorOutput;
 import com.rackspace.salus.monitor_management.web.model.LocalMonitorDetails;
@@ -115,7 +116,6 @@ public class MonitorConversionService {
 
     final MonitorCU monitor = new MonitorCU()
         .setMonitorName(input.getName())
-        .setMonitorType(input.getDetails().getType())
         .setLabelSelector(input.getLabelSelector())
         .setLabelSelectorMethod(input.getLabelSelectorMethod())
         .setResourceId(input.getResourceId())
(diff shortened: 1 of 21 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/model/LocalMonitorDetails.java
```diff
@@ -16,16 +16,29 @@
...
+import lombok.extern.slf4j.Slf4j;
 
+@Slf4j
 @ApiModel(parent=MonitorDetails.class)
 @Data @EqualsAndHashCode(callSuper = false)
 public class LocalMonitorDetails extends MonitorDetails {
   @NotNull @Valid
   LocalPlugin plugin;
+
+  public MonitorType getType() {
```
itzg: It seems risky for JSON serialization to embed the annotation logic as a getter within the DTO class especially since `type` is a synthetic field declared by this `@JsonTypeInfo`

https://github.com/racker/salus-telemetry-monitor-management/blob/6d8c5a14cf12103675bef08892f76f67a7259043/src/main/java/com/rackspace/salus/monitor_management/web/model/MonitorDetails.java#L27

Also the `type` within `MonitorDetails` is supposed to be `local` or `remote`, so I would expect some serialization unit tests to break.

Was there a reason not to put this annotation logic in `MonitorConversionService`?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -264,6 +266,7 @@ public Monitor createPolicyMonitor(@Valid MonitorCU newMonitor) {
     Monitor monitor = new Monitor()
         .setTenantId(POLICY_TENANT)
         .setMonitorName(newMonitor.getMonitorName())
+        .setMonitorType(MonitorType.cpu)
```
itzg: This one seems to be hardcoded.

src/main/java/com/rackspace/salus/monitor_management/web/model/LocalMonitorDetails.java
itzg: Ah, the unit test only checks for `id` 

https://github.com/racker/salus-telemetry-monitor-management/blob/6d8c5a14cf12103675bef08892f76f67a7259043/src/test/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiControllerTest.java#L127

and not the overall structure of the serialized JSON. The `$.details.type`would be the field of interest to verify.

jjbuchan: I can move it there.  I didn't really consider it at the time.  Since I was retrieving a value from a `MonitorDetails` object it just made sense to me to put it in here.  It is now part of the object available to anything that needs to use it vs. being in a helper class.

I think I do prefer it how I've done it... although, I see the risk with having this named `getType`.  No tests break with this change, but I can look into that more.  If I change it to `getMonitorType` and all output is as expected would that work for you?  Kinda makes sense to me to add `getAgentType` into here too...

jjbuchan: fyi, that tests returns the expected body

```
Body = {"id":"0005f54a-9688-37f5-0005-f54a968a3c2f","name":"K9VNKDgRpy","labelSelector":{"YPrPpf2ppA":"ybN8MG5OAV","kz6ZlmrP0f":"ltP3C0NOgL","rwxfTeZN1S":"6dnYDxQy2M","2sVZR_yHcv":"30tfvT2dBg","cXzRWRFPLT":"2CnJhzatQA"},"labelSelectorMethod":"OR","resourceId":"ZcJTD_nMzJ","interval":"PT465H51M15.580731887S","details":{"type":"local","plugin":{"type":"mem"},"type":"mem"},"createdTimestamp":"2019-09-03T21:01:57.479988Z","updatedTimestamp":"2019-09-03T21:01:57.464584Z"}
```

jjbuchan: eh.. or not.

itzg: Renaming the getter (and annotating with `@JsonIgnore`) would technically solve it. Your logic makes sense, however there's only one call location and the annotation retrieval borders on business logic. DTOs should be free of business logic.

jjbuchan: After refactoring I realized I was already using the `getType` directly on a `MonitorDetails` object in `TestMonitorService.performTestMonitorOnResource`.  But now I can change to using the monitorCU since we get that via
```
final MonitorCU monitorCU = monitorConversionService.convertFromInput(
        new DetailedMonitorInput()
            .setDetails(details)
    );
```

Since that does provide an easy way to get the field when we don't have a `Monitor` I'm good with the change.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
```diff
@@ -126,6 +128,7 @@ public MonitorCU convertFromInput(DetailedMonitorInput input) {
       monitor.setSelectorScope(ConfigSelectorScope.LOCAL);
 
       final LocalPlugin plugin = ((LocalMonitorDetails) details).getPlugin();
+      populateMonitorType(monitor, plugin);
```
itzg: Thanks. I like how this fits right in with existing code structure.

PR #86, opened 2019-09-13
Adam9500/salus 433 hide internal endpoints

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-433

# What

This PR is trying to catch the error when the API nodes are up but a service behind the API node isn't. 

# How

I am catching the particular error that gets thrown when another service isn't running. 

## How to test

Start this service after starting all of the docker containers. 
After this has started shut down mysql.
Send a creation request in.
Bring mysql back up and shut down kafka.
send the same creation request.

# Why

For unknown reasons Spring Boot doesn't seem to appreciate using the configuration options for not including the exception or the stack trace.

Instead I decided to utilize the pattern we are already using for catching exceptions and responding with HTTP status codes to give a friendly message.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorEventProducer.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorEventProducer.java
@@ -45,9 +45,7 @@ public class MonitorEventProducer {
         log.debug("Sending policyMonitorUpdateEvent={} on topic={}", event, topic);
         try {
             kafkaTemplate.send(topic, buildMessageKey(event), event).get();
-        } catch (InterruptedException e) {
-            throw new RuntimeKafkaException(e);
-        } catch (ExecutionException e) {
+        } catch (InterruptedException|ExecutionException e) {
             throw new RuntimeKafkaException(e);
         }
     }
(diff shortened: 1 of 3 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorEventProducer.java
```diff
@@ -28,13 +30,23 @@ public void sendMonitorEvent(MonitorBoundEvent event) {
         final String topic = properties.getMonitors();
 
         log.debug("Sending monitorBoundEvent={} on topic={}", event, topic);
-        kafkaTemplate.send(topic, event.getEnvoyId(), event);
+        try {
+            kafkaTemplate.send(topic, event.getEnvoyId(), event).get();
+        } catch (InterruptedException e) {
+            throw new RuntimeKafkaException(e);
+        } catch (ExecutionException e) {
```
itzg: One more that could be combined.

harageth: Weird it says locally that its been changed and pushed up.

PR #87, opened 2019-09-15
add db plugins

GeorgeJahad:
# Resolves

https://jira.rax.io/browse/SALUS-392
# What

Adding db plugins

## How to test

Run unit tests

# Why

Because we decided to make these plugins both remote and local, there are many duplicate files in this PR; (a separate copy of the schema for both local and remote versions of the plugins, and their corresponding tests).

I tried a couple of things to reduce this duplication but was unhappy with all of them.

First I tried removing the "remotePlugin" and "localPlugin" classes and just using a single "plugin" class for both.  That seemed to work, in so far as I could create either local or remote from the same schema.  Just specifying the "details.type" field in the api request determined which is used.

The problem was I didn't have a good way to force them to be only local or remote. For example, CPU could be made a remote monitor, which I wasn't happy with.  Also some validations only go on the local version of a monitor, and I couldn't easily come up with a way to handle that, without having separate schemas for each.

In any case, here is my experimental branch so you can see how I implemented that, if you are interested:
https://github.com/racker/salus-telemetry-monitor-management/compare/93e06d7e8dbda019e8845c02df8547bfa9e06d44...george/testReorgPlugins

Then, I tried using reflection/generics to generalize the code and tests, but I made even less progress there.  I wasn't able to get it really working and the things I was doing seemed very complicated; even if I had gotten them working, the increase in complexity would have not been worth the corresponding reduction in code duplication.

I'm not that experienced with generics/reflection so maybe there is an easier way to do it.  I'd be interested in hearing if you have any ideas along that lines, or any other suggestions for how we can reduce the duplication.

In the meantime, to ease the review of this PR, I've added some diff commands below that you can use from the osx command line.  They allow you to bring up the duplicate files in the intellij diff mode.  That way, if you prefer, you can just review the local files in github.  The remote files can be reviewed by just diffing them with the intellij diff commands below.  Note that you have to be in the monitor-management root directory,  on branch george/addDbPlugins.

```
idea diff src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Mysql.java src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/MysqlRemote.java

idea diff src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Postgresql.java src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/PostgresqlRemote.java

idea diff src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/SqlServer.java src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/SqlServerRemote.java

idea diff src/main/java/com/rackspace/salus/monitor_management/web/validator/PostgresqlRemoteValidator.java src/main/java/com/rackspace/salus/monitor_management/web/validator/PostgresqlValidator.java

idea diff src/test/java/com/rackspace/salus/monitor_management/web/model/telegraf/MysqlConversionTest.java src/test/java/com/rackspace/salus/monitor_management/web/model/telegraf/MysqlRemoteConversionTest.java

idea diff src/test/java/com/rackspace/salus/monitor_management/web/model/telegraf/PostgresqlConversionTest.java src/test/java/com/rackspace/salus/monitor_management/web/model/telegraf/PostgresqlRemoteConversionTest.java

idea diff src/test/java/com/rackspace/salus/monitor_management/web/model/telegraf/SqlServerConversionTest.java src/test/java/com/rackspace/salus/monitor_management/web/model/telegraf/SqlServerRemoteConversionTest.java

idea diff src/test/java/com/rackspace/salus/monitor_management/web/model/validator/PostgresqlRemoteValidatorTest.java src/test/java/com/rackspace/salus/monitor_management/web/model/validator/PostgresqlValidatorTest.java
```


itzg:
> Because we decided to make these plugins both remote and local, there are many duplicate files in this PR; (a separate copy of the schema for both local and remote versions of the plugins, and their corresponding tests).
> 
> I tried a couple of things to reduce this duplication but was unhappy with all of them.

Before I start reviewing I just wanted to say I appreciated the attempts to consolidate the changes and it sounds like the right choice was made in the end for the right reasons.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Postgresql.java
```diff
@@ -0,0 +1,45 @@
...
+import lombok.EqualsAndHashCode;
+
+@Data @EqualsAndHashCode(callSuper = true)
+@ApplicableAgentType(AgentType.TELEGRAF)
+@ApplicableMonitorType(MonitorType.postgresql)
+@AtMostOneOf
+public class Postgresql extends LocalPlugin {
+  @NotEmpty
+  @ValidLocalHost
+  String address;
```
itzg: Similar my comment on the mysql plugin, a regex validation would be good here to confirm the url structure is valid

```
postgres://[pqgotest[:password]]@localhost[/dbname]?sslmode=[disable|verify-ca|verify-full]
```

or their "connection string" format, which has an odd value for the `host`. Is that an executable path?
```
host=/run/postgresql user=telegraf database=telegraf
```

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Postgresql.java
```diff
@@ -0,0 +1,45 @@
...
+  String outputaddress;
```
itzg: I don't see this field in the docs
https://github.com/influxdata/telegraf/blob/master/plugins/inputs/postgresql/README.md

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Postgresql.java
```diff
@@ -0,0 +1,45 @@
...
+  @ValidGoDuration
+  String maxLifetime;
```
itzg: I don't see this field in the docs
https://github.com/influxdata/telegraf/blob/master/plugins/inputs/postgresql/README.md

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/SqlServer.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.validation.constraints.NotEmpty;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data @EqualsAndHashCode(callSuper = true)
+@ApplicableAgentType(AgentType.TELEGRAF)
+@ApplicableMonitorType(MonitorType.sqlserver)
+public class SqlServer extends LocalPlugin {
+  @NotEmpty
+  List<@ValidLocalHost String> servers;
```
itzg: Similar comment about regex validation on connection string

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/SqlServer.java
```diff
@@ -0,0 +1,47 @@
...
+  @JsonProperty("query_version")
+  @Min(2)
+  @Max(2)
+  Integer queryVersion = 2;
+  boolean azuredb;
+  // Jackson seems to exclude serialiation of fields whose names start with 'exclude' unless:
```
itzg: Wow, that's a bit too "helpful"

src/main/java/com/rackspace/salus/monitor_management/web/validator/PostgresqlRemoteValidator.java
```diff
@@ -0,0 +1,36 @@
...
+import com.rackspace.salus.monitor_management.web.model.telegraf.PostgresqlRemote;
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+import javax.validation.Constraint;
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import javax.validation.Payload;
+
+public class PostgresqlRemoteValidator implements ConstraintValidator <PostgresqlRemoteValidator.AtMostOneOf, PostgresqlRemote> {
```
itzg: I see what you mean how reflection might have avoided some duplication here...but these validators are small and simple enough that I agree not worth the complexity.

src/test/java/com/rackspace/salus/monitor_management/web/model/telegraf/MysqlConversionTest.java
```diff
@@ -0,0 +1,185 @@
...
+
+    Monitor monitor = createMonitor(content, "convertToOutput", AgentType.TELEGRAF,
+        ConfigSelectorScope.LOCAL
+    );
+
+    final DetailedMonitorOutput result = conversionService.convertToOutput(monitor);
+
+    final Mysql mysqlPlugin = assertCommon(result, monitor, Mysql.class, "convertToOutput");
+
+    List<String> l = new ArrayList<>();
```
itzg: FYI, now that we're on Java 11, could use `List.of("1", "2");`

src/test/java/com/rackspace/salus/monitor_management/web/model/validator/PostgresqlRemoteValidatorTest.java
```diff
@@ -0,0 +1,81 @@
...
+    DetailedMonitorInput input = new DetailedMonitorInput()
+        .setDetails(details);
+
+    final Set<ConstraintViolation<DetailedMonitorInput>> errors = validatorFactoryBean.validate(input);
+    assertThat(errors, hasSize(1));
+    assertThat(new ArrayList<>(errors).get(0).getMessage(),
+        containsString(PostgresqlRemoteValidator.AtMostOneOf.DEFAULT_MESSAGE));
+  }
+
+}
```
itzg: How about one more test case that sets one or the other of `databases` or `ignoredDatabases`?

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Mysql.java
```diff
@@ -0,0 +1,64 @@
...
+  boolean gatherEventWaits;
+  boolean gatherTableSchema;
+  boolean gatherFileEventsStats;
+  boolean gatherPerfEventsStatements;
+  @ValidGoDuration
+  String intervalSlow;
+  @Min(2)
+  @Max(2)
+  Integer metricVersion = 2;
+  String tlsCa;
```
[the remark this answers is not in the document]
GeorgeJahad: This is the local monitor schema, so it is fine here I think.  Perhaps we would do that for remote, but even then I hesitate because it would preclude usage for private pollers.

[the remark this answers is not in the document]
itzg: Good point. I forgot the user could stage their own TLS files.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Postgresql.java
GeorgeJahad: https://github.com/influxdata/telegraf/blob/master/plugins/inputs/postgresql/postgresql.go#L42-L47

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/SqlServer.java
GeorgeJahad: the sqlserver plugin supports a fairly complicated variety of connection formats/options: https://github.com/denisenkom/go-mssqldb

do you think we should support all of them?  if not, what subset should we support?

itzg: I was thinking this could just be the minimum regex to sanity check the leading part of the connection string and just `.*` the trailing, query parameter parts. Ensuring the connection string starts with "sqlserver://" could even be sufficient.

GeorgeJahad:
@itzg I think I've addressed all of your comments

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Mysql.java
```diff
@@ -0,0 +1,67 @@
...
+import lombok.EqualsAndHashCode;
+
+@Data @EqualsAndHashCode(callSuper = true)
+@ApplicableAgentType(AgentType.TELEGRAF)
+@ApplicableMonitorType(MonitorType.mysql)
+public class Mysql extends LocalPlugin {
+  public static final String REGEXP = "^((?<user>.+(:.+)?@)?(tcp(?<address>\\(.+\\))))?/(?<dbname>[^?]*)?(?<params>(\\?.+=[^&]+)?(&.+=.+)*)?$";
+  public static final String ERR_MESSAGE = "invalid mysql db connection string";
+  @NotEmpty
+  List<@ValidLocalHost @Pattern(regexp = Mysql.REGEXP, message = Mysql.ERR_MESSAGE) String> servers;
```
jjbuchan: Does it make sense for the local monitor to receive a list?

This might be something that the translation layer handles and is irrelevant for this PR, but I'd expect all that would be needed is a port, and in the end we'd even default that to 3306.

src/main/java/com/rackspace/salus/monitor_management/web/model/validator/ValidLocalHost.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.validation.Constraint;
+import javax.validation.Payload;
+import javax.validation.constraints.Pattern;
+
+@Pattern(regexp = ".*(localhost|127.0.0.1).*", message = "address must point to localhost")
+@Target({ElementType.FIELD, ElementType.TYPE_USE})
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Constraint(validatedBy = {})
+public @interface ValidLocalHost {
```
jjbuchan: This confused me for a while.  I was expecting another file to contain the actual constraint restrictions but realize now they're all handled by the `@Pattern` above.

Based on my earlier comment about whether we even need a list vs. just hardcoding localhost, this annotation may be less relevant but doesn't hurt to keep it for now either way.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Postgresql.java
GeorgeJahad: I'm guessing that is a bug in the docs.  In the go file, it shows this:
https://github.com/influxdata/telegraf/blob/master/plugins/inputs/postgresql/postgresql.go#L26-L30

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Mysql.java
```diff
@@ -0,0 +1,67 @@
...
+  boolean gatherIndexIoWaits;
+  boolean gatherEventWaits;
+  boolean gatherTableSchema;
+  boolean gatherFileEventsStats;
+  boolean gatherPerfEventsStatements;
+  @ValidGoDuration
+  String intervalSlow;
+  @Min(2)
+  @Max(2)
+  Integer metricVersion = 2;
```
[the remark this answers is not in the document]
itzg: I have added this as a specific example in the monitor translation proposal

https://github.com/racker/salus-docs/pull/9/files#diff-ac86d1d06a9365e9415c1d7f817e42adR37

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/MysqlRemote.java
```diff
@@ -0,0 +1,64 @@
...
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.Pattern;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data @EqualsAndHashCode(callSuper = true)
+@ApplicableAgentType(AgentType.TELEGRAF)
+@ApplicableMonitorType(MonitorType.mysql)
+public class MysqlRemote extends RemotePlugin {
```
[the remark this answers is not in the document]
jjbuchan: > purposely naming two classes the same can get awkward

I had the same concern.  Even accidentally including one when you want the other.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Mysql.java
GeorgeJahad: It is defined as a list, so the translation layer would have to change it to be a single string.  also, typically there will be a user name/password, so a default probably won't be that useful.

PR #88, opened 2019-09-18
Fix tests & a couple other things

jjbuchan:
# What

Added a comment to the Metadata handling method.  They are currently in a state that will work for tests but will fail if you use the api to update policy metadata.  I doubt anyone will run into this problem until we're able to find a solution.  I'm going to continue looking into it.

Flipped the condition in `consumeMetadataPolicyUpdateEvents`.  I started writing that logic one way and decided to change it but never flipped the if statement.  This also shows why I was wrong not to add tests for that kafka consumer (I've still left them off here, but will add them in future).

My recent PR led to recent test additions by George to fail due to the method signature being changed.  I've fixed all those tests so the build should now pass.

### DIFF
--- a/src/test/java/com/rackspace/salus/monitor_management/web/model/telegraf/MysqlConversionTest.java
+++ b/src/test/java/com/rackspace/salus/monitor_management/web/model/telegraf/MysqlConversionTest.java
@@ -21,6 +21,8 @@ import static com.rackspace.salus.monitor_management.web.model.telegraf.Conversi
 import static com.rackspace.salus.test.JsonTestUtils.readContent;
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.rackspace.salus.monitor_management.utils.MetadataUtils;
+import com.rackspace.salus.policy.manage.web.client.PolicyApi;
 import com.rackspace.salus.telemetry.entities.Monitor;
 import com.rackspace.salus.monitor_management.services.MonitorConversionService;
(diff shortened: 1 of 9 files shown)


src/test/java/com/rackspace/salus/monitor_management/web/model/telegraf/MysqlConversionTest.java
```diff
@@ -29,31 +31,42 @@
...
 public class MysqlConversionTest {
   @Configuration
   public static class TestConfig { }
 
+  @MockBean
+  PolicyApi policyApi;
+
+  @MockBean
+  MonitorRepository monitorRepository;
+
```
GeorgeJahad: I don't see these being used.  What happens if you leave them out?

jjbuchan: For these tests we don't care that they respond with `null` but we need the bean to exist.

```
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'com.rackspace.salus.monitor_management.services.MonitorConversionService': Unsatisfied dependency expressed through constructor parameter 3; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'com.rackspace.salus.telemetry.repositories.MonitorRepository' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
```

PR #89, opened 2019-09-24
Make tests act the same way as prod

jjbuchan:
# What

Convert the test to a SpringBootTest instead of DataJPATest

# How

Once converting the test definition the "weird" JPA steps were removed from MonitorManagement.  

The change from asserting on monitors to monitor ids in `assertThat(m, isOneOf(monitorsThatUseMetadata.toArray()));` was needed because the interval value is now different in `monitorsThatUseMetadata` compared to `m`.  With `DataJPATest`, the monitors stored in `monitorsThatUseMetadata` would be updated when the `handle...` method was ran.  These are now separated with `SpringBootTest`.

Additionally, the exception we assert on had to be changed since a different layer is now being tested.

Assertions for `getPluginMetadataFields` and `getMonitorMetadataFields` were removed since they would fail due to the lazily loaded exception now.

# Why

This change makes the test act the same way as production, which is much more ideal.  We do not have to do any workarounds to make the test work.

Those line removals had to be done to get things to work in prod, but would previously cause tests to fail.  With the change to `@SpringBootTest` the `LazyInitializationException` would occur when running tests if those lines were not removed.

### DIFF
--- a/src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagement_MetadataPolicyTest.java
+++ b/src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagement_MetadataPolicyTest.java
@@ -32,6 +32,7 @@ import com.rackspace.salus.monitor_management.config.DatabaseConfig;
 import com.rackspace.salus.monitor_management.config.MonitorContentProperties;
 import com.rackspace.salus.monitor_management.config.ServicesProperties;
 import com.rackspace.salus.monitor_management.config.ZonesProperties;
+import com.rackspace.salus.monitor_management.services.MonitorManagement_MetadataPolicyTest.TestConfig;
 import com.rackspace.salus.monitor_management.utils.MetadataUtils;
 import com.rackspace.salus.monitor_management.web.model.MonitorCU;
 import com.rackspace.salus.policy.manage.web.client.PolicyApi;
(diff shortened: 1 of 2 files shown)


itzg:
> Assertions for getPluginMetadataFields and getMonitorMetadataFields were removed since they would fail due to the lazily loaded exception now.

Have you tested yet if the `MonitorApiController` calls encounter lazy loading exceptions when building the DTOs?

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagement_MetadataPolicyTest.java
```diff
@@ -79,24 +80,30 @@
...
+import org.springframework.transaction.TransactionSystemException;
 import uk.co.jemos.podam.api.PodamFactory;
 import uk.co.jemos.podam.api.PodamFactoryImpl;
 
 @RunWith(SpringRunner.class)
-@DataJpaTest(showSql = false)
-@Import({ServicesProperties.class, ObjectMapper.class, MonitorManagement.class,
+@SpringBootTest
+@AutoConfigureDataJpa
+@ContextConfiguration(classes = {
```
itzg: If you want to shave off an annotation, I believe it'll work to move this `classes` into `@SpringBootTtest`

jjbuchan:
While testing manually I do a GET on the list of monitors and don't hit any issues.  Also, all of our other tests still pass.

We also do not want to pass those fields through to the API, so that might be benefiting us.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagement_MetadataPolicyTest.java
jjbuchan: I kinda like this way.  Maybe because it's more generic and can be used across all test types.

itzg: That's logical; however, FYI I have been starting to use the `classes` property of `@SpringBootTest` in new and converted unit tests :)

PR #91, opened 2019-09-26
Support Logical OR processing for monitors

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-606

# What

Support logical OR as well as logical AND processing for matching labels. This label matching is more complex than for resource management because we need to figure out which objects in the database are configured for OR and apply that query to them and which are configured for AND and apply that query to them.

# How

I duplicated the AND query and modified it for handling OR processing. 

I then query the database separately for both and combine the results together for returning.

## How to test

Run the unit tests. The specific functions that test this have the function ending in "UsingOr"

# Why

We considered handling this in a hybrid approach through JPA or Hibernate. Neither of those places seemed to have a good way of querying map values. 

Another hybrid option of doing this through JDBC still required the writing of the OR based query and then handling the rest of the processing server side. Since we already have the AND query it seems more logical to me to just handle this all at query time, since we need both queries.

### DIFF
--- a/src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
+++ b/src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
@@ -9,27 +9,27 @@ AND      monitors.id IN
   FROM   monitor_label_selectors
   WHERE  monitors.id IN
   (
-    SELECT id
-    FROM   monitors
+    SELECT first_monitors.id
+    FROM   monitors AS first_monitors
     WHERE  tenant_id = :tenantId
   )
   AND monitors.id IN
     (
-      SELECT monitor_id
+      SELECT monitor_label_selectors.monitor_id
       FROM monitor_label_selectors
       WHERE
-      monitor_id IN
+      monitor_label_selectors.monitor_id IN
       (
-        SELECT id
-        FROM monitors
+        SELECT inner_monitors.id
+        FROM monitors AS inner_monitors
         WHERE tenant_id = :tenantId
       )
       AND monitors.id IN
       (
-        SELECT monitor_id
+        SELECT monitor_label_selectors.monitor_id
         FROM monitor_label_selectors
         WHERE %s
-        GROUP BY id
+        GROUP BY monitor_label_selectors.monitor_id
         HAVING COUNT(*) >= 1
       )
  ))
--- a/src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
+++ b/src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
@@ -15,10 +15,10 @@ AND      monitors.id IN
   )
   AND monitors.id IN
     (
-      SELECT id
+      SELECT monitor_id
       FROM monitor_label_selectors
       WHERE
-      id IN
+      monitor_id IN
       (
         SELECT id
         FROM monitors
@@ -26,7 +26,7 @@ AND      monitors.id IN
       )
       AND monitors.id IN
       (
-        SELECT id
+        SELECT monitor_id
         FROM monitor_label_selectors
         WHERE %s
         GROUP BY id
--- a/src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
+++ b/src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
@@ -1,0 +1,35 @@
+SELECT   monitors.id
+FROM     monitors
+JOIN     monitor_label_selectors AS ml
+where    monitors.id = ml.monitor_id
+AND      monitors.label_selector_method = 'OR'
+AND      monitors.id IN
+         (
+    SELECT monitor_id
+    FROM   monitor_label_selectors
+    WHERE  monitors.id IN
+       (
+          SELECT id
+          FROM   monitors
+          WHERE  tenant_id = :tenantId)
+    AND    monitors.id IN
+        (
+      SELECT id
+      FROM monitor_label_selectors
+      WHERE
+         id IN
+         (
+            SELECT id
+            FROM monitors
+            WHERE tenant_id = :tenantId
+         )
+         AND monitors.id IN
+         (
+            SELECT id
+            FROM monitor_label_selectors
+            WHERE %s
+            GROUP BY id
+            HAVING COUNT(*) >= 1
+         )
+   ))
+ORDER BY monitors.id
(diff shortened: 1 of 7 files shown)


src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
```diff
@@ -0,0 +1,35 @@
+SELECT   monitors.id
+FROM     monitors
+JOIN     monitor_label_selectors AS ml
+where    monitors.id = ml.monitor_id
+AND      monitors.label_selector_method = 'OR'
+AND      monitors.id IN
```
itzg: I know the previous queries were like this, but I'm now wondering why the tenant ID clause isn't up here at the top level...seems like MySQL can optimize for that better if it's here, such as

```
AND monitors.tenant_id = :tenantId
```

src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
```diff
@@ -0,0 +1,35 @@
...
+         (
+    SELECT monitor_id
+    FROM   monitor_label_selectors
+    WHERE  monitors.id IN
+       (
+          SELECT id
+          FROM   monitors
+          WHERE  tenant_id = :tenantId)
```
itzg: ...my theory/hope with the comment above, is that it might reduce one level of WHERE...IN...SELECT here.

src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
harageth: Thats true. I can run both through the query planner and see what it tells me about them. I would hope that it can handle that through the optimizer, but it is a graph problem that might not get propagated through.

itzg: Looking at this again I see what you mean how it's the inner query that needs the `tenant_id` clause rather than this outer expression, which would be "too late" to be efficient.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -1646,6 +1650,31 @@ public void testMatchMonitorWithSupersetOfLabels() {
...
+    String tenantId = RandomStringUtils.randomAlphanumeric(10);
+    monitorManagement.createMonitor(tenantId, create);
+    entityManager.flush();
+
+    Page<Monitor> monitors = monitorManagement.getMonitorsFromLabels(labels, tenantId, Pageable.unpaged());
+    assertEquals(0L, monitors.getTotalElements()); //make sure we only returned the one value
+  }
+
+  @Test
+  public void testMatchMonitorWithSupersetOfLabelsUsingOr() {
```
itzg: I'm confused about this test name that ends with "UsingOr", yet on L1677 it is

```
    create.setLabelSelectorMethod(LabelSelectorMethod.AND);
```

itzg: I'm not sure how this test differs from `testMisMatchMonitorWithSupersetOfLabels`...but maybe the diff is making that hard to see.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -1693,6 +1747,38 @@ public void testMatchMonitorWithSubsetOfLabels() {
...
+  }
+
+  @Test
+  public void testMatchMonitorWithSubsetOfLabelsUsingOr() {
+    final Map<String, String> monitorLabels = new HashMap<>();
+    monitorLabels.put("os", "DARWIN");
+    monitorLabels.put("env", "test");
+    final Map<String, String> labels = new HashMap<>();
+    labels.put("os", "DARWIN");
+    labels.put("env", "test");
```
itzg: I would like to see one of these two "resource labels" with a value that differs from the monitor label selector to ensure the OR is satisfied by only one of the two matching.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
harageth: Uhhhh oops... I thought I had changed that.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
harageth: Yeah I did that in the Resource Management tests but not sure why I didn't do it here.

GeorgeJahad:
"TODO" still says to add  "or" test cases. is that still true?

harageth:
> "TODO" still says to add "or" test cases. is that still true?

No it isn't. I forgot to alter that on both of these PR's.

GeorgeJahad:
still in the "how to test" section as well

src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
```diff
@@ -0,0 +1,36 @@
...
+  WHERE  monitors.id IN
+  (
+    SELECT id
+    FROM   monitors
+    WHERE  tenant_id = :tenantId
+  )
+  AND monitors.id IN
+    (
+      SELECT id
+      FROM monitor_label_selectors
```
GeorgeJahad: This may be a dumb sql question but in the monitor_label_selectors table, the column name is monitor_id, not id, so why is id being selected in line 18 instead of monitor_id?

harageth: This is a very good question. I dont know how the tests are passing.

GeorgeJahad: yeh I can run this command from the mysql prompt:
```
select monitor_id from monitor_label_selectors;

```
but this returns an error:
```
select id from monitor_label_selectors;
ERROR 1054 (42S22): Unknown column 'id' in 'field list'
```

harageth: Yeah that was about the extent of what I was able to do. Right now I am reduced to running the spring boot application from the command line and my command line java resolves to version 8. So I need to update that to run the application. Either way I am pushing up and update for this now.

GeorgeJahad: before you push the update, i'd like to understand why the tests were passing.  if there is a bug in our tests, i'd like to know sooner, rather than later.

PR #92, opened 2019-10-02
Use logRequestFailure in exception handler and map RestTemplate errors

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-634

# What

Right now we're getting a vague response body like the following when creating a monitor in our dev cluster. The logs don't report any further details about the error, so it's hard to debug what is misconfigured or behaving incorrectly.

```json
{
  "timestamp": "2019-10-02T21:07:34.558+0000",
  "status": 400,
  "error": "Bad Request",
  "message": "URI is not absolute",
  "host": "monitor-management-bfcfd748c-fgw6z"
}
```

# How

Make use of changes in https://github.com/racker/salus-common/pull/40

## How to test

Use existing tests

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
@@ -16,6 +16,8 @@
 
 package com.rackspace.salus.monitor_management.web.client;
 
+import static com.rackspace.salus.common.web.RemoteOperations.mapRestClientExceptions;
+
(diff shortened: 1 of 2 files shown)


jjbuchan:
> All checks have failed

Might need to check that first?

itzg:
@jjbuchan thanks for noticing that. There's something weird with the resource-management build where it is deploying the PR/branch artifacts. I fixed this same kind of thing earlier in the week by rebuilding master...so I'll do that again to fix for now.

itzg:
...ah, the non-deploy config is doing a deploy

https://github.com/racker/salus-telemetry-resource-management/blob/master/cloudbuild.yaml#L32

PR #93, opened 2019-10-07
Add PATCH method to Monitor Management

jjbuchan:
# What

Adds a PATCH method to Monitor Management API to be able to perform individual updates to a Monitor, including resetting a value to `null`.

# How
Followed this as a guide - https://cassiomolin.com/2019/06/10/using-http-patch-in-spring/

Utilizes the JsonMergePatch from the JSON-P API.

1. Receive a `JsonMergePatch` in the body of the request in the controller
    * A custom `AbstractHttpMessageConverter` was introduced to be able to generate this
1. Get the stored Monitor with the id in the request path
1. Convert that Monitor to a `DetailedMonitorInput`
1. Merge the `JsonMergePatch` onto the `DetailedMonitorInput` to replace any values provided
1. Continue in a similar way to a PUT request, except if a `null` is seen use it instead of ignoring it.
    * This is handled by the `boolean patchOperation` flag

## How to test

Run tests.

In `MonitorManagementTest.java` we have a lot of specific tests for updating monitors, such as `testUpdateExistingMonitor_labelsChanged()`.  Should I recreate all of those but for PATCH?  Since intelligence will be using PATCH it seems like I should even with how tedious and repetitive it is?

# Why

We need to be able to reset set values to their default.  This is done by passing a `null` value in the request.  But we also need to distinguish between what was a provided null vs. what is unset.  If we were to map the request immediately to a `DetailedMonitorInput` both the fields purposely set to null and those that were not set at all would come through as null, which is not helpful.

I looked into a couple other solutions.

1. Setting all parameters in the request to `Optional`.  We would then have to evaluate each one to see whether it was `null` or whether it was an optional that contained `null`.
    * This would have required large changes across our code to change how those are treated
1. Add a boolean or bit fields alongside each field in the input and set it to true in the setter for that field.  If the field has a value provided it will get set to true, if it doesn't it will be left as false.
    * This would have been OK for the top level Monitor fields but it gets messy for all the input plugins
    * We would have to analyze every field on the plugin to see if it was set.
    * We'd probably need an individual handler for each plugin to do this
    * It would be a large amount of code changes required

# TODO

* The question in the **how to test** section?
* I'll update Insomnia and play around with it a bit to make sure no other issues show up

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -779,7 +779,13 @@ public class MonitorManagement {
     if (zonesChanged(updatedValues.getZones(), monitor.getZones(), patchOperation)) {
       // See above regarding:
       // JPA's EntityManager is a little strange with re-saving (aka merging) an entity
-      monitor.setZones(new ArrayList<>(updatedValues.getZones()));
+      if (updatedValues.getZones() == null) {
+        // policy monitors cannot use metadata, so this value cannot be null.
+        // ignore the change and keep the original zones.
+        monitor.setZones(new ArrayList<>(monitor.getZones()));
+      } else {
+        monitor.setZones(new ArrayList<>(updatedValues.getZones()));
+      }
     } else if (monitor.getZones() != null) {
       monitor.setZones(new ArrayList<>(monitor.getZones()));
     }
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -606,9 +606,12 @@ public class MonitorManagement {
   /**
    * Update an existing monitor.
    *
-   * @param tenantId      The tenant to update the monitor for.
-   * @param id            The id of the existing monitor.
-   * @param updatedValues The new monitor parameters to store.
+   * @param tenantId       The tenant to update the monitor for.
+   * @param id             The id of the existing monitor.
+   * @param updatedValues  The new monitor parameters to store.
+   * @param patchOperation Whether a patch or update/put is being performed. True for patch.
+   *                       null values are ignored for updates but utilized in patches.
+   *
    * @return The newly updated monitor.
    */
   public Monitor updateMonitor(String tenantId, UUID id, @Valid MonitorCU updatedValues, boolean patchOperation) {
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -63,6 +63,7 @@ import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
 import java.io.IOException;
 import java.io.InvalidClassException;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -269,7 +270,7 @@ public class MonitorManagement {
       monitor.setLabelSelectorMethod(newMonitor.getLabelSelectorMethod());
     }
 
-    metadataUtils.setMetadataFieldsForMonitor(tenantId, monitor);
+    metadataUtils.setMetadataFieldsForMonitor(tenantId, monitor, false);
 
     monitor = monitorRepository.save(monitor);
     final Set<String> affectedEnvoys = bindNewMonitor(tenantId, monitor);
@@ -598,6 +599,10 @@ public class MonitorManagement {
     return zones;
   }
 
+  public Monitor updateMonitor(String tenantId, UUID id, @Valid MonitorCU updatedValues) {
+    return updateMonitor(tenantId, id, updatedValues, false);
+  }
+
   /**
    * Update an existing monitor.
    *
@@ -606,7 +611,7 @@ public class MonitorManagement {
    * @param updatedValues The new monitor parameters to store.
    * @return The newly updated monitor.
    */
-  public Monitor updateMonitor(String tenantId, UUID id, @Valid MonitorCU updatedValues) {
+  public Monitor updateMonitor(String tenantId, UUID id, @Valid MonitorCU updatedValues, boolean patchOperation) {
     Monitor monitor = getMonitor(tenantId, id).orElseThrow(() ->
         new NotFoundException(String.format("No monitor found for %s on tenant %s",
             id, tenantId)));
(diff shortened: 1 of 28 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
```diff
@@ -170,6 +175,25 @@ public DetailedMonitorOutput updatePolicyMonitor(@PathVariable UUID uuid,
         ));
   }
 
+  @PatchMapping("/admin/policy-monitors/{uuid}")
+  @ApiOperation(value = "Patch specific Policy Monitor")
+  @JsonView(View.Admin.class)
+  public DetailedMonitorOutput patchPolicyMonitor(@PathVariable UUID uuid,
```
jjbuchan: While scanning over this myself, I realized we can maybe remove this (and associated tests).  Since policy monitors don't use metadata we may not need to set anything to `null`.

However, before doing so, I do have a point to chat about first around null monitoring zones.

src/test/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiControllerTest.java
```diff
@@ -432,6 +439,56 @@ public void testUpdateNonExistentMonitor() throws Exception {
...
+
+    mockMvc.perform(patch(url)
+        .content(objectMapper.writeValueAsString(update))
+        .contentType(MediaType.valueOf(JSON_MERGE_PATCH_TYPE))
+        .characterEncoding(StandardCharsets.UTF_8.name()))
+        .andDo(print())
+        .andExpect(status().isOk())
+        .andExpect(content()
+            .contentTypeCompatibleWith(MediaType.APPLICATION_JSON));
+  }
```
itzg: Maybe add `verify`s of `monitorManagement`, but probably not too important for this particular test case.

src/test/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiControllerTest.java
```diff
@@ -432,6 +439,56 @@ public void testUpdateNonExistentMonitor() throws Exception {
...
+    String tenantId = monitor.getTenantId();
+    UUID id = monitor.getId();
+    String url = String.format("/api/tenant/%s/monitors/%s", tenantId, id);
+
+    DetailedMonitorInput update = podamFactory.manufacturePojo(DetailedMonitorInput.class);
+    update.setLabelSelector(null);
+    update.setDetails(new LocalMonitorDetails().setPlugin(new Mem()));
+
+    mockMvc.perform(patch(url)
+        .content(objectMapper.writeValueAsString(update))
```
itzg: I'm wondering if this should use json content from a file to be even more precise about what PATCH content is being presented.

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
```diff
@@ -170,6 +175,24 @@ public DetailedMonitorOutput updatePolicyMonitor(@PathVariable UUID uuid,
         ));
   }
 
+  @PatchMapping("/admin/policy-monitors/{uuid}")
```
GeorgeJahad: the regular patch api specifies: "consumes = JSON_MERGE_PATCH_TYPE"

any reason why this doesn't?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -668,38 +676,50 @@ else if (monitor.getLabelSelector() != null) {
       monitor.setContent(updatedValues.getContent());
     }
 
-    if (zonesChanged(updatedValues.getZones(), monitor.getZones())) {
-      // Process potential changes to bound zones
+    PropertyMapper map;
+    if (patchOperation) {
```
GeorgeJahad: All the "if (patchOperation)" tests feel like a need for refactoring behind an interface.

I don't feel strongly enough about it to  request a change here, but if there were a few more of the if statements, I probably would.

GeorgeJahad:
> In MonitorManagementTest.java we have a lot of specific tests for
>   updating monitors, such as
>   testUpdateExistingMonitor_labelsChanged(). Should I recreate all of
>   those but for PATCH?

I'm not sure you have to recreate all of them, but right now I don't see any.  Seems like something more than 0 would be good.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
```diff
@@ -501,25 +511,89 @@ public void testUpdatePolicyMonitor() {
   }
 
   @Test
-  public void testUpdatePolicyMonitor_setResourceId() {
```
GeorgeJahad: testUpdatePolicyMonitor_setResourceId no longer exists.  is that intentional?

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
jjbuchan: It should.  Pushed a fix.

However, I was wondering whether we should allow that type and the general application/json for both of these to make it easier for users?  Otherwise they'll have to ensure they specify "application/merge-patch+json" for these patch requests.

> Resources that use the "application/merge-patch+json" media type are required to conform to the "application/json" media type

That's from the RFC.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
jjbuchan: I'm not sure how that happened.  Don't think I intended that.  Added it back in.

Thanks for catching.

jjbuchan:
> I'm not sure you have to recreate all of them, but right now I don't see any. Seems like something more than 0 would be good.

There's none of those specific tests but there are others that test more general cases.  My thinking was if we feel like any of them should be copied over then all of them probably should.  I almost ended up doing that before pushing the PR but other things came up so left it as is.

I'll bring them in tomorrow or on Monday.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -668,38 +676,50 @@ else if (monitor.getLabelSelector() != null) {
...
     }
 
-    PropertyMapper map = PropertyMapper.get();
-    map.from(updatedValues.getMonitorName())
-        .whenNonNull()
-        .to(monitor::setMonitorName);
-
-    metadataUtils.setMetadataFieldsForMonitor(tenantId, monitor);
-    monitor.setPluginMetadataFields(updatedValues.getPluginMetadataFields());
-
```
GeorgeJahad: I guess I'm not convinced you need to add a patched variant of all the tests,  just a few tests that excercise this area of the code, where it isn't already excercised by the existing tests.

jjbuchan: That section is already hit by https://github.com/racker/salus-telemetry-monitor-management/blob/6bd62aea1243e867e195252a9c5b226f7ed76858/src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagement_MetadataPolicyTest.java#L252

jjbuchan: My original thought with those tests is that they are testing very specific things and making sure only the expected bound monitors occur and no more.

As I look at them more, maybe it's not worth adding much extra beyond the above test for patch... although, since that is an "all null" test, i'll add a partial null and all values set examples too.  Seems like that might be enough.

itzg: That sounds like a great idea.

src/test/java/com/rackspace/salus/monitor_management/utils/MetadataUtilsTest.java
```diff
@@ -240,6 +240,27 @@ public void testSetMetadataFieldsForMonitor() {
     metadataUtils.setMetadataFieldsForMonitor(tenantId, monitor, false);
     assertThat(monitor.getMonitorMetadataFields()).hasSize(0);
 
+    metadataUtils.setMetadataFieldsForMonitor(tenantId, monitor, false);
+    assertThat(monitor.getMonitorMetadataFields()).hasSize(3);
+    assertThat(monitor.getMonitorMetadataFields())
```
GeorgeJahad: sorry, i'm missing something.  what has caused them number of metadata fields to go to 3?

jjbuchan: @GeorgeJahad your comment is on an old commit that I screwed up when merging master.
This is the current version.

https://github.com/racker/salus-telemetry-monitor-management/blob/fb8e3d3cd77bded71ac420fd12017252de294dd5/src/test/java/com/rackspace/salus/monitor_management/utils/MetadataUtilsTest.java#L223-L241

GeorgeJahad: that looks better, thanks!

PR #98, opened 2019-10-18
Convert merge-patch to patch

jjbuchan:
# What

Convert the existing JsonMergePatch method into a JsonPatch.

# How

* Updated `PatchHelper` to include null values when converting to the JsonStructure
   * Uses JacksonAnnotationIntrospector
      * This was the hardest thing to figure out
   * Without this an exception would be thrown when things like `pingInterval` were provided in the payload but were set to null in the existing monitor.
* Added a `convertToInput` method
   * Was originally for other reasons that never ended up being needed, but in the end it saves converting to an Output object first for any patches
* A little cleanup in `MonitorConversionService` for shared methods
* Update tests to use the Patch payload format.

# Why

MergePatch ended up having some issues.  Since it was a merge if you provided `{}` as the payload to something that currently was set to `{"key": "value"}` no change would occur since it was actually merging the keys from the payload into the object... and since there were no keys in there, nothing would be done.

That meant a PUT would be needed if you ever wanted to set something to `{}`, but with PUT you cannot set things to `null` so then the PATCH would be needed.

That behavior would make it difficult for something like Intelligence to develop against... this new PATCH functionality should cover all scenarios simply by using the `replace` operation.

# TODO

I'm pushing this to get eyes on it and because nothing much needs to change, but I want to look into the `JsonException` a little more to see if we can determine why it was triggered.  In some cases we'd probably want to return a 400 but in others a 422.  The 400 is probably more likely so that may end up being the default if I can't find a way to distinguish between them all.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/converter/PatchHelper.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/converter/PatchHelper.java
@@ -16,7 +16,10 @@
 
 package com.rackspace.salus.monitor_management.web.converter;
 
+import static com.rackspace.salus.monitor_management.config.JsonConfig.IGNORE_JSON_INCLUDE_ANNOTATIONS;
+
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jsr353.JSR353Module;
 import java.util.Set;
 import javax.json.JsonMergePatch;
 import javax.json.JsonPatch;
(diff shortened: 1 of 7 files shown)


itzg:
> I want to look into the JsonException a little more

Is that referring to the `pingInterval` scenario you described? Without having started the review yet, I'm fine with a 400 if further narrowing is not easy or possible.

src/main/java/com/rackspace/salus/monitor_management/web/converter/PatchHelper.java
```diff
@@ -31,15 +34,23 @@
...
   @Autowired
   public PatchHelper(ObjectMapper objectMapper, Validator validator) {
     this.objectMapper = objectMapper;
     this.validator = validator;
+    // A custom object mapper is required for patches to ensure fields that are normally
+    // excluded if nullare not excluded here.
+    // These null fields must be available to allow new values to be set.
+    this.objectMapperAllowNulls = new ObjectMapper()
+        .setAnnotationIntrospector(IGNORE_JSON_INCLUDE_ANNOTATIONS)
+        .registerModule(new JSR353Module());
```
itzg: This instantiation got me curious and I found that Spring Boot is declaring a `Jackson2ObjectMapperBuilder` in https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/jackson/JacksonAutoConfiguration.html which you could autowire here to build the custom `ObjectMapper`. I'm hopeful that builder automatically takes into account the module registrations.

PR #99, opened 2019-10-21
Remove @JsonInclude from Monitors

jjbuchan:
Following on from https://github.com/racker/salus-telemetry-monitor-management/pull/98

We realized that we probably want to send any fields that can be input by the user back to them (even if they are null).  This will help make the behavior more consistent/intuitive for intelligence.

The logic from https://github.com/racker/salus-telemetry-monitor-management/pull/98 to ignore fields with these annotations is going to remain in place until we have got further along with the translation layer since I'm unsure if it could be beneficial there at all.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/DiskIo.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/DiskIo.java
@@ -16,8 +16,6 @@
 
 package com.rackspace.salus.monitor_management.web.model.telegraf;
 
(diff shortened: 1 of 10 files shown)


jjbuchan:
This kinda thing was what I was left wondering about.

> populate fixed values, such as `metric_version = 2` of the mysql plugin

The options seemed to be to leave it on the plugin class and exclude it from being relayed to the customer or populate it after the fact.  Since you've stated "populate" I think the need for ignoring the JsonInclude annotations on a patch is gone.

I'll create a PR to remove that logic.  If needed at least it'll be in one place to revert in future if needed

PR #103, opened 2019-10-30
Add System Monitor

jjbuchan:
Adds a System monitor - https://github.com/influxdata/telegraf/tree/master/plugins/inputs/system

This will replace the agent.load_average check in ele.  It also comes with the bonus of providing uptime metrics.


src/test/java/com/rackspace/salus/monitor_management/services/MonitorConversionServiceTest.java
```diff
@@ -237,6 +238,49 @@ public void convertFromInput_mem() throws JSONException, IOException {
     JSONAssert.assertEquals(content, result.getContent(), true);
   }
 
+  @Test
+  public void convertToOutput_system() throws IOException {
```
itzg: It's totally not obvious, especially since we never retrofitted the existing unit tests, but the new approach is to create a unit test class per plugin class, such as

https://github.com/racker/salus-telemetry-monitor-management/blob/master/src/test/java/com/rackspace/salus/monitor_management/web/model/telegraf/CpuConversionTest.java

PR #107, opened 2019-11-25
Added Oracle Agent tests and migrated the helper functions

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-693

# What

Adds the Oracle Agent as something that Monitor Management understands as well as the dataguard, rman, and tablespace checks for that agent. 

# How

It follows our paradigm for adding new monitors to Monitor Management.

## How to test

Unit tests have been added following the Conversion Service format that Geoff has created. This will only test the validity of the Monitors as far as conversions and Monitor Management is concerned.

# Why

It is the approach that we took with all of our other monitors

# TODO
Everything should be done

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/LocalPlugin.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/LocalPlugin.java
@@ -46,9 +46,9 @@ import com.rackspace.salus.monitor_management.web.model.telegraf.System;
     @Type(name = "postgresql", value = Postgresql.class),
     @Type(name = "sqlserver", value = SqlServer.class),
     @Type(name = "system", value = System.class),
-    @Type(name = "rman", value = Rman.class),
-    @Type(name = "tablespace", value = Tablespace.class),
-    @Type(name = "dataguard", value = Dataguard.class)
+    @Type(name = "oracle_rman", value = Rman.class),
+    @Type(name = "oracle_tablespace", value = Tablespace.class),
+    @Type(name = "oracle_dataguard", value = Dataguard.class)
 })
 public abstract class LocalPlugin {
 
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/LocalPlugin.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/LocalPlugin.java
@@ -20,6 +20,9 @@ import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonSubTypes.Type;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeInfo.Id;
+import com.rackspace.salus.monitor_management.web.model.oracle.Dataguard;
+import com.rackspace.salus.monitor_management.web.model.oracle.Rman;
+import com.rackspace.salus.monitor_management.web.model.oracle.Tablespace;
 import com.rackspace.salus.monitor_management.web.model.telegraf.Cpu;
 import com.rackspace.salus.monitor_management.web.model.telegraf.Disk;
 import com.rackspace.salus.monitor_management.web.model.telegraf.DiskIo;
@@ -42,7 +45,10 @@ import com.rackspace.salus.monitor_management.web.model.telegraf.System;
     @Type(name = "mysql", value = Mysql.class),
     @Type(name = "postgresql", value = Postgresql.class),
     @Type(name = "sqlserver", value = SqlServer.class),
-    @Type(name = "system", value = System.class)
+    @Type(name = "system", value = System.class),
+    @Type(name = "rman", value = Rman.class),
+    @Type(name = "tablespace", value = Tablespace.class),
+    @Type(name = "dataguard", value = Dataguard.class)
 })
 public abstract class LocalPlugin {
 
--- a/src/test/java/com/rackspace/salus/monitor_management/web/model/oracle/DataguardConversionTest.java
+++ b/src/test/java/com/rackspace/salus/monitor_management/web/model/oracle/DataguardConversionTest.java
@@ -55,7 +55,7 @@ public class DataguardConversionTest {
 
   @Test
   public void convertToOutput_dataguard() throws IOException {
-    final String content = readContent("/ConversionTests/MonitorConversionServiceTest_dataguard.json");
+    final String content = readContent("/ConversionTests/MonitorConversionServiceTest_oracle_dataguard.json");
 
     Monitor monitor = createMonitor(content, "convertToOutput", AgentType.ORACLE,
         ConfigSelectorScope.LOCAL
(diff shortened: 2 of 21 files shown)


src/test/java/com/rackspace/salus/monitor_management/web/model/oracle/DataguardConversionTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@JsonTest
+@Import({MonitorConversionService.class, MetadataUtils.class})
+public class DataguardConversionTest {
+  @Configuration
+  public static class TestConfig { }
```
GeorgeJahad: i may be revealing my ignorance here, but why is TestConfig empty?

src/test/java/com/rackspace/salus/monitor_management/web/model/oracle/DataguardConversionTest.java
```diff
@@ -0,0 +1,105 @@
...
+  MonitorRepository monitorRepository;
+
+  @Autowired
+  MonitorConversionService conversionService;
+
+  @Autowired
+  MetadataUtils metadataUtils;
+
+  @Test
+  public void convertToOutput_rman() throws IOException {
```
GeorgeJahad: rman?

src/test/java/com/rackspace/salus/monitor_management/web/model/oracle/DataguardConversionTest.java
```diff
@@ -0,0 +1,105 @@
...
+    final String content = readContent("/ConversionTests/MonitorConversionServiceTest_dataguard.json");
+
+    Monitor monitor = createMonitor(content, "convertToOutput", AgentType.ORACLE,
+        ConfigSelectorScope.LOCAL
+    );
+
+    final DetailedMonitorOutput result = conversionService.convertToOutput(monitor);
+
+    final Dataguard dataguardPlugin = assertCommon(result, monitor, Dataguard.class, "convertToOutput");
+    assertThat(dataguardPlugin.getFilePath());
```
GeorgeJahad: Do these asserts need to invoke a method, (like the .isEqualTo() below)?

src/test/java/com/rackspace/salus/monitor_management/web/model/oracle/DataguardConversionTest.java
harageth: It was pulled over on a copy from the dependency of @itzg's tests. I had tried a few different runs removing a few of these that I didn't think we needed but hadn't tried removing this. It works without it, so I removed it.

src/test/java/com/rackspace/salus/monitor_management/web/model/oracle/TablespaceConversionTest.java
```diff
@@ -0,0 +1,102 @@
...
+  MonitorRepository monitorRepository;
+
+  @Autowired
+  MonitorConversionService conversionService;
+
+  @Autowired
+  MetadataUtils metadataUtils;
+
+  @Test
+  public void convertToOutput_rman() throws IOException {
```
GeorgeJahad: rman

src/test/java/com/rackspace/salus/monitor_management/web/model/oracle/DataguardConversionTest.java
harageth: They follow the same convention that Geoff had in his tests. It wouldn't hurt to check it against whats actually in the json test file.

GeorgeJahad: which test?

harageth: Color me mistaken. I thought he was doing it this way in the CpuConversion tests but you are correct he is checking the values against the json values.

src/test/java/com/rackspace/salus/monitor_management/web/model/oracle/DataguardConversionTest.java
itzg: FYI in some `@SpringBootTest` cases I use this technique to force Spring to ignore all of the application components and instead use only the `@Import`ed ones. I have since learned I can achieve the same result with the `classes` attribute of the annotation, such as

https://github.com/racker/salus-telemetry-monitor-management/blob/7aac638487eb4db9dcc9800ac0bf526a9c7e1531/src/test/java/com/rackspace/salus/monitor_management/services/ResourceEventListenerTest.java#L37-L41

src/main/java/com/rackspace/salus/monitor_management/web/model/LocalPlugin.java
```diff
@@ -42,7 +45,10 @@
     @Type(name = "mysql", value = Mysql.class),
     @Type(name = "postgresql", value = Postgresql.class),
     @Type(name = "sqlserver", value = SqlServer.class),
-    @Type(name = "system", value = System.class)
+    @Type(name = "system", value = System.class),
+    @Type(name = "rman", value = Rman.class),
+    @Type(name = "tablespace", value = Tablespace.class),
+    @Type(name = "dataguard", value = Dataguard.class)
```
itzg: These need to be qualified by "oracle" somehow, such as "oracleRman" or "oracle_rman". Maybe the latter since we use underscores in the labels to do "namespacing".

src/test/java/com/rackspace/salus/monitor_management/web/model/oracle/DataguardConversionTest.java
```diff
@@ -0,0 +1,105 @@
...
+    );
+
+    final DetailedMonitorOutput result = conversionService.convertToOutput(monitor);
+
+    final Dataguard dataguardPlugin = assertCommon(result, monitor, Dataguard.class, "convertToOutput");
+    assertThat(dataguardPlugin.getFilePath()).isEqualTo("./oracleDatabaseOutput");
+    final List<String> databaseNames = new LinkedList<>();
+    databaseNames.add("backupDB");
+    databaseNames.add("prodDB");
+    assertThat(dataguardPlugin.getDatabaseNames()).isEqualTo(databaseNames);
```
itzg: The benefit of AssertJ is that it has handy, type-specific assertions such as [containsExactlyInAnyOrder](http://joel-costigliola.github.io/assertj/core-8/api/org/assertj/core/api/ListAssert.html#containsExactlyInAnyOrder-ELEMENT...-). With that you don't need to create the temp list for the expected value.

src/test/java/com/rackspace/salus/monitor_management/web/model/oracle/DataguardConversionTest.java
```diff
@@ -0,0 +1,105 @@
...
+    final Map<String, String> labels = new HashMap<>();
+    labels.put("os", "linux");
+    labels.put("test", "convertFromInput");
+
+    final LocalMonitorDetails details = new LocalMonitorDetails();
+    final Dataguard plugin = new Dataguard();
+    plugin.setFilePath("./oracleDatabaseOutput");
+    final List<String> databaseNames = new LinkedList<>();
+    databaseNames.add("backupDB");
+    databaseNames.add("prodDB");
```
itzg: Since we're using Java 11 now, you can start switching these create-and-populate cases to [List.of](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html#of(E...))

PR #110, opened 2019-12-13
Use tcp as default protocol in NetResponse

jjbuchan:
I've had this change sitting locally for a while.

udp is not a common choice - I don't think ele even supports it?

Let's default to tcp if it nothing is specified.


harageth:
Is UDP the default?

itzg:
> Is UDP the default?

Apparently there isn't a default
https://github.com/influxdata/telegraf/blob/master/plugins/inputs/net_response/net_response.go#L217-L225

jjbuchan:
Right now if you don't include the protocol it'll fail with a 400.  adding tcp as the default avoids that.

PR #115, opened 2019-12-18
Add dns monitor

jjbuchan:
# Resolves

https://jira.rax.io/browse/SALUS-626

# What

Adds the ability to use the dns_query telegraf plugin type.

# Why

We need to migrate ele remote.dns checks.

# TODO

This adds the raw input plugin details, but I'll follow up with some monitor translation PRs to make it more customer friendly.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Dns.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Dns.java
@@ -19,34 +19,35 @@ package com.rackspace.salus.monitor_management.web.model.telegraf;
 import com.rackspace.salus.monitor_management.web.model.ApplicableAgentType;
 import com.rackspace.salus.monitor_management.web.model.ApplicableMonitorType;
 import com.rackspace.salus.monitor_management.web.model.Protocol;
+import com.rackspace.salus.monitor_management.web.model.RecordType;
 import com.rackspace.salus.monitor_management.web.model.RemotePlugin;
 import com.rackspace.salus.monitor_management.web.model.validator.ValidGoDuration;
 import com.rackspace.salus.monitor_management.web.model.validator.ValidHostAndPort;
 import com.rackspace.salus.telemetry.model.AgentType;
 import com.rackspace.salus.telemetry.model.MonitorType;
+import java.util.List;
 import javax.validation.constraints.NotNull;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 
 @Data @EqualsAndHashCode(callSuper = false)
 @ApplicableAgentType(AgentType.TELEGRAF)
-@ApplicableMonitorType(MonitorType.net_response)
-public class NetResponse extends RemotePlugin {
+@ApplicableMonitorType(MonitorType.dns)
+public class Dns extends RemotePlugin {
+
+  List<String> servers;
+
+  List<String> domains;
 
   @NotNull
-  Protocol protocol = Protocol.tcp;
+  Protocol network = Protocol.udp;
 
   @NotNull
-  @ValidHostAndPort
-  String address;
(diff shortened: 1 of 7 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/model/validator/ValidHostAndPort.java
```diff
@@ -33,7 +33,7 @@
 @Documented
 @Constraint(validatedBy = ValidHostAndPortValidator.class)
 public @interface ValidHostAndPort {
-  String message() default "Must be a host:port";
+  String message() default "Must be a host:port or metadata";
```
itzg: A nit, but technically it's a "template variable", which in turn typically reference metadata, but could also be any field of the resource.

jjbuchan: Thanks.  Fixed.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Dns.java
```diff
@@ -0,0 +1,53 @@
...
+
+  List<String> servers;
+
+  List<String> domains;
+
+  @NotNull
+  Protocol network = Protocol.udp;
+
+  @NotNull
+  RecordType recordType = RecordType.A;
```
jjbuchan: I just noticed for HTTP methods instead of an enum we used
```
@Pattern(regexp = "GET|PUT|POST|DELETE|HEAD|OPTIONS|PATCH|TRACE", message = "invalid http method")
```

@GeorgeJahad @harageth do you have any preference between enums vs. Patterns?  Based on the responses below I'd lean towards the pattern (but I still don't like all the inconsistency in our responses).

Response for invalid pattern
```
{
  "timestamp": "2019-12-20T16:39:39.132+0000",
  "status": 400,
  "error": "Bad Request",
  "errors": [
    {
      "codes": [
        "Pattern.detailedMonitorInput.details.plugin.method",
        "Pattern.details.plugin.method",
        "Pattern.method",
        "Pattern.java.lang.String",
        "Pattern"
      ],
      "arguments": [
        {
          "codes": [
            "detailedMonitorInput.details.plugin.method",
            "details.plugin.method"
          ],
          "arguments": null,
          "defaultMessage": "details.plugin.method",
          "code": "details.plugin.method"
        },
        [],
        {
          "arguments": null,
          "defaultMessage": "GET|PUT|POST|DELETE|HEAD|OPTIONS|PATCH|TRACE",
          "codes": [
            "GET|PUT|POST|DELETE|HEAD|OPTIONS|PATCH|TRACE"
          ]
        }
      ],
      "defaultMessage": "invalid http method",
      "objectName": "detailedMonitorInput",
      "field": "details.plugin.method",
      "rejectedValue": "GET2",
      "bindingFailure": false,
      "code": "Pattern"
    }
  ],
  "message": "Validation failed for object='detailedMonitorInput'. Error count: 1",
  "path": "/api/tenant/aaaaaa/monitors",
  "host": "MV906GG8WL"
}
```

Response for invalid enum
```
{
  "timestamp": "2019-12-20T16:42:26.577+0000",
  "status": 400,
  "error": "Bad Request",
  "message": "Failed to parse JSON at com.rackspace.salus.monitor_management.web.model.DetailedMonitorInput[\"details\"]->com.rackspace.salus.monitor_management.web.model.RemoteMonitorDetails[\"plugin\"]->com.rackspace.salus.monitor_management.web.model.telegraf.Dns[\"recordType\"]",
  "host": "MV906GG8WL"
}
```

itzg: I would have preferred enums, but yeah the validation response is apparently terrible for enum fields.

PR #118, opened 2019-12-23
Modify monitor models based on their translations

jjbuchan:
# What

Modifies the monitor models so that the user can provide more friendly fields and we will translate them on the backend to what telegraf expects.

# How

The changes correspond to the translations in https://github.com/Rackspace-Segment-Support/salus-data-loader-content/pull/1

## How to test

Tests currently pass, but I'm not sure if everything is set up correctly with the dataloader where all these PRs can be merged yet.

# Why

There are some fields that we do not want the user to set or we want them to provide a single value instead of a list.  This allows for that.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Disk.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Disk.java
@@ -16,14 +16,12 @@
 
 package com.rackspace.salus.monitor_management.web.model.telegraf;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.rackspace.salus.monitor_management.web.model.ApplicableAgentType;
 import com.rackspace.salus.monitor_management.web.model.ApplicableMonitorType;
 import com.rackspace.salus.monitor_management.web.model.LocalPlugin;
 import com.rackspace.salus.telemetry.model.AgentType;
 import com.rackspace.salus.telemetry.model.MonitorType;
-import java.util.Arrays;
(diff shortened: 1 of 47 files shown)


itzg:
Regarding your question, these two deployment issues need to be completed:
https://jira.rax.io/browse/SALUS-726
https://jira.rax.io/browse/SALUS-727

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Disk.java
```diff
@@ -31,7 +31,7 @@
 @ApplicableAgentType(AgentType.TELEGRAF)
 @ApplicableMonitorType(MonitorType.disk)
 public class Disk extends LocalPlugin {
-  List<String> mountPoints;
+  String mount;
```
itzg: I personally like that telegraf allows for multiple mount paths (and devices in diskio); however, is this change to better align with the single-target of MaaS checks? Migration can always create one monitor-per-check with single-value lists.

jjbuchan: Just noting that we talked about this and agree to leave things as is.

We'll follow the 1 monitor per target format.  It should make the managing of monitors by customers/support a little simpler and may have implications in terms of ticketing and how that can be handled by support or automation.

jjbuchan:
Fixed a conflict in `src/test/resources/ConversionTests/MonitorConversionServiceTest_http.json`

PR #119, opened 2019-12-30
Update disk monitors and add metrics to translate errors

jjbuchan:
# What

Branched off of https://github.com/racker/salus-telemetry-monitor-management/pull/118

Removes some disk/diskio fields we shouldn't need to expose and handled the exceptions thrown by https://github.com/racker/salus-telemetry-model/pull/95

# Why

By logging metrics for the exceptions we'll be able to know whether something has gone wrong and can look deeper into the logs if needed.  When it comes to setting up monitoring we'll have to decide on an adequate threshold of errors.  Currently any errors in the translator will occur silently - the monitor will not be bound but no one would be aware unless they really looked.

Also, `@NotBlank` was added to the disk mount because if an empty value is provided the telegraf plugin would monitor all disks.  We have made the decision where each individual monitor will monitor one target only.

### DIFF
--- a/src/test/java/com/rackspace/salus/monitor_management/services/MeterRegistryTestConfig.java
+++ b/src/test/java/com/rackspace/salus/monitor_management/services/MeterRegistryTestConfig.java
@@ -16,20 +16,17 @@
 
 package com.rackspace.salus.monitor_management.services;
 
-import com.rackspace.salus.telemetry.entities.BoundMonitor;
-import com.rackspace.salus.telemetry.translators.MonitorTranslator;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import org.springframework.boot.test.context.TestConfiguration;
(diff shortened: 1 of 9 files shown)


src/test/java/com/rackspace/salus/monitor_management/services/MeterRegistryTestConfig.java
```diff
@@ -16,20 +16,17 @@
...
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.context.annotation.Bean;
 
-/**
- * Indicates that an issue was encountered with either a {@link MonitorTranslator} or the
- * {@link BoundMonitor}'s rendered content that prevented translation.
- */
-class MonitorContentTranslationException extends Exception {
+@TestConfiguration
+public class MeterRegistryTestConfig {
```
itzg: It's funny how git thought this was a rename of the exception's file.

jjbuchan: Yeah, odd.

PR #120, opened 2020-01-06
Add exclude resources option during monitor binding

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-695

# What

Adds an `excludedResourceIds` field to monitors which allows for a monitor to be selectively excluded from binding to the given resource that would be normally be selected by labels.

# How

Tweaks the binding processing that happens during monitor creation and updates and resource events. There was also adjust model/DTO and validation changes needed to ensure that users can't improperly combine excludedResourceIds with a resourceId selector.

## How to test

The following primary scenarios are covered by the new tests
  - on monitor update, new exclusions
  - on monitor update, removed exclusions
  - on monitor update, new labels select resource that should be excluded
  - on monitor update, new labels select resource that should be excluded
  - on resource event
  - invalid update, was using resource ID selector, now adding exclusions
  - invalid update, was using exclusions, now adding resource ID selector

Speaking of which, all of the new tests were added to a new test file to avoid further bloating the infamous `MonitorManagementTest` file. I also wanted to make use of the embedded database rather than mock all the repository interactions.

# Depends on

https://github.com/racker/salus-telemetry-model/pull/96

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1023,10 +1023,15 @@ public class MonitorManagement {
   }
 
   private static boolean excludedResourceIdsChanged(Set<String> updated, Set<String> prev, boolean patchOperation) {
-    if (patchOperation) {
-      return updated == null || !updated.equals(prev);
+    // JPA will populate the absence of values as an empty set
+    if (prev != null && prev.isEmpty()) {
+      // ...but let's simplify comparison by treating as null
+      prev = null;
     }
-    return updated != null && !updated.equals(prev);
+    if (patchOperation) {
+      return !Objects.equals(updated, prev);
+    }
+    return updated != null && !Objects.equals(updated, prev);
   }
 
   private boolean labelSelectorChanged(Monitor original, MonitorCU newValues, boolean patchOperation) {
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Rackspace US, Inc.
+ * Copyright 2020 Rackspace US, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@ package com.rackspace.salus.monitor_management.services;
 import static com.rackspace.salus.telemetry.entities.Monitor.POLICY_TENANT;
 import static com.rackspace.salus.telemetry.etcd.types.ResolvedZone.createPrivateZone;
 import static com.rackspace.salus.telemetry.etcd.types.ResolvedZone.createPublicZone;
+import static org.springframework.util.CollectionUtils.isEmpty;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.google.common.collect.Streams;
@@ -84,6 +85,7 @@ import javax.annotation.Nullable;
 import javax.persistence.EntityManager;
 import javax.persistence.PersistenceContext;
 import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -98,7 +100,6 @@ import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.stereotype.Service;
 import org.springframework.util.Assert;
-import org.springframework.util.CollectionUtils;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
 
@@ -269,6 +270,7 @@ public class MonitorManagement {
         .setMonitorType(newMonitor.getMonitorType())
         .setLabelSelector(newMonitor.getLabelSelector())
         .setResourceId(newMonitor.getResourceId())
+        .setExcludedResourceIds(newMonitor.getExcludedResourceIds())
         .setInterval(newMonitor.getInterval())
         .setContent(newMonitor.getContent())
         .setAgentType(newMonitor.getAgentType())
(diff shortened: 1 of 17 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1155,6 +1233,25 @@ private static boolean labelSelectorMethodChanged(Monitor original, MonitorCU ne
...
+  private List<ResourceDTO> findResourcesByLabels(String tenantId,
+                                                  Map<String, String> labelSelector,
+                                                  @NotNull LabelSelectorMethod labelSelectorMethod,
+                                                  Set<String> excludedResourceIds) {
+    List<ResourceDTO> selectedResources;
+    selectedResources = resourceApi
+        .getResourcesWithLabels(tenantId, labelSelector, labelSelectorMethod)
+        .stream()
+        // filter away resources in the exclusion set
+        .filter(resourceDTO -> excludedResourceIds == null ||
```
jjbuchan: `filter` always confuses me.  I always expect it to remove anything that is true, not keep it.

jjbuchan: ...nothing to change here; just me complaining :)

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -753,14 +763,67 @@ public Monitor updateMonitor(String tenantId, UUID id, @Valid MonitorCU updatedV
...
+      affectedEnvoys.addAll(
+          processExcludedResourceIdsModified(tenantId, monitor,
+              updatedValues.getExcludedResourceIds()
+          )
+      );
+    } else if (monitor.getExcludedResourceIds() != null) {
+      // See above regarding:
+      // JPA's EntityManager is a little strange with re-saving (aka merging) an entity
+      monitor.setExcludedResourceIds(new HashSet<>(monitor.getExcludedResourceIds()));
+    }
```
jjbuchan: This whole block could be in a `processExcludedResourceChange` method instead?

jjbuchan: and could it be moved up with `processLabelSelectorChange` and `processResourceIdChange`?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -753,14 +763,67 @@ public Monitor updateMonitor(String tenantId, UUID id, @Valid MonitorCU updatedV
...
 
     return monitor;
   }
 
-  private Set<String> handleLabelSelectorChange(String tenantId, Monitor monitor, MonitorCU updatedValues, boolean patchOperation) {
+  /**
+   * @return the affected envoy IDs
+   */
+  private Set<String> processExcludedResourceIdsModified(String tenantId, Monitor monitor,
+                                                         Set<String> updatedExcludedResourceIds) {
```
jjbuchan: I'm not sure if I like this.

It feels like we're redoing what could/should have already been done in the processing of changed labels?

We might bind something there then immediately unbind it here?

Is it better for those two methods to be combined?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -753,14 +763,67 @@ public Monitor updateMonitor(String tenantId, UUID id, @Valid MonitorCU updatedV
...
+    resourceApi
+        .getResourcesWithLabels(
+            tenantId, monitor.getLabelSelector(), monitor.getLabelSelectorMethod())
+        .stream()
+        // filter away ones that would become excluded anyway
+        .filter(resourceDTO ->
+            updatedExcludedResourceIds == null || !updatedExcludedResourceIds.contains(resourceDTO.getResourceId()))
+        // filter away ones that weren't excluded before
+        .filter(resourceDTO -> monitor.getExcludedResourceIds() == null ||
+            !monitor.getExcludedResourceIds().contains(resourceDTO.getResourceId()))
```
jjbuchan: I think this is doing the opposite of your comment?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
jjbuchan: I'm not sure we're handling the case of label selectors changing at the same time as exclusions?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -753,14 +763,67 @@ public Monitor updateMonitor(String tenantId, UUID id, @Valid MonitorCU updatedV
...
+      envoyIds.addAll(extractEnvoyIds(nowExcluded));
+
+      // JPA's EntityManager is a little strange with re-saving (aka merging) an entity
+      monitor.setExcludedResourceIds(new HashSet<String>(updatedExcludedResourceIds));
+    } else {
+      monitor.setExcludedResourceIds(null);
+    }
+
+    resourceApi
+        .getResourcesWithLabels(
```
jjbuchan: I also don't like potentially calling this twice for a monitor update - even if it is a rare occurrence.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -949,14 +1012,28 @@ private static boolean intervalChanged(Duration updatedInterval, Duration prevIn
   private static boolean zonesChanged(List<String> updatedZones, List<String> prevZones, boolean patchOperation) {
     if (patchOperation) {
       return updatedZones == null ||
+          // use size and containsAll to allow order difference
```
jjbuchan: 👍

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
itzg: I'll tweak the comment there actually. There was another one in there where I too got myself confused about the "direction" of `filter()` and worded the comment in the "keeping" sense.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
itzg: Yeah, I noticed we weren't totally consistent with these kinds of blocks. I think it needs to stay separate from the two conditions you mentioned, but I'll look at how to combine this more like those.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
itzg: Good point. I'll double check there's not clashing bind/unbind happening between the two.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
itzg: I'll reword that comment too.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
itzg: Agreed. I'll include that with investigating the other comment.

itzg:
@harageth latest commit includes the consolidated changes from @jjbuchan 's comment.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1155,6 +1199,23 @@ private static boolean labelSelectorMethodChanged(Monitor original, MonitorCU ne
     return affectedEnvoys;
   }
 
+  /**
+   * Queries the resource API for resource selected by the given label selector and also applies
+   * resource ID exclusions specified by the monitor.
+   */
+  private List<ResourceDTO> findResourcesByLabels(String tenantId,
```
jjbuchan: Not something to add here, but reading over this makes me think we should have a `findResourcesForMonitor` method.  I don't think it exists already.

I'm not sure what all the use-cases it'd have other than for Intelligence to consume that raw list.

jjbuchan: I'd assume it'd either call this or the find by resource id method.

itzg: Actually, that's basically what this method became and I like that method name better than the one I'm using there.

jjbuchan: As mentioned in slack `findResourcesForMonitor` would be a method that either calls this or (something like) `findResourceById` depending on which of resourceId or labelSelector is set.

I think there's still a benefit of having this as it's own thing vs. consolidating everything in there across the board.

PR #121, opened 2020-01-09
Support polymorphic metadata values

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-491

# What

# How

Given that `Resource`'s metadata field is changed to a `Map<String,Object>` this change alters the monitor content renderer to allow specific uses of placeholder variables, such as

```
   "some_config": "${resource.metadata.listOfSomething}"
```

to be replaced by a string (as it is currently), number, array, or object. For a list that would become:

```
    "some_config": ["item1","item2"]
```

## NOTE

As of https://github.com/racker/salus-telemetry-monitor-management/pull/118 this feature is not as useful since there are any remaining list-type fields since we realized they need to be scalars to handle downstream correlation of events, tickets, etc.

## How to test

Unit test updated

# Why

<Why was the final approach taken? Were alternatives considered?>

# TODO

<outstanding tasks before this PR is considered 'ready'>

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentRenderer.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentRenderer.java
@@ -93,10 +93,9 @@ public class MonitorContentRenderer {
     //
     //   "${resource.metadata.url}"
     //
-    // It then has a matching group to group the is best explained with an example:
     // if the delimiters is configured with the following, note the required space between delimiters
     //   ${ }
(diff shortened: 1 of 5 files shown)


itzg:
Related changes, not yet PR'ed are

- https://github.com/racker/salus-telemetry-model/compare/itzg/polymorphic-metadata-values?expand=1
- https://github.com/racker/salus-telemetry-resource-management/compare/itzg/polymorphic-metadata-values?expand=1

PR #122, opened 2020-01-09
Tweak order of translation operations

jjbuchan:
# What

1. Changes the field in http monitors from `address` to `url`
1. Renamed `translateOne` to `translateBoundMonitor`
1. Updates the order things are done in translations

# How

1. Renamed things
2. Renamed things
3. Get the list of all operations to be performed on a bound monitor and pass that to `translateBoundMonitor` instead of having it receive all and then decide what ones should be executed or not.

## How to test

All tests still pass?

# Why

1. For the http change, `address` was what the old telegraf version required.  Now we can use whatever we like and I just like `url` more.

2. When reading the code I wasn't sure if `One` referred to the number of operations being translated at a time or the number of bound monitors.  This might be clearer?

3. For the translation change, previously due to the `renameType` operations the order the operations were performed in could have caused problems.  i.e. if the type was changed prior to other operations being ran, their `type` test would have failed and they would have been skipped.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
@@ -121,9 +121,9 @@ public class MonitorContentTranslationService {
     return boundMonitors.stream()
         .map(boundMonitor -> {
           try {
-            return translateOne(
+            return translateBoundMonitor(
                 boundMonitor,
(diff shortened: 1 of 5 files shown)


jjbuchan:
@itzg I figured I should add a test for the translation order... and upon doing so realized it still passed before my changes.  I now see everything was fine before since within the loop of translate operations it was only modifying the `contentTree` and the if condition was always checking the original monitor object's fields.  It would only set the new contentTree onto the monitor once all were complete.

Oh well, still a little cleaner now I think but not exactly needed.

PR #126, opened 2020-01-15
Add apache monitor

jjbuchan:
# What

Adds the apache monitor that will be tied to https://github.com/influxdata/telegraf/tree/master/plugins/inputs/apache

# How

Copied a previous PR for adding a monitor.

Some fields of note:

I set `insecureSkipVerify` to `Boolean` and not `boolean` so it defaults to null.  It only needs to be set if the other tls options are included so a default of null is what we want.

Instead of `responseTimeout` I used `timeout`.  Having a consistent timeout field across our monitors seems sane.

Here is my previous discovery info about the telegraf plugin that shows the ele metrics do map to these new ones - https://github.com/Rackspace-Segment-Support/salus-docs-internal/blob/master/migration/ele/check-to-monitor-mapping.md#agentapache

# Why

We need it for the migration.

# TODO

Add translations.
Another PR that tweaks some other monitors' fields.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Apache.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Apache.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2020 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.rackspace.salus.monitor_management.web.model.telegraf;
+
+import com.rackspace.salus.monitor_management.web.model.ApplicableAgentType;
+import com.rackspace.salus.monitor_management.web.model.ApplicableMonitorType;
+import com.rackspace.salus.monitor_management.web.model.LocalPlugin;
+import com.rackspace.salus.monitor_management.web.model.validator.ValidGoDuration;
+import com.rackspace.salus.telemetry.model.AgentType;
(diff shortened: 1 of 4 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Apache.java
```diff
@@ -0,0 +1,44 @@
...
+  @NotBlank
+  String url;
+  String username;
+  String password;
+  @ValidGoDuration
+  String timeout;
+  String tlsCa;
+  String tlsCert;
+  String tlsKey;
+  Boolean insecureSkipVerify;
```
itzg: I wonder if this needs a `@JsonInclude(Include.NON_NULL)` to fully have the desired effect when not set.

jjbuchan: I don't think so.

Doesn't that mean it would only (de)serialize the field and include it in the response if it was non null?  If a customer requests their monitors we want to include this field whether it is null or not.

This comment made me look up where we use it right now and in Procstat is the only place - I think we need to remove it from there?

itzg: I was thinking more of the JSON encoding that ends up at Envoy and then converted to TOML for telegraf. You would need to confirm that the resulting TOML doesn't also convey a `null` but instead the field is not set at all.

jjbuchan: Hmm... I was assuming it did ignore it but I'll try confirm that.  I'd have thought it would be a problem for any field not just Boolean ones so feel confident it must be fine.

If it doesn't, we'd probably have to handle that a different way.

itzg: Looks like nil values could be normalized away here
https://github.com/racker/salus-telemetry-envoy/blob/a8dc055f185f0ed1a71c198359f1e1d042b3c3a9/agents/converters.go#L94-L106

jjbuchan: Thanks.

I was just looking in that direction to add some extra tests.

jjbuchan: I think this proves no changes are needed? - https://github.com/racker/salus-telemetry-envoy/pull/71

PR #127, opened 2020-01-15
Fix small things in monitor models

jjbuchan:
# What

* Sets `@EqualsAndHashCode(callSuper = false)` instead of `callSuper = true`
   * There's no reason to call super and it can cause problems in tests
* Change http timeout field
* Use `Boolean` for skipping insecure things
   * That field is only used if the other tls settings are provided so a null default is best.
* Add `insecureSkipVerify` to a few places it was missing.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
@@ -28,7 +28,7 @@ import javax.validation.constraints.Pattern;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 
-@Data @EqualsAndHashCode(callSuper = true)
+@Data @EqualsAndHashCode(callSuper = false)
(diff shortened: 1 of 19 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/HttpResponse.java
```diff
@@ -45,6 +45,6 @@
   String tlsCa;
   String tlsCert;
   String tlsKey;
-  boolean insecureSkipVerify;
+  Boolean insecureSkipVerify;
```
itzg: Like other review, I think this might need `@JsonInclude(NON_NULL)` to ensure it is absent rather than a `null` value.

jjbuchan: I think this shows nothing is needed for it to be excluded from toml

https://github.com/racker/salus-telemetry-envoy/pull/71

PR #132, opened 2020-01-17
Fix ping timeout type

jjbuchan:
# What

Changes the `timeout` type from Integer to String of `@ValidGoDuration`.

All other `timeout` parameters use this.  I'm not certain that there aren't other fields that need to change though.

# Why

We need to pass the units down to telegraf otherwise it looks like it will use the default which @itzg says is nanoseconds.

# TODO

Whatever monitors are configured in dev/stage/prod right now might need to be recreated to get the correct format... but I don't think anything will completely break since those details are only stored in the string content - it can still (de)serialize that value to a string.


src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Ping.java
```diff
@@ -33,7 +34,8 @@
   String target;
   Integer count;
   Integer pingInterval;
-  Integer timeout;
+  @ValidGoDuration
+  String timeout;
```
harageth: Why not use the same format here as for the interval?

jjbuchan:
Closing in favor of https://github.com/racker/salus-telemetry-monitor-management/pull/133

PR #133, opened 2020-01-20
Use Duration for any time based fields

jjbuchan:
# What

Updates any time field to Duration from the existing int or string.

# Why

To add consistency to the API.

# TODO

The policy metadata durations should probably be changed to expect the string format of duration?  You can specify something like `0.3` via the api to represent 300ms but to achieve that via metadata we'd have to store the value as `300`ms since `Duration.ofSeconds(0.3)` is not allowed.  We should instead set `PT300MS`.

Translations need to be added for each monitor to convert these values to their correct format.  Some monitors use timeouts in the format `5s` whereas others would need `5.0` or `5`.


src/test/java/com/rackspace/salus/monitor_management/utils/MetadataUtilsTest.java
```diff
@@ -148,23 +148,23 @@ public void setNewMetadataValues_plugin() {
...
-    assertThat(plugin.getPingInterval()).isEqualTo(67);
+    assertThat(plugin.getPingInterval()).isEqualTo(Duration.ofSeconds(67));
   }
 
   @Test
   public void setUpdateMetadataValue_INT() {
     Ping plugin = new Ping();
     MonitorMetadataPolicyDTO policy = (MonitorMetadataPolicyDTO) new MonitorMetadataPolicyDTO()
-        .setKey("pingInterval")
+        .setKey("count")
```
GeorgeJahad: was there a reason why the keys/values were changed in this test?

src/test/java/com/rackspace/salus/monitor_management/web/model/telegraf/X509ConversionTest.java
```diff
@@ -126,10 +127,10 @@ public void convertToOutput_x509() throws IOException {
     validatorFactoryBean.afterPropertiesSet();
     Set<ConstraintViolation<X509Cert>> violations = validatorFactoryBean.validate(x509Plugin);
     assertEquals(violations.size(), 0);
-    x509Plugin.setTimeout("xx");
```
GeorgeJahad: why the change from timeout to target?

jjbuchan: Because "xx" cannot be set in the timeout field since it has been changed to be a Duration. This step is testing for violations and in this case a violation is something that is of the correct data type but will fail due to restrictions on it via annotations.

itzg: And `target` is the only remaining field with a validation to verify, right?

jjbuchan: Correct

PR #135, opened 2020-01-20
Added plugin schema for packages agent

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-594

# What

Adds a plugin definition that aligns with the JSON config consumed by the new packages agent:

https://github.com/racker/salus-packages-agent/blob/3c23a7d6947c16a670e9680168f5ccf3af6d1a0f/config.go

## How to test

Added a conversion test


harageth:
What is the packages agent sending back through the pipeline to be alarmed on?

harageth:
> Technically...if we mapped string/regex operations in the TICKscript a user could devise a event task that "alerts" on say a matching version of a package, etc.
And that's kind of what I was expecting to hear. Just wanted to be sure I properly understood what this packages agent was being used for

PR #136, opened 2020-01-21
Add redis monitor

jjbuchan:
# What

Adds the ability to use the redis telegraf plugin type.

# Why

We need to migrate ele remote.redis checks.

# TODO

Add monitor translations

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Redis.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Redis.java
@@ -18,7 +18,7 @@ package com.rackspace.salus.monitor_management.web.model.telegraf;
 
 import com.rackspace.salus.monitor_management.web.model.ApplicableAgentType;
 import com.rackspace.salus.monitor_management.web.model.ApplicableMonitorType;
-import com.rackspace.salus.monitor_management.web.model.RemotePlugin;
+import com.rackspace.salus.monitor_management.web.model.LocalPlugin;
 import com.rackspace.salus.telemetry.model.AgentType;
 import com.rackspace.salus.telemetry.model.MonitorType;
 import javax.validation.constraints.NotEmpty;
@@ -28,7 +28,7 @@ import lombok.EqualsAndHashCode;
 @Data @EqualsAndHashCode(callSuper = false)
 @ApplicableAgentType(AgentType.TELEGRAF)
 @ApplicableMonitorType(MonitorType.redis)
-public class Redis extends RemotePlugin {
+public class Redis extends LocalPlugin {
   @NotEmpty
   String url;
   String password;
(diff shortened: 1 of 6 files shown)


jjbuchan:
I pushed https://github.com/racker/salus-telemetry-monitor-management/pull/136/commits/a19a4a4573920e5a7abe34bca2b3adaf165ec3ed since our existing check is currently agent.redis not remote.redis.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Redis.java
```diff
@@ -0,0 +1,39 @@
...
+import javax.validation.constraints.NotEmpty;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+@Data @EqualsAndHashCode(callSuper = false)
+@ApplicableAgentType(AgentType.TELEGRAF)
+@ApplicableMonitorType(MonitorType.redis)
+public class Redis extends LocalPlugin {
+  @NotEmpty
+  String url;
```
itzg: I think this now needs `@ValidLocalHost` since it is a "local" plugin

jjbuchan:
Also, I haven't added any validation to the model but it's worth pointing out this section of the plugin in case anyone else has a preference - https://github.com/influxdata/telegraf/blob/master/plugins/inputs/redis/redis.go#L104-L107

tcp/unix prefixes are required for the url.  If one is not provided the plugin will prepend `tcp://`.  In ele, checks are configured with only the hostname - no prefix.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Redis.java
jjbuchan: I'll add that, but as a separate issue I think that validation needs to be expanded to include any local subnet.  i.e. 172 and 192 addresses.

jjbuchan: and based on what's configured right now in ele, we're the only one that wouldn't meet this criteria.... since we just plug in the hostname of the server

itzg: I would have considered those addresses covered by a remote, but private zone monitor.

itzg: > and based on what's configured right now in ele, we're the only one that wouldn't meet this criteria.... since we just plug in the hostname of the server

If that's the case, then sounds like a remote copy of the plugin is already needed.

jjbuchan: > If that's the case, then sounds like a remote copy of the plugin is already needed.

I've not looked but I'm assuming we're running those checks on the same host as is configured in the check.  It's just chef plugging in the current host when it runs, so essentially the same as localhost.

> I would have considered those addresses covered by a remote, but private zone monitor.

Hmmmmm, maybe.

PR #138, opened 2020-01-22
Adam9500/fix label matching

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-737

# What

Make sure that when a monitor is created with an empty list of labels and no resourceId set that we should still match to it.

# How

It alters the SQL query to not grab any monitors with the resourceId set and does an outer join and looks for NULL values where a monitor wouldn't have had any label values

## How to test

Run the MonitorManagement tests

# Why

@itzg rightly thought an outer join was needed in the query to make it work. It has to be done in the query because thats telling the java server what monitors to even look at. So we would have no way of checking if a monitor doesn't have any labels there.

# TODO

Everything should be done


src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -1782,13 +1784,48 @@ public void testGetMonitorsFromLabels() {
...
+
+    // Create a "select all" type of monitor where label selector is empty
+    createMonitorsForTenant(1, tenantId, Collections.emptyMap());
+
+    MonitorCU create = podamFactory.manufacturePojo(MonitorCU.class);
+    create.setSelectorScope(ConfigSelectorScope.LOCAL);
+    create.setZones(Collections.emptyList());
+    create.setResourceId(RandomStringUtils.randomAlphabetic(10));
+    String tenantId2 = RandomStringUtils.randomAlphabetic(10);
+    monitorManagement.createMonitor(tenantId2, create);
```
jjbuchan: Why are you doing this when you already create things above?

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -1782,13 +1784,48 @@ public void testGetMonitorsFromLabels() {
...
+    Map<String, String> labels = Collections.singletonMap("mykey", "myvalue");
+
+    // Create monitors which don't have the labels we care about
+    createMonitorsForTenant(monitorsWithoutLabels, tenantId);
+
+    // Create monitors which do have the labels we care about
+    createMonitorsForTenant(monitorsWithLabels, tenantId, labels);
+
+    // Create a "select all" type of monitor where label selector is empty
+    createMonitorsForTenant(1, tenantId, Collections.emptyMap());
```
jjbuchan: Maybe this could be a random number too?

harageth: That seems reasonable

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
harageth: I needed a monitor that had a resourceId set on a different tenant.

src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
```diff
@@ -14,23 +15,23 @@ AND      monitors.id IN
...
-        SELECT monitor_label_selectors.monitor_id
-        FROM monitor_label_selectors
-        WHERE %s
-        GROUP BY monitor_label_selectors.monitor_id
-        HAVING COUNT(*) >= 1
-      )
- ))
+      SELECT inner_monitors.id
+      FROM   monitors AS inner_monitors
+      WHERE  tenant_id = :tenantId
```
GeorgeJahad: there seems to be a lot of tenant id code in this file.  can it be simplified with something like a common table expression?  https://dev.mysql.com/doc/refman/8.0/en/with.html#common-table-expressions

src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
```diff
@@ -14,23 +15,23 @@ AND      monitors.id IN
     WHERE  tenant_id = :tenantId
   )
   AND monitors.id IN
+  (
+    SELECT monitor_label_selectors.monitor_id
```
itzg: After switching to hiding whitespace changes, this looked like what I expected :)

src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
harageth: There are only two references to tenantId in each SQL file. And ultimately using a common table expression would only move one of those references to a different location.

GeorgeJahad: fair enough.

still the tenant id code looks redundant to me, since the tenant id check has been added to line 5, can't you remove the others?

something like this:

```
SELECT              monitors.id
FROM                monitors
LEFT OUTER JOIN     monitor_label_selectors AS ml ON monitors.id = ml.monitor_id
where               monitors.label_selector_method = 'OR'
AND                 monitors.tenant_id = :tenantId
AND                 monitors.resource_id IS NULL
AND                 (ml.monitor_id IS NULL OR monitors.id IN
    (
      SELECT   monitor_label_selectors.monitor_id
      FROM     monitor_label_selectors
      WHERE    %s
      GROUP BY monitor_label_selectors.monitor_id
      HAVING COUNT(*) >= 1
    )
)
ORDER BY monitors.id
```

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
jjbuchan: I guess it also seemed odd that you used this method vs. using `createMonitorsForTenant`

jjbuchan: Would it make more sense to have this use labels vs. resourceId?  Or maybe you need some of both.

src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
harageth: I think this is reasonable. 

My original thought is that the inner query tells the query planner that we do want to limit the monitor_label_selectors table based on the tenant_id. I would need to run it through the query planner to have it tell me though. On the other hand its an optimization and maintainability is probably more important right now.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
harageth: We can't use `createMonitorsForTenant` here because it is expecting to set labels rather than the resourceId (which isn't to say that I shouldn't have created a new overloaded function for setting the resourceId). The resourceId needs to be set so that we can test and make sure that `getMonitorsFromLabels` is not returning monitors that have a resourceId set.

src/main/resources/sql-queries/monitor_label_matching_OR_query.sql
itzg: Sorry I'm late to the party on this one. It might be tough to get too fancy with the syntax here if h2 doesn't support the common table expressions...I haven't researched that yet.

harageth: @itzg thats a good point. I didn't even think about h2 support. This did lead me down a thought about stored procedures, which makes me wonder if we would have to change anything for initialization of tests to load any expected stored procedures down the road.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1728,7 +1728,7 @@ private void decrementBoundCounts(List<BoundMonitor> needToDelete) {
    */
   @SuppressWarnings("Duplicates")
   public Page<Monitor> getMonitorsFromLabels(Map<String, String> labels, String tenantId, Pageable page) {
-    if(labels.size() == 0) {
+    if(labels == null || labels.isEmpty()) {
```
itzg: Thanks. I had already re-forgotten about that.

harageth: Yeah I remembered it while debugging yesterday afternoon.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -1782,13 +1798,46 @@ public void testGetMonitorsFromLabels() {
     // Create monitors which do have the labels we care about
     createMonitorsForTenant(monitorsWithLabels, tenantId, labels);
 
+    // Create a "select all" type of monitor where label selector is empty
+    createMonitorsForTenant(1, tenantId, Collections.emptyMap());
```
itzg: For consistency, this should use a random number of monitors, but I'm fine with a fixed count.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -1782,13 +1798,46 @@ public void testGetMonitorsFromLabels() {
...
+    int monitorsWithLabels = new Random().nextInt(10) + 10;
+    int monitorsWithoutLabels = new Random().nextInt(10) + 20;
+    int monitorsThatMatchEverything = new Random().nextInt(10) + 20;
+    int monitorsWithResourceId = new Random().nextInt(10) + 20;
+    String tenantId = RandomStringUtils.randomAlphabetic(10);
+
+    Map<String, String> labels = Collections.singletonMap("mykey", "myvalue");
+
+    // Create monitors which don't have the labels we care about
+    createMonitorsForTenant(monitorsWithoutLabels, tenantId);
```
jjbuchan:
```suggestion
    createMonitorsForTenant(monitorsWithoutLabels, tenantId, Map.of("label", "no-match");
```

Would something like this make more sense?  Currently this seems to be leaving `labelSelector` as null or empty?  and that makes me wonder why the test is passing.

harageth: From debugging through this I can confirm that the testing harness is actually adding random values for the labels here. I can certainly make it clearer by doing what you suggest and will update the `monitorsWithoutLabels` to something more relevant. Or I could just update `monitorsWithoutLabels` and leave it at that.

GeorgeJahad: I'm guessing it's matching because the podam is creating the dummy labels that don't match.  If that is the case maybe it is sufficient to the name of the count from "monitorsWithoutLabels"  to "monitorsWithNonMatchingLabels"?

GeorgeJahad: oops, didn't see @harageth's comment before i wrote mine.  sounds like we are basically saying the same thing.

jjbuchan: Oh right, I overlooked that podam was involved here.

jjbuchan: and yeah, a name change could be helpful

PR #141, opened 2020-01-30
Added a summary field to DetailedMonitorOutput

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-746

# What

Web frontends for Salus will need a way to convey the difference between several monitors all with the same type.

# How

Introduces a marker-annotation that is picked up by `MonitorConversionService` (not to be confused with "monitor translation") when building `DetailedMonitorOutput` objects returned from the monitor REST API.

## How to test

Updated existing tests to verify the annotated fields specific to each plugin type.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
@@ -31,21 +31,29 @@ import com.rackspace.salus.monitor_management.web.model.MonitorCU;
 import com.rackspace.salus.monitor_management.web.model.MonitorDetails;
 import com.rackspace.salus.monitor_management.web.model.RemoteMonitorDetails;
 import com.rackspace.salus.monitor_management.web.model.RemotePlugin;
+import com.rackspace.salus.monitor_management.web.model.SummaryField;
 import com.rackspace.salus.monitor_management.web.model.ValidationGroups;
 import com.rackspace.salus.policy.manage.web.model.MonitorMetadataPolicyDTO;
 import com.rackspace.salus.telemetry.entities.Monitor;
 import com.rackspace.salus.telemetry.model.ConfigSelectorScope;
 import com.rackspace.salus.telemetry.repositories.MonitorRepository;
+import java.beans.PropertyDescriptor;
 import java.io.IOException;
 import java.io.InvalidClassException;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.time.Duration;
 import java.time.format.DateTimeFormatter;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.UUID;
 import javax.json.JsonException;
 import javax.json.JsonPatch;
 import javax.validation.ConstraintViolationException;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
+import org.springframework.beans.BeanUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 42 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/model/SummaryField.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation is used to mark one or more fields of a {@link LocalPlugin} or {@link RemotePlugin}
+ * subclass to be included in the summary field of a {@link DetailedMonitorOutput}.
+ */
```
GeorgeJahad: It seems like we need a comment describing which fields in a plugin should be so annotated.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
```diff
@@ -122,9 +130,56 @@ public DetailedMonitorOutput convertToOutput(Monitor monitor) {
...
+          } else {
+            log.warn("Summary field {} was not readable from {}", fieldName, plugin);
+          }
+        } else {
+          log.warn("Summary field {} was not a proper property in {}", fieldName, plugin);
+        }
+      }
+    }
+
+    return summary;
```
GeorgeJahad: should there be a confirmation that size is non-zero?

src/main/java/com/rackspace/salus/monitor_management/web/model/SummaryField.java
itzg: Will do

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
itzg: No. Some monitors like cpu don't have any fields that are meaningful in the summary. I gave Faraji a heads up that this will be the case sometimes.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
```diff
@@ -122,9 +130,56 @@ public DetailedMonitorOutput convertToOutput(Monitor monitor) {
...
+    final Map<String, String> summary = new HashMap<>();
+
+    final Class<?> pluginClass = plugin.getClass();
+
+    for (Field field : pluginClass.getDeclaredFields()) {
+      if (field.getAnnotation(SummaryField.class) != null) {
+        final String fieldName = field.getName();
+
+        final PropertyDescriptor propertyDescriptor = BeanUtils
+            .getPropertyDescriptor(pluginClass, fieldName);
```
jjbuchan: Is this how you read the field without doing this? - https://github.com/racker/salus-telemetry-monitor-management/blob/6af95288eda2c79d444f29605d21991e18baf423/src/main/java/com/rackspace/salus/monitor_management/utils/MetadataUtils.java#L54-L56

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
```diff
@@ -122,9 +130,56 @@ public DetailedMonitorOutput convertToOutput(Monitor monitor) {
...
     return detailedMonitorOutput;
   }
 
+  private Map<String, String> buildSummaryFromDetails(MonitorDetails details) {
+    if (details instanceof LocalMonitorDetails) {
+      return buildSummaryFromPlugin(((LocalMonitorDetails) details).getPlugin());
+    } else if (details instanceof RemoteMonitorDetails) {
+      return buildSummaryFromPlugin(((RemoteMonitorDetails) details).getPlugin());
+    } else if (details == null) {
+      return Map.of();
```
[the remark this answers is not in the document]
itzg: Hmm, I added this condition because it did actually happen during unit tests. I have since forgotten when, so I'll remove and try to recreate.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
itzg: Yeah. The PropertyDescriptor approach is a bit more tedious, but leverages the getter for access. Your approach is probably more robust overall since I'm not sure what a public field exposes for the read method. I can change this to use Field#get for consistency.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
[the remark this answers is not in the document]
itzg: I fixed the tests rather than adding the awkward condition here.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
itzg: I went ahead and switched to the same Field#get approach since it is much simpler looking.

PR #142, opened 2020-02-04
Add boolean metadata handler and some other small things

jjbuchan:
# What / How / Why

Adds a `BOOL` metadata value handler so we can use metadata for boolean values.

Set `followRedirects` to `Boolean` instead of `boolean` so it can come through as null (and use metadata) rather than defaulting to false.

Made a Duration change following on from https://github.com/racker/salus-telemetry-monitor-management/pull/134.  Tests still provided the old `int` payload instead of the duration format so I never caught that I only updated half of the methods previously.

Added a test that hits the BOOL logic.

# Related PRs

https://github.com/Rackspace-Segment-Support/salus-data-loader-content/pull/10
https://github.com/racker/salus-telemetry-model/pull/111

### DIFF
--- a/src/test/java/com/rackspace/salus/monitor_management/utils/MetadataUtilsTest.java
+++ b/src/test/java/com/rackspace/salus/monitor_management/utils/MetadataUtilsTest.java
@@ -117,7 +117,7 @@ public class MetadataUtilsTest {
             .setValue("true")
     );
 
-    // url has a different value than the policy.  even though it was previously using metadata, it is now excluded
+    // url is set and has no policy value.  even though it was previously using metadata, it is now excluded
     // body has a different value than the policy, but is now null so will continue to use metadata regardless
     // timeout has the same value as the metadata and was previously set using it, so it remains as metadata
     // any field that was not previously metadata and is now null will be set to metadata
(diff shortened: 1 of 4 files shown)


src/test/java/com/rackspace/salus/monitor_management/utils/MetadataUtilsTest.java
```diff
@@ -95,6 +96,48 @@ public void getMetadataFieldsForUpdate_Ping() {
...
+            .setKey("timeout")
+            .setValueType(MetadataValueType.DURATION)
+            .setValue("PT12S"),
+        "followRedirects", (MonitorMetadataPolicyDTO) new MonitorMetadataPolicyDTO()
+            .setKey("followRedirects")
+            .setValueType(MetadataValueType.BOOL)
+            .setValue("true")
+    );
+
+    // url has a different value than the policy.  even though it was previously using metadata, it is now excluded
```
itzg: url is not in the policy and is set -- isn't that why it is excluded from the `getMetadataFieldsForUpdate` results?

jjbuchan: Yep.

I think both comments are accurate but I've tweaked it to be more like your wording.

PR #143, opened 2020-02-05
Fix unassigned bound monitor bug

jjbuchan:
# What

Fix a bug related to bound monitors not being assigned to pollers.

When a previously connected envoy reconnects look to see if there are any bound monitors that are not assigned to an envoy and then try to assign them.

# How

When a `ReattachedResourceZoneEvent` is seen also trigger the same logic that happens when  a `NewResourceZoneEvent` is seen.

## How to test

New tests + manually

# Why

If all pollers in a zone went down and new monitors were created during that time, the bound monitors would be created but not assigned to any envoy.  When the envoys reconnected their previously assigned bound monitors would be updated to attach to the same envoy as before, but any of the new monitors created during the time when they were down would be left with a `null` envoy id.

i.e. Bound Monitors with a `null` envoyId were only being assigned when a new envoy attached and not when a previously seen one attached.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -653,6 +653,10 @@ public class MonitorManagement {
     else {
       log.debug("No bound monitors were previously assigned to envoy={}", fromEnvoyId);
     }
+
+    // Also check to make sure there are no unassigned bound monitors
+    // This can happen if new monitors are created while all monitors in a zone are down
+    handleNewEnvoyInZone(tenantId, zoneName);
   }
 
   private ResolvedZone resolveZone(String tenantId, String zone) {
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -655,7 +655,7 @@ public class MonitorManagement {
     }
 
     // Also check to make sure there are no unassigned bound monitors
-    // This can happen if new monitors are created while all monitors in a zone are down
+    // This can happen if new monitors are created while all poller-envoys in a zone are down
     handleNewEnvoyInZone(tenantId, zoneName);
   }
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -2620,7 +2622,7 @@ public void testHandleZoneResourceChanged_publicZone() {
 
     // EXECUTE
 
-    // The main thing being tested is that a null zone tenant ID
+    // The main thing being tested is that a null zone tenant ID...
```
GeorgeJahad: "..."?

jjbuchan: this message continues below https://github.com/racker/salus-telemetry-monitor-management/pull/143/files#diff-c832012763cecfc29a5d4de8c6b48bccR2630

I got confused by the sudden end as well and was going to finish the sentence myself, then saw it was below so figured adding the `...` may at least show that its not a sudden end.

*shrug*

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -653,6 +653,10 @@ public void handleEnvoyResourceChangedInZone(@Nullable String tenantId,
     else {
       log.debug("No bound monitors were previously assigned to envoy={}", fromEnvoyId);
     }
+
+    // Also check to make sure there are no unassigned bound monitors
+    // This can happen if new monitors are created while all monitors in a zone are down
```
itzg: is this the wording meant:
```suggestion
    // This can happen if new monitors are created while all poller-envoys in a zone are down
```

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -2650,6 +2652,112 @@ public void testHandleZoneResourceChanged_publicZone() {
...
+            .setResourceId("r-5"),
+        new BoundMonitor()
+            .setEnvoyId("e-2")
+            .setResourceId("r-6"),
+        new BoundMonitor()
+            .setEnvoyId("e-2")
+            .setResourceId("r-7")
+    ));
+
+    // one event for preexisting and one for unassigned monitors
```
itzg: Would have been even nicer for only one event to go out per envoy, but the additional refactoring to support that probably isn't worth it.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
GeorgeJahad: ok, thanks

PR #144, opened 2020-02-10
Added endpoint for generating monitor plugins schema

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-788

# What

The UI(s) interacting with Salus to create/update monitors would benefit from having a live, programmatic way of determining the schema for the monitor plugins.

# How

Use the library https://github.com/mbknor/mbknor-jackson-jsonSchema to dynamically generate a JSON schema of the combination of `LocalPlugin` and `RemotePlugin`.

## How to test

Added unit tests

# Why

I had first looked at Springfox since we are already using it for swagger documentation generation; however, it didn't seem to have an API for programmatically accessing a subset of the schema.

Next, I discovered the FasterXML/Jackson module https://github.com/FasterXML/jackson-module-jsonSchema ; however, it didn't support polymorphic types with `@JsonSubTypes`. Findings some issues related to that led me to https://github.com/mbknor/mbknor-jackson-jsonSchema


itzg:
Here is an example of the schema response:

https://gist.github.com/itzg/7eedced2b694d53f7b516d43dabb4cc3

src/main/java/com/rackspace/salus/monitor_management/services/SchemaService.java
```diff
@@ -0,0 +1,48 @@
...
+    RemotePlugin remote;
+  }
+
+  @Autowired
+  public SchemaService(ObjectMapper objectMapper) {
+    schemaGen = new JsonSchemaGenerator(objectMapper);
+  }
+
+  public JsonNode getMonitorPluginsSchema() {
+    return schemaGen.generateJsonSchema(MonitorPluginScopes.class, "monitor-plugins-scopes",
```
harageth: I havent looked at the documentation we generated for a while. Something like this would be cool if we could inject it into that documentation somehow.

itzg: Agreed but then JSON schema is just one part of the overall Swagger/OpenAPI specification, so we might still be stuck with Springfox as the leading solution for that.

src/main/java/com/rackspace/salus/monitor_management/web/model/LocalPlugin.java
```diff
@@ -44,17 +44,17 @@
     @Type(name = "disk", value = Disk.class),
```
harageth: Do we want a reminder comment at the top of this list of @Type's to remind people to reorder when they add new ones?

src/main/java/com/rackspace/salus/monitor_management/web/model/RemotePlugin.java
```diff
@@ -26,20 +26,19 @@
...
+    @Type(name = "dns", value = Dns.class),
     @Type(name = "http", value = HttpResponse.class),
-    @Type(name = "net_response", value = NetResponse.class),
     @Type(name = "mysql", value = MysqlRemote.class),
+    @Type(name = "net_response", value = NetResponse.class),
+    @Type(name = "ping", value = Ping.class),
     @Type(name = "postgresql", value = PostgresqlRemote.class),
     @Type(name = "sqlserver", value = SqlServerRemote.class),
-    @Type(name = "dns", value = Dns.class)
+    @Type(name = "ssl", value = X509Cert.class)
```
harageth: Same for here with commenting about maintaining this as an ordered list

src/main/java/com/rackspace/salus/monitor_management/web/model/LocalPlugin.java
itzg: Good idea

PR #145, opened 2020-02-10
Enable role based json views

jjbuchan:
# What

Enable role based json views instead of endpoint driven views.

# How

1. Add `@EnableRoleBasedJsonViews` to app
   * https://github.com/racker/salus-common/pull/49
1. Remove `@JsonView` annotations on api controllers
1. Add default role mappings to `application-dev.yml`

# Why

See https://github.com/racker/salus-common/pull/49

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/TelemetryMonitorManagementApplication.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/TelemetryMonitorManagementApplication.java
@@ -18,16 +18,17 @@ package com.rackspace.salus.monitor_management;
 
 import com.rackspace.salus.common.messaging.EnableSalusKafkaMessaging;
 import com.rackspace.salus.common.util.DumpConfigProperties;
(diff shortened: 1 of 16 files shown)


src/main/java/com/rackspace/salus/monitor_management/TelemetryMonitorManagementApplication.java
```diff
@@ -18,16 +18,17 @@
...
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
-import org.springframework.context.annotation.Import;
 
 @SpringBootApplication
 @EnableSalusKafkaMessaging
 @EnableEtcd
-@Import(ExtendedErrorAttributesConfig.class)
+@EnableRoleBasedJsonViews
+@EnableExtendedErrorAttributes
```
itzg: Thanks, I must have missed this one.

PR #150, opened 2020-03-02
resourceId case insensitive comparison

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-694

# What

Make sure that the resource id for exclusions is doing a case insensitive check in the code since that is what the database is doing. 

# How

When we get the resourceId's back we toLowerCase them for comparison but don't alter the values being stored at all. 

## How to test

Run unit tests

# Why

The collation in the database meant that we were already doing case insensitive comparisons. Behind the scenes its just a bit shift one way or the other between upper and lower case conversion, and I am not aware of a better/easier way to do case insensitive comparison in java.

# TODO

Should be done

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1211,15 +1211,16 @@ public class MonitorManagement {
                                                   Map<String, String> labelSelector,
                                                   @NotNull LabelSelectorMethod labelSelectorMethod,
                                                   Set<String> excludedResourceIds) {
+    Set<String> finalExcludedResourceIds = excludedResourceIds.stream()
+        .map(String::toLowerCase)
+        .collect(Collectors.toSet());
+
     return resourceApi
         .getResourcesWithLabels(tenantId, labelSelector, labelSelectorMethod)
         .stream()
         // filter to keep resources that are not in the given exclusion set
-        .filter(resourceDTO -> excludedResourceIds == null ||
-            !excludedResourceIds.stream()
-                .map(String::toLowerCase)
-                .collect(Collectors.toSet())
-                .contains(resourceDTO.getResourceId().toLowerCase()))
+        .filter(resourceDTO -> finalExcludedResourceIds == null ||
+                finalExcludedResourceIds.contains(resourceDTO.getResourceId().toLowerCase()))
         .collect(Collectors.toList());
   }
 
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1216,8 +1216,10 @@ public class MonitorManagement {
         .stream()
         // filter to keep resources that are not in the given exclusion set
         .filter(resourceDTO -> excludedResourceIds == null ||
-            !excludedResourceIds.stream().map(String::toLowerCase).collect(Collectors.toList()).contains(resourceDTO.getResourceId().toLowerCase()))
-            //!excludedResourceIds.contains(resourceDTO.getResourceId()))
+            !excludedResourceIds.stream()
+                .map(String::toLowerCase)
+                .collect(Collectors.toSet())
+                .contains(resourceDTO.getResourceId().toLowerCase()))
         .collect(Collectors.toList());
   }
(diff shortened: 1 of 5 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1216,7 +1216,8 @@ private static boolean labelSelectorMethodChanged(Monitor original, MonitorCU ne
         .stream()
         // filter to keep resources that are not in the given exclusion set
         .filter(resourceDTO -> excludedResourceIds == null ||
-            !excludedResourceIds.contains(resourceDTO.getResourceId()))
+            !excludedResourceIds.stream().map(String::toLowerCase).collect(Collectors.toList()).contains(resourceDTO.getResourceId().toLowerCase()))
```
itzg: Can you change this to `toSet` in order to retain the tiny optimization I had done with `excludedResourceIds`?
```suggestion
            !excludedResourceIds.stream().map(String::toLowerCase).collect(Collectors.toSet()).contains(resourceDTO.getResourceId().toLowerCase()))
```

jjbuchan: This feels like it could be inefficient.  Do you know if `.map(String::toLowerCase).collect(Collectors.toList())` will be done once, or for every item returned by `getResourcesWithLabels`?

Might want to build that list before this filter function?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1216,7 +1216,8 @@ private static boolean labelSelectorMethodChanged(Monitor original, MonitorCU ne
...
+            //!excludedResourceIds.contains(resourceDTO.getResourceId()))
```
jjbuchan: cleanup

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1445,7 +1446,9 @@ void handleResourceChangeEvent(ResourceEvent event) {
           .stream()
           // but filter to include only monitors that don't exclude this resource
           .filter(monitor -> monitor.getExcludedResourceIds() == null ||
-              !monitor.getExcludedResourceIds().contains(resourceId))
+              !monitor.getExcludedResourceIds().stream().map(String::toLowerCase).collect(
+                  Collectors.toList()).contains(resourceId.toLowerCase()))
+              //!monitor.getExcludedResourceIds().contains(resourceId))
```
jjbuchan: same comments as above

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1445,7 +1446,9 @@ void handleResourceChangeEvent(ResourceEvent event) {
           .stream()
           // but filter to include only monitors that don't exclude this resource
           .filter(monitor -> monitor.getExcludedResourceIds() == null ||
-              !monitor.getExcludedResourceIds().contains(resourceId))
+              !monitor.getExcludedResourceIds().stream().map(String::toLowerCase).collect(
+                  Collectors.toList()).contains(resourceId.toLowerCase()))
```
itzg: And here
```suggestion
                  Collectors.toSet()).contains(resourceId.toLowerCase()))
```

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
itzg: Can remove this

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
jjbuchan: @itzg may know the answer to this?  I can look into it too in a bit.

itzg: I don't know off hand

jjbuchan:
Just make sure the build / tests pass.  it currently shows failures.

harageth:
> Just make sure the build / tests pass. it currently shows failures.

Fixing that right now. It was causing NPE's when the excludedResourceId's set was null.

PR #151, opened 2020-03-02
Added activation of PolicyApiCacheConfig

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-731

# What

From issue description:

Things like monitor management have to request data from policy management, but that data will change infrequently.

These responses should be cached so if policy mgmt it does not cause many problems and so that general api requests can be faster.

# How

`@Import` the new `PolicyApiCacheConfig` added in https://github.com/racker/salus-policy-management/pull/25 and add the appropriate dependencies for ehcache usage. (Transitive dependencies don't work since our convention is to exclude all transitives from the client dependencies.)

# How to test

Existing unit tests

### DIFF
--- a/src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
+++ b/src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
@@ -28,6 +28,7 @@ import static org.hamcrest.Matchers.nullValue;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyBoolean;
 import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.times;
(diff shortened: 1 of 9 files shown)


jjbuchan:
See https://github.com/racker/salus-policy-management/pull/25#issuecomment-595959821 for a description of my changes.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
```diff
@@ -929,7 +930,7 @@ public void testRefreshBoundPolicyMonitorsForTenant_partiallyExistingBoundMonito
     org.assertj.core.api.Assertions.assertThat(captorOfBoundMonitorList.getValue())
         .containsExactlyInAnyOrderElementsOf(newBound);
 
-    verify(policyApi).getEffectivePolicyMonitorIdsForTenant(tenantId);
+    verify(policyApi).getEffectivePolicyMonitorIdsForTenant(tenantId, false);
```
GeorgeJahad: how do you decide which of these should use the cache and which shouldn't?

jjbuchan: > This change allows us to bypass the cache whenever policy metadata is changed to ensure future requests pull in the new field and do not continue to use the old value.
>
> That has been done through the modifications in racker/salus-telemetry-monitor-management#151. When Monitor Management receives an event stating policy data has changed in any way it will clear the cache while processing that change. For other requests to the policy api, such as a customer creating a new monitor via the api, those will simply query the cache instead of having to make the full request.

So if it is a method related to receiving an event for a policy change it should be `false`, if it is anything else such as creating/updating a monitor it should be `true`.

`refreshBoundPolicyMonitorsForTenant` is called when policies change.

PR #154, opened 2020-03-06
Added JsonSchemaDefault to fields that use explicit default

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-830

# What

The JSON schema generator used at runtime doesn't seem to leverage `@JsonProperty(defaultValue = "true")`, so those cases also need a `@JsonSchemaDefault("true")`.

## How to test

Existing unit tests and manually accessing `/schema/monitor-plugins` endpoint.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Cpu.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Cpu.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Rackspace US, Inc.
+ * Copyright 2020 Rackspace US, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@
 package com.rackspace.salus.monitor_management.web.model.telegraf;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.kjetland.jackson.jsonSchema.annotations.JsonSchemaDefault;
 import com.rackspace.salus.monitor_management.web.model.ApplicableAgentType;
 import com.rackspace.salus.monitor_management.web.model.ApplicableMonitorType;
 import com.rackspace.salus.monitor_management.web.model.LocalPlugin;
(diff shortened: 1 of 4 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Cpu.java
```diff
@@ -30,6 +31,7 @@
 @ApplicableMonitorType(MonitorType.cpu)
 public class Cpu extends LocalPlugin {
   boolean percpu;
+  @JsonSchemaDefault("true")
   @JsonProperty(defaultValue = "true")
   boolean totalcpu = true;
```
harageth: What am I missing for the point of setting the default 3 different ways?

itzg: Please refer to the `What` in the description. I agree it's not ideal.

harageth: totalcpu = true is needed to actually set the default value.
@JsonProperty(defaultValue = "true") is necessary for serializing and deserializing the json. Do we need to set the defaultValue here as well as on the totalcpu field?
@JsonSchemaDefault("true") is necessary for the JSON schema generator api.

itzg: I would clarify the wording of the first one to be:

`totalcpu = true` is needed in cases, such as unit tests, where the object is programmatically instantiated

itzg: ...and I'm pretty sure, yes, the second scenario is needed for JSON deserialization when the property is not present even though the class field has a default.

itzg:
@harageth I'm going to poke around in that library's code a bit because I now see in the "Nullable types" section of its README that it does have some `@JsonProperty` awareness. I won't poke much because Scala hurts my eyes :) .

itzg:
@harageth thanks to the rapid turnaround on https://github.com/mbknor/mbknor-jackson-jsonSchema/issues/117 I was able to revert the `@JsonSchemaDefault` changes since the generated schema default can now be derived from `@JsonProperty(defaultValue = "...")`

PR #158, opened 2020-03-16
updated boolean parser

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-783

# What

Since we are updating how the validation works we should update how we are parsing the values as well.

# How

It switches the parse function to our new `BooleanParser.parseBoolean` function to match how the validation is happening

## How to test

Run the unit tests

# Why

We didn't like the standard `parseBoolean` function and how it assumed anything that wasn't 'true' was just false. 

# TODO

Should be done

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/utils/MetadataUtils.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/utils/MetadataUtils.java
@@ -16,6 +16,7 @@
 
 package com.rackspace.salus.monitor_management.utils;
 
+import com.rackspace.salus.common.util.BooleanParser;
 import com.rackspace.salus.monitor_management.web.model.MonitorCU;
 import com.rackspace.salus.policy.manage.web.client.PolicyApi;
 import com.rackspace.salus.policy.manage.web.model.MonitorMetadataPolicyDTO;
@@ -113,7 +114,7 @@ public class MetadataUtils {
                 break;
               case BOOL:
                 Boolean boolValue = (Boolean) f.get(object);
-                if (boolValue == Boolean.parseBoolean(policy.getValue())) {
+                if (boolValue == BooleanParser.parseBoolean(policy.getValue())) {
                   metadataFields.add(f.getName());
                 }
                 break;
@@ -169,7 +170,7 @@ public class MetadataUtils {
           f.set(object, Duration.parse(policy.getValue()));
           break;
         case BOOL:
-          f.set(object, Boolean.parseBoolean(policy.getValue()));
+          f.set(object, BooleanParser.parseBoolean(policy.getValue()));
           break;
       }
     } catch (IllegalAccessException|NoSuchFieldException e) {
--- a/src/main/java/com/rackspace/salus/monitor_management/utils/MetadataUtils.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/utils/MetadataUtils.java
@@ -114,7 +114,7 @@ public class MetadataUtils {
                 break;
               case BOOL:
                 Boolean boolValue = (Boolean) f.get(object);
-                if (boolValue == BooleanParser.parseBoolean(policy.getValue())) {
+                if (boolValue == BooleanUtils.toBoolean(policy.getValue().toLowerCase(), "true", "false")) {
                   metadataFields.add(f.getName());
                 }
                 break;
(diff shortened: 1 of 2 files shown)


src/test/java/com/rackspace/salus/monitor_management/utils/MetadataUtilsTest.java
```diff
@@ -248,6 +251,31 @@ public void setUpdateMetadataValue_STRING_LIST() {
...
+        .setMonitorType(MonitorType.http)
+        .setKey("followRedirects")
+        .setValueType(MetadataValueType.BOOL)
+        .setValue("true");
+
+    MetadataUtils.updateMetadataValue(monitor, policy);
+    assertThat(monitor.getFollowRedirects()).isEqualTo(true);
+  }
+
+  @Test(expected = BooleanFormatException.class)
```
itzg: I was going to compare the `@Test(expected = ...)` usage here to other test cases, but I guess we didn't have any negative test cases for the other data type parsing.

This approach is fine, but using an `assertThatThrownBy` is what we're trying to use more:
https://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#exception-assertion

The added benefit is asserting the exception's message, etc.

harageth: I remember having this same conversation in the past and tried to look it up but wasn't finding it and found where we were still using this in a few tests.

itzg: Yes, I think we have :) but I'm not sure I was able to find the docs before. As you said, some tests are using the "expected" approach, so it's valid and consistent.

harageth: updated. I prefer using `assertThatThrownBy` too. And it helped make sure that the message is more readable too.

src/main/java/com/rackspace/salus/monitor_management/utils/MetadataUtils.java
```diff
@@ -113,7 +114,7 @@ public MetadataUtils(PolicyApi policyApi) {
                 break;
               case BOOL:
                 Boolean boolValue = (Boolean) f.get(object);
-                if (boolValue == Boolean.parseBoolean(policy.getValue())) {
+                if (boolValue == BooleanParser.parseBoolean(policy.getValue())) {
```
jjbuchan: Similar to my comments on other repos, I think we should use the apache commons class instead of creating our own.

src/test/java/com/rackspace/salus/monitor_management/utils/MetadataUtilsTest.java
```diff
@@ -248,6 +252,32 @@ public void setUpdateMetadataValue_STRING_LIST() {
...
+
+  @Test
+  public void setUpdateMetadataValue_BOOLfails() {
+    HttpResponse monitor = new HttpResponse();
+    MonitorMetadataPolicyDTO policy = (MonitorMetadataPolicyDTO) new MonitorMetadataPolicyDTO()
+        .setKey("followRedirects")
+        .setValueType(MetadataValueType.BOOL)
+        .setValue("not a boolean value");
+
+    assertThatThrownBy(() -> MetadataUtils.updateMetadataValue(monitor, policy))
```
jjbuchan: 👍

src/test/java/com/rackspace/salus/monitor_management/utils/MetadataUtilsTest.java
```diff
@@ -274,8 +274,8 @@ public void setUpdateMetadataValue_BOOLfails() {
         .setValue("not a boolean value");
 
     assertThatThrownBy(() -> MetadataUtils.updateMetadataValue(monitor, policy))
-      .isInstanceOf(BooleanFormatException.class)
-      .hasMessageContaining("'not a boolean value' is not a valid boolean value");
+      .isInstanceOf(IllegalArgumentException.class)
+      .hasMessageContaining("The String did not match either specified value");
```
itzg: That's a bummer the exception thrown by `toBoolean` has such a vague message.

harageth: Yeah... Although on the plus side the validation part of things should obfuscate that

jjbuchan: Yep, the reality is that the api validation should prevent this from ever being hit here.  It would only appear if we added more methods that hits this directly without coming through the api validation.

PR #160, opened 2020-03-20
Add clone monitor functionality

jjbuchan:
# What

Adds the ability to clone monitors across different tenants.

# How

The original monitor's tenant is modified, id is removed, and any metadata fields will be updated to reflect the metadata in place for the new tenant.  The monitor is then stored under the new tenant (which generates a new id) and then bound to that tenant's resources.

An admin api endpoint is introduced to perform this now.

# Why

This may be needed on its own for the case of devices moving between accounts or whole accounts being taken over by a different account...

...but the current reason for adding it is as part of the changes required to improve the usability/flexibility of policy monitors.

Rather than a policy monitor only being bound to a tenant's devices, the entire monitor will be cloned to the account so that customer can modify the monitor if required.

More info on this will be added to https://github.com/racker/salus-docs

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
@@ -402,4 +402,40 @@ public class MonitorConversionService {
       return null;
     }
   }
+
+  public Object getPluginFromMonitor(Monitor monitor) throws InvalidClassException {
+    DetailedMonitorInput input = new DetailedMonitorInput(convertToOutput(monitor));
+    MonitorDetails details = input.getDetails();
+
+    Object plugin;
+    if (details instanceof LocalMonitorDetails) {
+      plugin = ((LocalMonitorDetails) details).getPlugin();
+    } else if (details instanceof RemoteMonitorDetails) {
+      plugin = ((RemoteMonitorDetails) details).getPlugin();
+    } else {
+      throw new InvalidClassException(String.format(
+          "Unexpected class found when extracting monitor plugin content=%s",
+          monitor.getContent()));
+    }
+
+    return plugin;
+  }
+
+  public void refreshClonedPlugin(String tenantId, Monitor monitor) {
+    Object plugin;
+    try {
+      plugin = getPluginFromMonitor(monitor);
+    } catch (InvalidClassException e) {
+      log.error("Failed to refresh cloned plugin for monitor={}", monitor, e);
+      return;
+    }
+
+    metadataUtils.setMetadataFieldsForClonedPlugin(tenantId, monitor, plugin);
+    try {
+      monitor.setContent(objectMapper.writeValueAsString(plugin));
+    } catch (JsonProcessingException e) {
+      log.warn("Failed to serialize plugin details of monitor={}", monitor, e);
+      throw new IllegalStateException("Failed to serialize plugin details");
+    }
+  }
 }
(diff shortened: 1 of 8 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -291,6 +294,48 @@ public Monitor createMonitor(String tenantId, @Valid MonitorCU newMonitor) throw
...
+    monitorConversionService.refreshClonedPlugin(newTenant, clonedMonitor);
+
+    // assign id to monitor and store in db
+    clonedMonitor.setId(null);
+    clonedMonitor = monitorRepository.save(clonedMonitor);
+
+    // bind monitor
+    Set<String> affectedEnvoys = bindNewMonitor(newTenant, clonedMonitor);
+    log.info("Binding policy monitor={} to {} envoys on tenant={}",
+        monitor, affectedEnvoys.size(), newTenant);
```
itzg: Since it's "Binding", this log seems like it should be before `bindNewMonitor`

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -629,6 +630,26 @@ public void testCreateNewMonitorInvalidZone() {
...
+    create.setSelectorScope(ConfigSelectorScope.REMOTE);
+
+    String tenantId = RandomStringUtils.randomAlphanumeric(10);
+
+    when(zoneManagement.getAvailableZonesForTenant(any(), any()))
+        .thenReturn(new PageImpl<>(zones, Pageable.unpaged(), zones.size()));
+
+    Monitor monitor = monitorManagement.createMonitor(tenantId, create);
+
+    assertThat(monitor.getZones(), hasSize(5));
```
itzg: I'm not sure where `5` came from. Does podam always generate lists of size 5?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorConversionService.java
```diff
@@ -402,4 +402,40 @@ private MonitorDetails convertContentToDetails(Monitor monitor) {
...
+  public Object getPluginFromMonitor(Monitor monitor) throws InvalidClassException {
+    DetailedMonitorInput input = new DetailedMonitorInput(convertToOutput(monitor));
+    MonitorDetails details = input.getDetails();
+
+    Object plugin;
+    if (details instanceof LocalMonitorDetails) {
+      plugin = ((LocalMonitorDetails) details).getPlugin();
+    } else if (details instanceof RemoteMonitorDetails) {
+      plugin = ((RemoteMonitorDetails) details).getPlugin();
+    } else {
```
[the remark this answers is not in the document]
jjbuchan: Yeah, I almost switched it to just one `if/else`.  I can't remember if there was any other reason for doing it other than to be more specific with the attempted conversions and the exception that we'd have to handle.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
jjbuchan: Yep, it's a podam thing for anytime you're creating a list.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
jjbuchan: I was about to change that then realize why it's here.  It's reporting the amount of envoys it's binding to.  It doesn't know that until after everything has passed through `bindNewMonitor`.

P.S. This was just copied from one of our other similar methods.

itzg: Oh yeah, needs to be after the call then. It was just the age-old present/past tense on log wording that caught my eye.

PR #162, opened 2020-03-27
Update Policy Monitor Handling

jjbuchan:
# What

Updates how policy monitors are applied to customer tenants.

# How

Rather than a policy monitor only being bound to a tenant's devices, the entire monitor will be cloned to the account so that customer can modify the monitor if required.

Once the monitor has been cloned it acts like any other monitor on the tenant except it cannot be deleted - customer's instead would have to opt out of the policy for it to be removed.

More info on this will be added to https://github.com/racker/salus-docs - for now see the internal docs repo which documents the proposal for this.

## How to test

I've added new tests / tweaked existing ones & manually tested via local dev.

# Why

It improves the customer experience by adding more flexibility.


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1312,7 +1302,20 @@ public void removePolicyMonitor(UUID id) {
...
   void handleMonitorPolicyEvent(MonitorPolicyEvent event) {
     log.info("Handling monitor policy event={}", event);
-    refreshBoundPolicyMonitorsForTenant(event.getTenantId());
+
+    String tenantId = event.getTenantId();
+    UUID policyId = event.getPolicyId();
+    UUID monitorId = event.getMonitorId();
+
+    // get effective policies
+    List<UUID> activePolicies = policyApi.getEffectiveMonitorPolicyIdsForTenant(event.getTenantId(), false);
```
itzg: nit, since you have the variable
```suggestion
    List<UUID> activePolicies = policyApi.getEffectiveMonitorPolicyIdsForTenant(tenantId, false);
```

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1405,51 +1408,33 @@ void handleMetadataPolicyEvent(MetadataPolicyEvent event) {
...
-        continue;
-      }
-      Set<String> newBindings = bindNewMonitor(tenantId, monitor.get());
-      affectedEnvoys.addAll(newBindings);
-      log.info("Binding policy monitor={} to {} envoys on tenant={}",
-          monitor.get(), newBindings.size(), tenantId);
-    }
+    List<UUID> policyIds = policyApi.getEffectiveMonitorPolicyIdsForTenant(tenantId, false);
+    List<UUID> policyIdsInUse = monitorRepository.findByTenantIdAndPolicyIdIsNotNull(tenantId)
+        .stream().map(Monitor::getPolicyId).filter(Objects::nonNull).collect(Collectors.toList());
```
itzg: Just curious why the `.filter(Objects::nonNull)` even though the query is for `PolicyIdIsNotNull`

jjbuchan:
Pushed a couple more commits.  One for review comments and one for something I noticed while starting work on my next issue.

The fact test pass before and after that commit show we're lacking some around policy monitor updates.  I'll look into them in a follow up PR.  They're not a "normal" action that should happen so it doesn't really affect us right now; but we need to ensure in future if they are done that they don't break things.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
jjbuchan: Good point.  That isn't needed.

PR #163, opened 2020-03-27
made boolean values Objects rather than primitives

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-844

# What

It is making sure that the schema api will properly display the optional parameters for Faraji

# How

The library that is being used for introspection on the objects doesn't recognize that primitives are optional values, so instead we made sure to make the values Object's instead.

## How to test

Run the unit tests

# Why

There couldn't have been any solution that required


src/test/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiControllerTest.java
```diff
@@ -874,7 +874,7 @@ public void testCreateMonitor_intervalParsing_valid() throws Exception {
                 .setSelectorScope(ConfigSelectorScope.LOCAL)
                 .setMonitorType(MonitorType.cpu)
                 .setAgentType(AgentType.TELEGRAF)
-                .setPluginMetadataFields(Collections.emptyList())
+                .setPluginMetadataFields(List.of("percpu", "collectCpuTime", "reportActive"))
```
itzg: I can't tell why the metadata fields would have populated differently here. Maybe @jjbuchan needs to take a closer look at this.

src/test/resources/MonitorApiControllerTest/converted_monitor_duration.json
```diff
@@ -1 +1 @@
-{"type":"cpu","percpu":false,"totalcpu":true,"collectCpuTime":false,"reportActive":false}
\ No newline at end of file
+{"type":"cpu","percpu":null,"totalcpu":true,"collectCpuTime":null,"reportActive":null}
```
itzg: This seems to correlate with my question above, but it's a different file.

src/test/resources/MonitorSchemaControllerTest/monitor_plugins_schema_partial.json
```diff
@@ -48,11 +48,7 @@
       },
       "title": "cpu",
       "required": [
-        "type",
-        "percpu",
-        "totalcpu",
-        "collectCpuTime",
-        "reportActive"
+        "type"
```
itzg: This is good to see. I was expecting these now nullable fields to render in the schema as non-required.

src/test/resources/MonitorApiControllerTest/converted_monitor_duration.json
harageth: Why wouldn't the values here change to being null once we altered everything to Objects and didn't set values and didn't set defaults?

src/test/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiControllerTest.java
harageth: Yeah I tried to look up how this was being generated but got lost. And the debugger is showing the above stubMonitorResp pluginMetadataFields as being null. So I was rather confused.

src/test/resources/MonitorApiControllerTest/converted_monitor_duration.json
itzg: I don't understand your question, but my comment here wasn't important. I was just pointing out that the same three field names adjusted here are the same three in the commented code above.

src/test/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiControllerTest.java
jjbuchan: Previous default `false` values were being set for those so they were not utilizing policy metadata.  Now they are set to `null` they will use whatever is found in Policy Management.  `pluginMetadataFields` contains the list of any fields in the plugin that are using metadata.

In these cases policy mgmt is not returning any value to set in it's place, but if a policy was added in future it would know to update these fields on this monitor.

src/main/java/com/rackspace/salus/monitor_management/web/model/telegraf/Cpu.java
```diff
@@ -29,9 +29,9 @@
 @ApplicableAgentType(AgentType.TELEGRAF)
 @ApplicableMonitorType(MonitorType.cpu)
 public class Cpu extends LocalPlugin {
-  boolean percpu;
+  Boolean percpu;
   @JsonProperty(defaultValue = "true")
-  boolean totalcpu = true;
-  boolean collectCpuTime;
-  boolean reportActive;
+  Boolean totalcpu = true;
```
jjbuchan: Do we definitely want this to always be `true` or should it be `null` and moved to be a policy instead?

itzg: Good question. That could now be controlled by policy.

harageth: Do we need to start documenting which policies we definitely need to have setup? What happens if this isn't set and there are no policies covering it? Seems like Telegraf might complain.

itzg: The data loader content repo basically serves that purpose. As for telegraf complaining, it won't either way with that field...I just got tired of setting that one true and the others false for those early demos :)

harageth:
> ...but have you verified that once the fields are set to `null` that it will still be configured as we expect?
> 
> I.e. does a `null` value (and therefore the field being excluded from the monitor config) behave the same as the previous `false` value?

The answer is no. The CPU plugin has different defaults to what we had set. 
https://github.com/influxdata/telegraf/tree/master/plugins/inputs/cpu


> Similarly with changing these to `null` it allows these fields to use policy metadata. Do we need to add any policies for them?

I suppose that depends on whether we care about the previous defaults and the defaults that the plugin imposes. If we are allowing null values my direct thought process would be to match the plugin defaults to our defaults.

PR #165, opened 2020-04-07
Add field to monitor payload to show whether it is tied to a policy

jjbuchan:
Adds a boolean `policy` field to the payload returned from any api query for a monitor.

This will allow the customer to know which monitors were created via Policy Mgmt and which monitors were created by themselves.

I purposely chose not to return the actual policyId it relates to since policies are not customer facing objects.


jjbuchan:
Even after wiping maven's cache i've not been able to reproduce the build check failure locally.  Going to merge this and hope for the best.

PR #171, opened 2020-04-24
Added monitor name and summary to BoundMonitorDTO

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-890

# What

For the resource details page, Intelligence is needing the monitor name to be part of the GET bound-monitors response. Talking more we remembered that since monitor name is optional then the "summary" field is also needed.

# How

Added fields to DTO. Bringing over the monitor name was easy. Incorporating the summary required jumping through a couple small hoops via a new method in the MonitorConversionService.

Bump in version for salus-telemetry-protocol was totally unrelated, but I had forgotten to push that earlier when that module was updated.

## How to test

Updated unit tests

### DIFF
--- a/src/test/java/com/rackspace/salus/monitor_management/web/model/BoundMonitorDTOTest.java
+++ b/src/test/java/com/rackspace/salus/monitor_management/web/model/BoundMonitorDTOTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Rackspace US, Inc.
+ * Copyright 2020 Rackspace US, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +24,7 @@ import static org.junit.Assert.assertThat;
 
 import com.rackspace.salus.telemetry.entities.BoundMonitor;
 import java.lang.reflect.Field;
+import java.util.Map;
 import org.junit.Test;
 import org.springframework.util.ReflectionUtils;
 import uk.co.jemos.podam.api.PodamFactory;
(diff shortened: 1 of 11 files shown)


src/test/java/com/rackspace/salus/monitor_management/web/model/BoundMonitorDTOTest.java
```diff
@@ -37,14 +38,16 @@ public void testFieldsCovered() {
     final BoundMonitor boundMonitor = podamFactory.manufacturePojo(BoundMonitor.class);
 
     final BoundMonitorDTO dto = new BoundMonitorDTO(boundMonitor);
+    // manually fill summary since it gets populated by MonitorConversionService
+    dto.setMonitorSummary(Map.of());
```
harageth, line 42: Why fill the summary if you aren't going to check the value?

src/test/java/com/rackspace/salus/monitor_management/web/model/BoundMonitorDTOTest.java
```diff
@@ -37,14 +38,16 @@ public void testFieldsCovered() {
...
 
     // First verification approach is to check that all fields are populated with something.
     // This approach makes sure that the verification further down doesn't miss a new field.
 
     for (Field field : BoundMonitorDTO.class.getDeclaredFields()) {
       field.setAccessible(true);
       final Object value = ReflectionUtils.getField(field, dto);
-      assertThat(value, notNullValue());
+      assertThat(field.getName(), value, notNullValue());
```
harageth, line 50: Ahhh I see that in here it is making sure that none of the values in the DTO object are null.

PR #172, opened 2020-04-27
Adam9500/salus 836 update resource details

harageth:
# Resolves

SALUS-836

# What

This is updating this code for the new ResourceDTO constructor. 

# How

Adds a utility function for calling etcd for envoy information needed for the ResourceDTO. Since we updated that it made some of those exact same calls superfluous so I simplified that code as well.

## How to test

This is mostly a unit test update PR so just run the unit tests

# Why

Because we updated the ResourceDTO to provide more information to different types of users and had to update the references here.

# TODO

Merge Resource Management PR first

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1109,8 +1109,10 @@ public class MonitorManagement {
       List<BoundMonitor> newBoundMonitors = new ArrayList<>();
       for (String zone : addedZones) {
         try {
+          // passing in null because envoyId is not necessary for bindRemoteMonitor.
+          // A future ticket should look at whether we can change the header for bindRemoteMonitor to accept a Resource instead of a ResourceDTO
           newBoundMonitors.add(
-              bindRemoteMonitor(monitor, new ResourceDTO(resource.get(), getEnvoyIdForResource(resource.get())), zone));
+              bindRemoteMonitor(monitor, new ResourceDTO(resource.get(), null), zone));
         } catch (InvalidTemplateException e) {
           log.warn("Unable to render monitor={} onto resource={}",
               monitor, resource.get(), e);
@@ -1273,7 +1275,8 @@ public class MonitorManagement {
     if (StringUtils.isNotBlank(monitor.getResourceId())) {
       Optional<Resource> r = resourceRepository.findByTenantIdAndResourceId(tenantId, monitor.getResourceId());
       if (r.isPresent()) {
-        selectedResources.add(new ResourceDTO(r.get(), getEnvoyIdForResource(r.get())));
+        // We only need the resourceId's from selectedResources. It's superfluous to call etcd to populate the envoyId here.
+        selectedResources.add(new ResourceDTO(r.get(), null));
       } else {
         // It is possible to create monitors for resources that do not yet exist so this
         // is only a warning, but many of them may signal a problem.
@@ -1600,7 +1603,9 @@ public class MonitorManagement {
 
     if (!selectedMonitors.isEmpty()) {
       affectedEnvoys.addAll(
-          upsertBindingToResource(selectedMonitors, new ResourceDTO(resource.get(), getEnvoyIdForResource(resource.get())), event.getReattachedEnvoyId())
+          upsertBindingToResource(selectedMonitors,
+              new ResourceDTO(resource.get(), event.getReattachedEnvoyId()),
+              event.getReattachedEnvoyId())
       );
     }
 
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -505,14 +505,10 @@ public class MonitorManagement {
         // agent monitors can only bind to resources that have (or had) an envoy
         if (resource.isAssociatedWithEnvoy()) {
 
-          final ResourceInfo resourceInfo = envoyResourceManagement
-              .getOne(tenantId, resource.getResourceId())
-              .join();
-
           try {
             boundMonitors.add(
                 bindAgentMonitor(monitor, resource,
-                    resourceInfo != null ? resourceInfo.getEnvoyId() : null)
+                    resource.getEnvoyId() != null ? resource.getEnvoyId() : null)
             );
           } catch (InvalidTemplateException e) {
             log.warn("Unable to render monitor={} onto resource={}",
@@ -2153,6 +2149,7 @@ public class MonitorManagement {
   }
 
   private String getEnvoyIdForResource(Resource resource) {
-    return envoyResourceManagement.getOne(resource.getTenantId(), resource.getResourceId()).join().getEnvoyId();
+    ResourceInfo info = envoyResourceManagement.getOne(resource.getTenantId(), resource.getResourceId()).join();
+    return info == null ? null : info.getEnvoyId() ;
   }
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -486,7 +486,7 @@ public class MonitorManagement {
     if (!StringUtils.isBlank(resourceId)) {
       Optional<Resource> r = resourceRepository.findByTenantIdAndResourceId(monitor.getTenantId(), resourceId);
       resources = new ArrayList<>();
-      r.ifPresent(resource -> resources.add(new ResourceDTO(resource)));
+      r.ifPresent(resource -> resources.add(new ResourceDTO(resource, getEnvoyIdForResource(resource))));
     } else {
       resources = findResourcesByLabels(
           tenantId, monitor.getLabelSelector(), monitor.getLabelSelectorMethod(),
@@ -1114,7 +1114,7 @@ public class MonitorManagement {
       for (String zone : addedZones) {
         try {
           newBoundMonitors.add(
-              bindRemoteMonitor(monitor, new ResourceDTO(resource.get()), zone));
+              bindRemoteMonitor(monitor, new ResourceDTO(resource.get(), getEnvoyIdForResource(resource.get())), zone));
         } catch (InvalidTemplateException e) {
           log.warn("Unable to render monitor={} onto resource={}",
               monitor, resource.get(), e);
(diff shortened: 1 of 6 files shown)


src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementExcludeResourceIdsTest.java
```diff
@@ -184,7 +184,6 @@ public void testCreate_withResources() {
 
     verify(resourceApi).getResourcesWithLabels("t-1", emptyMap(), LabelSelectorMethod.AND);
 
-    verify(envoyResourceManagement).getOne("t-1", "r-1");
```
itzg, line 187: I thought the refactored `getEnvoyIdForResource` would still call this and still need to be verified.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementExcludeResourceIdsTest.java
```diff
@@ -314,8 +317,6 @@ public void testUpdate_withResources() throws InvalidTemplateException {
         .getResourcesWithLabels("t-1", emptyMap(), LabelSelectorMethod.AND);
 
     verify(envoyResourceManagement).getOne("t-1", "r-exclude-include");
-    verify(envoyResourceManagement).getOne("t-1", "r-include-exclude");
-    verify(envoyResourceManagement).getOne("t-1", "r-include-include");
```
itzg, line 318: Why did these verify's need to be removed?

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementExcludeResourceIdsTest.java
```diff
@@ -416,7 +421,6 @@ public void testUpdate_labelChangeSelectsExcluded()
     verify(resourceApi)
         .getResourcesWithLabels("t-1", Map.of("stage", "after"), LabelSelectorMethod.AND);
 
-    verify(envoyResourceManagement).getOne("t-1", "r-include-include");
```
itzg, line 419: Same question

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -415,9 +415,10 @@ public void testCreateNewMonitor_usingResourceId() {
     create.setResourceId(resourceId);
     final Resource resource = podamFactory.manufacturePojo(Resource.class);
     resource.setResourceId(resourceId);
+    resource.setTenantId(tenantId);
     when(resourceRepository.findByTenantIdAndResourceId(anyString(), any()))
         .thenReturn(Optional.of(resource));
-    when(envoyResourceManagement.getOne(anyString(), anyString()))
+    when(envoyResourceManagement.getOne(tenantId, resource.getResourceId()))
```
itzg, line 421: Perhaps you meant to modify the verify that was on line 574? The `when` here won't perform any assertions and actual execution.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementExcludeResourceIdsTest.java
harageth: This is a good question. It comes back to the refactoring I did here where it no longer calls `envoyResourceManagement.getOne`
https://github.com/racker/salus-telemetry-monitor-management/pull/172/files#diff-9f3caf1ed599810e6ed74b7a5556cd14R511

The only time `envoyResourceManagement.getOne` is called is when the resourceId is set https://github.com/racker/salus-telemetry-monitor-management/pull/172/files#diff-9f3caf1ed599810e6ed74b7a5556cd14R486

Since the label selector query to ResourceManagement is returning ResourceDTO's with envoy ID's in them we don't need to make another etcd call here https://github.com/racker/salus-telemetry-monitor-management/pull/172/files#diff-9f3caf1ed599810e6ed74b7a5556cd14L508-L510

itzg: Ah, yes, this is monitor-management so I see what you mean that the ResourceDTO it retrieved would have envoyId already populated.

What confuses me still is that here

https://github.com/racker/salus-telemetry-monitor-management/pull/172/files#diff-9f3caf1ed599810e6ed74b7a5556cd14R489

`getEnvoyIdForResource` is called which in turn calls `envoyResourceManagement.getOne(...)` here

https://github.com/racker/salus-telemetry-monitor-management/pull/172/files#diff-9f3caf1ed599810e6ed74b7a5556cd14R2152

itzg: ...oh and that's because monitor-management is directory querying the DB a few lines up from there

https://github.com/racker/salus-telemetry-monitor-management/pull/172/files#diff-9f3caf1ed599810e6ed74b7a5556cd14R487

It's gotten awkward now that we have apparently both DTO retrieval between services and direct DB queries.

harageth: Yes that is the only path that should call `envoyResourceManagement.getOne(...)` anymore

So ultimately that path is only hit iff the monitor has a ResourceId set. 

Yeah. I was attempting to simplify things with this change so that etcd would only be queried in a single code path. But it definitely makes the tests more difficult to figure out where a `verify` is going to need to be used.

itzg: OK, so the removal of `verify`s is probably legit and the remaining `verifyNoMoreInteractions` should catch any misses.

harageth: Yes

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementExcludeResourceIdsTest.java
```diff
@@ -314,8 +317,6 @@ public void testUpdate_withResources() throws InvalidTemplateException {
         .getResourcesWithLabels("t-1", emptyMap(), LabelSelectorMethod.AND);
 
     verify(envoyResourceManagement).getOne("t-1", "r-exclude-include");
-    verify(envoyResourceManagement).getOne("t-1", "r-include-exclude");
-    verify(envoyResourceManagement).getOne("t-1", "r-include-include");
     verify(envoyResourceManagement, never()).getOne("t-1", "r-exclude-exclude");
 
     verifyNoMoreInteractions(resourceApi, envoyResourceManagement, monitorEventProducer);
```
itzg, line 322: Ah yep, this will fail if the `verify`s were actually called

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1114,7 +1110,7 @@ private static boolean labelSelectorMethodChanged(Monitor original, MonitorCU ne
       for (String zone : addedZones) {
         try {
           newBoundMonitors.add(
-              bindRemoteMonitor(monitor, new ResourceDTO(resource.get()), zone));
+              bindRemoteMonitor(monitor, new ResourceDTO(resource.get(), getEnvoyIdForResource(resource.get())), zone));
```
jjbuchan, line 1113: I don't think this needs to `getEnvoyIdForResource`

jjbuchan: For remote monitors the envoyId of the customer resource is irrelevant, so by not making that call we can prevent an unnecessary call to etcd.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1277,7 +1273,7 @@ private static boolean labelSelectorMethodChanged(Monitor original, MonitorCU ne
     if (StringUtils.isNotBlank(monitor.getResourceId())) {
       Optional<Resource> r = resourceRepository.findByTenantIdAndResourceId(tenantId, monitor.getResourceId());
       if (r.isPresent()) {
-        selectedResources.add(new ResourceDTO(r.get()));
+        selectedResources.add(new ResourceDTO(r.get(), getEnvoyIdForResource(r.get())));
```
jjbuchan, line 1276: I don't think `getEnvoyIdForResource` is needed here either?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1604,7 +1600,7 @@ void handleResourceChangeEvent(ResourceEvent event) {
 
     if (!selectedMonitors.isEmpty()) {
       affectedEnvoys.addAll(
-          upsertBindingToResource(selectedMonitors, new ResourceDTO(resource.get()), event.getReattachedEnvoyId())
+          upsertBindingToResource(selectedMonitors, new ResourceDTO(resource.get(), getEnvoyIdForResource(resource.get())), event.getReattachedEnvoyId())
```
jjbuchan, line 1603: `upsertBindingToResource` already does this:

```
final ResourceInfo resourceInfo = envoyResourceManagement
        .getOne(resource.getTenantId(), resource.getResourceId())
        .join();
```

so i think you're doing double-work?

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementExcludeResourceIdsTest.java
jjbuchan: > It's gotten awkward now that we have apparently both DTO retrieval between services and direct DB queries.

I think the "plan" was to clean those up eventually?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
harageth: Lets look into changing the bindRemoteMonitor header into a Resource instead of a ResourceDTO (to complete this ticket we can just pass in null and leave a comment on why we are passing in null)

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
harageth: So this one does need the ResourceDTO for the mustache template rendering. So I am going to leave it as is but remove the other etcd calls.

PR #175, opened 2020-04-29
Added swagger-codegen-maven-plugin to try out its static html generator

itzg:
Trying out this maven plugin

https://github.com/swagger-api/swagger-codegen/tree/master/modules/swagger-codegen-maven-plugin

as a companion to our current swagger plugin, which still remains necessary for Spring MVC -> swagger.json. In fact the ordering of the plugins is important because this additional plugin takes the `swagger.json` produced by the first one, both during the `generate` phase.

With the language "html" it produces good old fashioned HTML without any of that "xmp" stuff:

https://gist.github.com/itzg/f0dfcf27ba56f67ebd6cce121a14ea97

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -298,6 +298,24 @@
       </plugin>
 
       <plugin>
+        <groupId>io.swagger</groupId>
+        <artifactId>swagger-codegen-maven-plugin</artifactId>
+        <version>2.4.13</version>
+        <executions>
+          <execution>
+            <id>generate-html</id>
+            <goals><goal>generate</goal></goals>
+            <configuration>
+              <inputSpec>${project.build.directory}/generated/swagger/swagger.json</inputSpec>
+              <language>html</language>
+              <output>${project.build.directory}/generated/swagger/</output>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
+
+      <plugin>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -297,6 +297,24 @@
...
+      <plugin>
+        <groupId>io.swagger</groupId>
+        <artifactId>swagger-codegen-maven-plugin</artifactId>
+        <version>2.4.13</version>
+        <executions>
+          <execution>
+            <id>generate-html</id>
+            <goals><goal>generate</goal></goals>
+            <configuration>
+              <inputSpec>${project.build.directory}/generated/swagger/swagger.json</inputSpec>
```
harageth, line 309: This will actually need to be switched to the `converted.html` file in that same directory. (Actually we can do both and upload both the external and internal api docs)

itzg: True. I was just starting simple by pointing at the "raw" swagger.json. Rather than `converted.html` maybe you meant `convertedOutput.json`.

harageth:
Oh hey @itzg look what I found!

itzg:
> Oh hey @itzg look what I found!

What's funny is that I also noticed it in my pending reviews, went to email you, and then saw your comment :)

PR #176, opened 2020-04-30
Added caching of ZoneApiClient getAvailableZones

itzg:
# Resolves

Part of https://jira.rax.io/browse/SALUS-852

# What

During one of the performance runs with about 2,500 total connections to the ambassador I noticed that ZoneApiClient.getAvailableZones stood out as one of the longer running methods relative to others during an attach:

a8e555dd

Since a flood of Envoys from a tenant especially with reconnect attempts would arrive at a good cache-hit rate this seemed like a viable cacheable method to introduce.

# How

Added similar cache config and properties setup as what we have in a couple places now, specifically PolicyApiCacheConfig. 

## How to test

Added unit test for getAvailableZones and included a caching aspect.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/config/ZoneApiCacheConfig.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/config/ZoneApiCacheConfig.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2020 Rackspace US, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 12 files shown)


jjbuchan:
Is this only going to be used for envoy connections?

The bad thing I see is that if they create a new private zone they might have to wait a minute before it'll work.  I'm ok with that since it's not a common scenario, but just wanna confirm the usage.

itzg:
> Is this only going to be used for envoy connections?
> 
> The bad thing I see is that if they create a new private zone they might have to wait a minute before it'll work. I'm ok with that since it's not a common scenario, but just wanna confirm the usage.

You're correct on both parts. Yeah, my compromise for now was to set the default TTL to only 1 minute thinking we could later enhance this to listen for zone change events and invalidate the cache accordingly. I can write up a Jira to capture that follow on work since I wanted to keep this immediate PR small and simple.

PR #177, opened 2020-04-30
Remove direct binding of policy monitors

jjbuchan:
# What

1. Fixed a log message to use the cloned monitor and not the original policy monitor
1. Updated call to `getDefaultZonesForResource` to use cache
    * I can only guess that the original implementation was a copy/paste fail.
1. Removed call to `getPolicyMonitorsForResource`

## How to test

Tested locally and added mockito verifications for policyApi.  These verify tests fail without the removal of `getPolicyMonitorsForResource`.

# Why

This was missed when the cloning of policy monitors was introduced.  There were no failing tests so it slipped by.

Since we now clone policy monitors to the tenant, they act like anything else when binding is performed; we do not need to do a separate operation for policy monitors when a new resource is seen.


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1579,10 +1579,8 @@ void handleResourceChangeEvent(ResourceEvent event) {
               !monitor.getExcludedResourceIds().stream().map(String::toLowerCase).collect(
                   Collectors.toSet()).contains(lowerResourceId))
           .collect(Collectors.toList());
-      //grab monitors that are using resourceId instead of labels
+      // append monitors that are using resourceId instead of labels
       selectedMonitors.addAll(monitorsWithResourceId);
-      // Append all relevant policy monitors
-      selectedMonitors.addAll(getPolicyMonitorsForResource(resource.get()));
```
GeorgeJahad, line 1585: i think this is the line that was causing the problem i was seeing. is that right?

jjbuchan: Yup

PR #178, opened 2020-05-01
Bind monitor to resource even if no envoy associated

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-887

# What

We want to make sure that local monitors still get bound to the resources even if there is no envoy yet.

# How

It removes the if statement checking to see if the resources is associated with an envoy. 

## How to test

The unit tests were updated.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -501,23 +501,17 @@ public class MonitorManagement {
       // AGENT MONITOR
 
       for (ResourceDTO resource : resources) {
-
-        // agent monitors can only bind to resources that have (or had) an envoy
-        if (resource.isAssociatedWithEnvoy()) {
-
-          try {
-            boundMonitors.add(
-                bindAgentMonitor(monitor, resource,
-                    resource.getEnvoyId() != null ? resource.getEnvoyId() : null)
-            );
-          } catch (InvalidTemplateException e) {
-            log.warn("Unable to render monitor={} onto resource={}",
-                monitor, resource, e);
-            invalidTemplateErrors.increment();
-          }
+        try {
+          boundMonitors.add(
+              bindAgentMonitor(monitor, resource,
+                  resource.getEnvoyId() != null ? resource.getEnvoyId() : null)
+          );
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -501,23 +501,17 @@ private void validateMonitoringZones(String tenantId, Monitor monitor, MonitorCU
       // AGENT MONITOR
 
       for (ResourceDTO resource : resources) {
-
-        // agent monitors can only bind to resources that have (or had) an envoy
-        if (resource.isAssociatedWithEnvoy()) {
```
itzg, line 506: Oh...so it sometimes/sorta worked like we thought it did, but only after an envoy had at least once attached for that resource.

jjbuchan: > like we thought it did

Who's this "we" ;)

harageth: Its the Royal "We"

harageth:
Also I mentioned this in standup but I am going to wait till Monday so that @GeorgeJahad can see this so he isn't surprised if it breaks any of his end to end testing.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -2956,6 +2956,8 @@ public void testBindMonitor_AgentWithNoPriorEnvoy() {
     assertThat(result, hasSize(0));
 
     verify(resourceApi).getResourcesWithLabels("t-1", monitor.getLabelSelector(), monitor.getLabelSelectorMethod());
+    BoundMonitor boundMonitor = monitorManagement.bindAgentMonitor(monitor, resource, null);
```
jjbuchan, line 2959: Not sure why you're doing this?  `monitorManagement.bindMonitor` should already call `bindAgentMonitor`?

harageth: For the next line. `monitorManagement.bindMonitor` returns the list of envoyId's. I needed the `BoundMonitor` to verify the saveAll call.

PR #179, opened 2020-05-04
Fix bugs when handling policy events

jjbuchan:
# Resolves

https://jira.rax.io/browse/SALUS-894

# What

Enhances the logic that handles `MonitorPolicyEvent`s.  Specifically ensures that a tenant does not end up with more or less cloned monitors than it should.

# How

The documentation for `handlePolicyRemovalEvent` and `handlePolicyAdditionEvent` should help clarify things.

Basically, anytime a new policy overrides another but uses the same `monitorId` just update the `policyId` stored on the clone with the correct value.  No unbinding/binding should need to occur.

## How to test

Ran new tests with coverage enabled to ensure the right things were being hit.

# Why

In dev I created a `GLOBAL` monitor policy then an `ACCOUNT_TYPE` policy where both used the same `monitorId`.  When I removed the latter policy it's corresponding cloned monitor was removed from my test tenant.  Instead I would expect that monitor to remain because the original `GLOBAL` policy was still in place and essentially took over from the removed one.

I wouldn't expect this to be something we'd hit in prod as I don't envision multiple policies using the same `monitorId` but this helps handle those cases if we do.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1442,6 +1442,7 @@ public class MonitorManagement {
         monitorRepository.findByTenantIdAndPolicyIdIsNotNull(tenantId), Monitor::getPolicyId);
 
     // if the new policy overrides an existing one using the same monitorId, find the existing one
+    // by getting all monitors on the tenant that were cloned from the same monitorId as is in the new event.
     List<Monitor> existingPolicyMonitor = existingPolicyIds.stream()
         .map(id -> monitorPolicyRepository.findById(id).orElseGet(() -> {
           log.warn("Monitor is tied to non-existent policy={}", id);
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -16,6 +16,7 @@
 
 package com.rackspace.salus.monitor_management.services;
 
+import static com.google.common.collect.Collections2.transform;
 import static com.rackspace.salus.telemetry.entities.Monitor.POLICY_TENANT;
 import static com.rackspace.salus.telemetry.entities.Resource.REGION_METADATA;
 import static com.rackspace.salus.telemetry.etcd.types.ResolvedZone.createPrivateZone;
@@ -35,6 +36,8 @@ import com.rackspace.salus.monitor_management.web.model.ZoneAssignmentCount;
 import com.rackspace.salus.monitor_management.web.validator.ValidUpdateMonitor;
 import com.rackspace.salus.policy.manage.web.client.PolicyApi;
 import com.rackspace.salus.policy.manage.web.model.MonitorMetadataPolicyDTO;
+import com.rackspace.salus.policy.manage.web.model.MonitorPolicyDTO;
+import com.rackspace.salus.policy.manage.web.model.PolicyDTO;
 import com.rackspace.salus.resource_management.web.client.ResourceApi;
 import com.rackspace.salus.resource_management.web.model.ResourceDTO;
 import com.rackspace.salus.telemetry.entities.BoundMonitor;
@@ -142,6 +145,8 @@ public class MonitorManagement {
   private final Counter monitorMetadataContentUpdateErrors;
   private final Counter invalidTemplateErrors;
   private final Counter orphanedBoundMonitorRemoved;
+  private final Counter orphanedPolicyMonitors;
+  private final Counter policyIntegrityErrors;
 
   @Autowired
   public MonitorManagement(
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1387,19 +1396,123 @@ public void removePolicyMonitor(UUID id) {
...
       clonePolicyMonitor(tenantId, policyId, monitorId);
-    } else if (!activePolicies.contains(event.getPolicyId()) && monitor.isPresent()){
-      unbindAndRemoveMonitor(monitor.get());
+      return;
+    }
+
+    if (existingPolicyMonitor.size() > 1) {
+      log.error("More than one cloned monitor exists for policy={} on tenant={}", event.getPolicyId(), tenantId);
+      policyIntegrityErrors.increment();
+    }
```
itzg, line 1465: Should it also remove the others in the list or are you thinking we need to manually investigate and resolve these cases anyway?

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
```diff
@@ -744,22 +753,139 @@ public void testHandleMonitorPolicyEvent_newPolicyForTenant() {
...
+   * In this case the removed policy was previously overriding another that used the same monitorId.
+   * The removal causes the other policy to become "effective" and the existing cloned monitor
+   * will be updated to reference that policy instead of the removed one.
+   *
+   * cloneMonitor should not be triggered.
+   */
+  @Test
+  public void testHandleMonitorPolicyEvent_removePolicyInUse_replacedByOther() {
+    String tenantId = RandomStringUtils.randomAlphanumeric(10);
+    UUID newPolicyId = UUID.randomUUID();
```
itzg, line 854: nit: "new" threw me off in the variable name...maybe "other" or "preExisting"?

jjbuchan: I almost went with `other`.  I'll take another look.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
jjbuchan: I'm thinking manual so we can see where things are going wrong.  It also means we can make sure the right one is in use - we're defaulting to the first in the list which may not be accurate.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1387,19 +1396,123 @@ public void removePolicyMonitor(UUID id) {
...
-    // get effective policies
-    List<UUID> activePolicies = policyApi.getEffectiveMonitorPolicyIdsForTenant(tenantId, false);
-    Optional<Monitor> monitor = monitorRepository.findByTenantIdAndPolicyId(tenantId, policyId);
+    List<UUID> effectivePolicies = policyApi.getEffectiveMonitorPolicyIdsForTenant(tenantId, false);
+    if (!effectivePolicies.contains(policyId)) {
+      log.debug("Policy={} is not relevant to tenant={}, no action necessary", policyId, tenantId);
+      return;
+    }
+
+    Collection<UUID> existingPolicyIds = transform(
```
GeorgeJahad, line 1441: out of curiosity, isn't transform() just a map()? any special reason for using it here?

jjbuchan: Yeah, it is.  I found it a little while ago when searching for a java equivalent of `_.pluck()`.

 I found writing out something like:

```
boundMonitors.stream()
          .map(BoundMonitor::getResourceId)
          .collect(Collectors.toList());
```
all the time to be more tedious than
```
transform(boundMonitors, BoundMonitor::getResourceId);
```

No other reason.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1387,19 +1396,123 @@ public void removePolicyMonitor(UUID id) {
...
+        monitorRepository.findByTenantIdAndPolicyIdIsNotNull(tenantId), Monitor::getPolicyId);
+
+    // if the new policy overrides an existing one using the same monitorId, find the existing one
+    List<Monitor> existingPolicyMonitor = existingPolicyIds.stream()
```
GeorgeJahad, line 1445: When I have a long set of transforms like this I like to add a comment that describes it, something like this:
// get the monitors for all tenants policies whose monitorid matches the event

jjbuchan: Added a comment.  I'm not sure if it helps or hurts.

All the things around monitors vs. policy monitors vs. cloned monitors makes documenting those things concisely kinda difficult.

PR #180, opened 2020-05-07
Add endpoint to view all monitor translations

jjbuchan:
# Related PR
https://github.com/racker/salus-telemetry-model/pull/133

# What

Adds a new API endpoint to list all the translations in place for each monitor type.

# How

Don't hate me.  It's all using streams.  I group the list of translators by agentType then monitorType and consolidate all the `info()` details into a list for that type.

The end result is a list of `MonitorTranslationDetails`.

## How to test

Added a test and manual.  This is what gets shown in dev:

```
[
  {
    "agentType": "TELEGRAF",
    "monitorType": "apache",
    "translations": [
      "'url' becomes the singleton array named 'urls'",
      "'timeout' becomes a Golang formatted duration",
      "The key 'timeout' is renamed to 'responseTimeout'"
    ]
  },
  {
    "agentType": "TELEGRAF",
    "monitorType": "disk",
    "translations": [
      "'mount' becomes the singleton array named 'mountPoints'"
    ]
  },
  {
    "agentType": "TELEGRAF",
    "monitorType": "diskio",
    "translations": [
      "'device' becomes the singleton array named 'devices'"
    ]
  },
  {
    "agentType": "TELEGRAF",
    "monitorType": "dns",
    "translations": [
      "'timeout' becomes a number of seconds",
      "The monitor's type is renamed to 'dns_query'",
      "'dnsServer' becomes the singleton array named 'servers'",
      "'domain' becomes the singleton array named 'domains'"
    ]
  },
  {
    "agentType": "TELEGRAF",
    "monitorType": "http",
    "translations": [
      "The monitor's type is renamed to 'http_response'",
      "'url' becomes the singleton array named 'urls'",
      "'timeout' becomes a Golang formatted duration",
      "The key 'timeout' is renamed to 'responseTimeout'"
    ]
  },
  {
    "agentType": "TELEGRAF",
    "monitorType": "mysql",
    "translations": [
      "'intervalSlow' becomes a Golang formatted duration",
      "The value of 'metric_version' is set to 2"
    ]
  },
  {
    "agentType": "TELEGRAF",
    "monitorType": "net",
    "translations": [
      "'interface' becomes the singleton array named 'interfaces'"
    ]
  },
  {
    "agentType": "TELEGRAF",
    "monitorType": "net_response",
    "translations": [
      "The value of 'host' and 'port' are combined into 'host:port' and named 'address'",
      "'readTimeout' becomes a Golang formatted duration",
      "'timeout' becomes a Golang formatted duration"
    ]
  },
  {
    "agentType": "TELEGRAF",
    "monitorType": "ping",
    "translations": [
      "'timeout' becomes a number of seconds",
      "'pingInterval' becomes a number of seconds",
      "'target' becomes the singleton array named 'urls'",
      "'deadline' becomes a number of seconds"
    ]
  },
  {
    "agentType": "TELEGRAF",
    "monitorType": "postgresql",
    "translations": [
      "'maxLifetime' becomes a Golang formatted duration"
    ]
  },
  {
    "agentType": "TELEGRAF",
    "monitorType": "redis",
    "translations": [
      "'url' becomes the singleton array named 'servers'"
    ]
  },
  {
    "agentType": "TELEGRAF",
    "monitorType": "smtp",
    "translations": [
      "The value of 'host' and 'port' are combined into 'host:port' and named 'address'"
    ]
  },
  {
    "agentType": "TELEGRAF",
    "monitorType": "sqlserver",
    "translations": [
      "The value of 'query_version' is set to 2",
      "The key 'queryExclusions' is renamed to 'exclude_queries'"
    ]
  },
  {
    "agentType": "TELEGRAF",
    "monitorType": "ssl",
    "translations": [
      "The monitor's type is renamed to 'x509_cert'",
      "'timeout' becomes a Golang formatted duration",
      "'target' becomes the singleton array named 'sources'"
    ]
  }
]
```

# Why

To help with the documentation of Monitors.  We link to telegraf plugins but due to the translations we do they do not map 1:1.  This endpoint can be referenced in the documentation as the place to see how things differ.

# TODO

1. Add endpoint to public api service.
1. Modifications may be needed but I wasn't sure if it was worth digging into that yet.
    * Agent Version and Selector Scope fields may be needed as we grow the list of translators
    * I wasn't sure the best way to display all that.
       * Maybe the endpoint would require the agent version to be provided and only the relevant  translations would be shown.
    * Currently we don't set selectorScope for any translation and if it stays that way it helps keep things simpler.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
@@ -40,6 +40,7 @@ import java.util.Objects;
 import java.util.UUID;
 import java.util.stream.Collectors;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.collections.ListUtils;
 import org.apache.maven.artifact.versioning.ArtifactVersion;
 import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.apache.maven.artifact.versioning.InvalidVersionSpecificationException;
@@ -119,27 +120,12 @@ public class MonitorContentTranslationService {
     return operators.stream()
         // first ensure everything is in order of priority
         .sorted(Comparator.comparingInt(MonitorTranslationOperator::getOrder))
-        .collect(
-            Collectors.collectingAndThen(
-                // create a group of operators for each agent type
-                Collectors.groupingBy(MonitorTranslationOperator::getAgentType,
-                    Collectors.collectingAndThen(
-                        // within that, create a group of operators for each monitor type
-                        Collectors.groupingBy(MonitorTranslationOperator::getMonitorType,
-                            // build a list of all the  translator info strings for that monitor type
-                            Collectors.mapping(op -> op.getTranslatorSpec().info(), Collectors.toList())),
-
-                        // then generate a MonitorTranslationDetails for each monitor type
-                        monitorMap -> monitorMap.entrySet().stream().map(
-                            monitorEntry -> new MonitorTranslationDetails()
-                                .setMonitorType(monitorEntry.getKey())
-                                .setTranslations(monitorEntry.getValue())))),
-
-                // then assign the agent type to each MonitorTranslationDetails
-                agentMap -> agentMap.entrySet().stream()
-                    .flatMap(agentEntry -> agentEntry.getValue()
-                        .map(detail -> detail.setAgentType(agentEntry.getKey())))))
-
+        // create a hashmap of (agenttype/monitortype:translations)
+        .collect(Collectors.toMap(op -> new MonitorTranslationDetails().setAgentType(op.getAgentType()).setMonitorType(op.getMonitorType()),
+            op -> List.of(op.getTranslatorSpec().info()),
+            ListUtils::union))
+        // convert the map into a stream of details
+        .entrySet().stream().map(entry -> entry.getKey().setTranslations(entry.getValue()))
         // then sort it by agent type and then monitor type
         .sorted(Comparator.comparing(MonitorTranslationDetails::getAgentType, Comparator.comparing(Enum::toString))
             .thenComparing(v -> v.getMonitorType().toString()))
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
@@ -130,10 +130,10 @@ public class MonitorContentTranslationService {
                             Collectors.mapping(op -> op.getTranslatorSpec().info(), Collectors.toList())),
 
                         // then generate a MonitorTranslationDetails for each monitor type
-                        typeMap -> typeMap.entrySet().stream().map(
-                            typeEntry -> new MonitorTranslationDetails()
-                                .setMonitorType(typeEntry.getKey())
-                                .setTranslations(typeEntry.getValue())))),
+                        monitorMap -> monitorMap.entrySet().stream().map(
+                            monitorEntry -> new MonitorTranslationDetails()
+                                .setMonitorType(monitorEntry.getKey())
+                                .setTranslations(monitorEntry.getValue())))),
 
                 // then assign the agent type to each MonitorTranslationDetails
                 agentMap -> agentMap.entrySet().stream()
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
@@ -141,8 +141,8 @@ public class MonitorContentTranslationService {
                         .map(detail -> detail.setAgentType(agentEntry.getKey())))))
 
         // then sort it by agent type and then monitor type
-        .sorted(Comparator.comparing(MonitorTranslationDetails::getAgentType).reversed()
-            .thenComparing(MonitorTranslationDetails::getMonitorType).reversed())
+        .sorted(Comparator.comparing(MonitorTranslationDetails::getAgentType, Comparator.comparing(Enum::toString))
+            .thenComparing(v -> v.getMonitorType().toString()))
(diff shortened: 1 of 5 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
```diff
@@ -112,6 +113,40 @@ public void delete(UUID operatorId) {
...
+                                .setTranslations(typeEntry.getValue())))),
+
+                // then assign the agent type to each MonitorTranslationDetails
+                agentMap -> agentMap.entrySet().stream()
+                    .flatMap(agentEntry -> agentEntry.getValue()
+                        .map(detail -> detail.setAgentType(agentEntry.getKey())))))
+
+        // then sort it by agent type and then monitor type
+        .sorted(Comparator.comparing(MonitorTranslationDetails::getAgentType).reversed()
+            .thenComparing(MonitorTranslationDetails::getMonitorType).reversed())
```
jjbuchan, line 145: Just noticed this sorting isn't working as I expected.

jjbuchan: Realized my mistake.  This was comparing the order of the enum values vs. comparing the string representation of them.  Explains why I unexpectedly had the reverse the order to get tests to pass.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
```diff
@@ -112,6 +113,40 @@ public void delete(UUID operatorId) {
     monitorTranslationOperatorRepository.deleteById(operatorId);
   }
 
+  public List<MonitorTranslationDetails> getMonitorTranslationDetails() {
+    List<MonitorTranslationOperator> operators = monitorTranslationOperatorRepository.findAll();
+
+    return operators.stream()
```
itzg, line 119: I think you're ready for the big leagues with Spring Webflux/Reactive :)

src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
```diff
@@ -112,6 +113,40 @@ public void delete(UUID operatorId) {
...
+
+                        // then generate a MonitorTranslationDetails for each monitor type
+                        monitorMap -> monitorMap.entrySet().stream().map(
+                            monitorEntry -> new MonitorTranslationDetails()
+                                .setMonitorType(monitorEntry.getKey())
+                                .setTranslations(monitorEntry.getValue())))),
+
+                // then assign the agent type to each MonitorTranslationDetails
+                agentMap -> agentMap.entrySet().stream()
+                    .flatMap(agentEntry -> agentEntry.getValue()
```
itzg, line 140: For some reason flatMap is always the one that melts my brain. I'm also realizing the huge benefit of the IntelliJ auto comments if shows for stream code.

jjbuchan: > realizing the huge benefit of the IntelliJ auto comments

You mean this?
image
or are you seeing something else?

itzg: Yep

jjbuchan: > flatMap is always the one that melts my brain

I had to look up the difference between `map` vs. `flatMap` quite a few times.  It's easiest for me to understand when you have a list of lists because then it just turns it all into a single list, but the main thing that helped me is that `map` is always a 1:1 operations where as `flatMap` could return 0 results, 1:1 results, or 1:many results.

In this case each `agentMap` entry will return many  `MonitorTranslationDetails` objects.

GeorgeJahad: the way i remember it is if map() returns a list of lists, [["a"], ["b"]]
then flatMap() returns a list: ["a", "b"].  so basically i ask myself do i want a list or a list of lists.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
GeorgeJahad: I looked at this code for half an hour feeling certain I could simplify it.

I couldn't.

nice work!

GeorgeJahad: So this is my attempt at making it simpler.  It's a bit hacky, but does pass your tests.  What do you think?
https://github.com/racker/salus-telemetry-monitor-management/compare/jjbuchan/add-endpoint-for-translations...george/beingAnnoying

itzg: Very interesting. So at this line

https://github.com/racker/salus-telemetry-monitor-management/compare/jjbuchan/add-endpoint-for-translations...george/beingAnnoying#diff-7f42ca1164a8e03a683f4388aaf26b4bR124

are you leveraging `MonitorTranslationOperator` to be a hash key type?

GeorgeJahad: It's this line:
https://github.com/racker/salus-telemetry-monitor-management/compare/jjbuchan/add-endpoint-for-translations...george/beingAnnoying#diff-7f42ca1164a8e03a683f4388aaf26b4bR124

and yes, that's the hacky part. for clarity, i probably should create a key class that just contains agenttype/monitortype, but i wanted to keep the number of lines to a minimum ;)

Basically there are two simplifications here:
1. instead of creating a hash with key agentType whose value is a hash of key monitorType whose value is some info strings, this creates a hash with key agentType/monitorType whose value is some info strings.
2. The collector.toMap() function seems clearer than the nested levels of collectingAndThen()/collect() methods.

PR #184, opened 2020-07-02
monitor search endpoint

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-902

# What

Make sure that we have functions to search for monitors

# How

Adds an api endpoint to search monitors based on partial id or name

## How to test

Unit tests have been added to the rproject

# Why

The repository functions were already added so this support was straightforward.

# TODO

Open a new JIRA covering database indexing

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -2249,7 +2249,7 @@ public class MonitorManagement {
     return info == null ? null : info.getEnvoyId();
   }
 
-  public Page<Monitor> monitorSearch(String tenantId, String searchCriteria, Pageable page) {
+  public Page<Monitor> getMonitorsBySearchString(String tenantId, String searchCriteria, Pageable page) {
     return monitorRepository.search(tenantId, searchCriteria, page);
   }
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -12,6 +12,7 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
  */
 
 package com.rackspace.salus.monitor_management.services;
@@ -2245,6 +2246,10 @@ public class MonitorManagement {
 
   private String getEnvoyIdForResource(Resource resource) {
     ResourceInfo info = envoyResourceManagement.getOne(resource.getTenantId(), resource.getResourceId()).join();
-    return info == null ? null : info.getEnvoyId() ;
+    return info == null ? null : info.getEnvoyId();
+  }
+
+  public Page<Monitor> monitorSearch(String tenantId, String searchCriteria, Pageable page) {
+    return monitorRepository.search(tenantId, searchCriteria, page);
   }
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
@@ -45,6 +45,7 @@ import javax.json.JsonPatch;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Pageable;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
@@ -333,8 +334,13 @@ public class MonitorApiController {
   }
 
   @GetMapping("/tenant/{tenantId}/search")
-  @ApiOperation("Finds all monitors that match the searchCriteria either in the monitorName or the ID")
-  public PagedContent<Monitor> getMonitorsBySearchString(@PathVariable String tenantId, @RequestParam("q") String searchCriteria, Pageable pageable) {
-    return PagedContent.fromPage(monitorManagement.getMonitorsBySearchString(tenantId, searchCriteria, pageable));
+  @ApiOperation("Finds all monitors that match the searchCriteria either in the monitorName or the ID. Dynamic sorting is not supported and will be ignored.")
+  public PagedContent<Monitor> getMonitorsBySearchString(@PathVariable String tenantId,
+      @RequestParam("q") String searchCriteria,
+      Pageable pageable) {
+    // Because the search is happening in a native query sorting is not supported and causes an exception if the Pageable has a sorting parameter
+    // So this is ignoring the sorting provided
+    Pageable page = PageRequest.of(pageable.getPageNumber(), pageable.getPageSize());
+    return PagedContent.fromPage(monitorManagement.getMonitorsBySearchString(tenantId, searchCriteria, page));
   }
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
@@ -12,6 +12,7 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
  */
 
 package com.rackspace.salus.monitor_management.web.controller;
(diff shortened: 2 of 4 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
```diff
@@ -330,4 +331,10 @@ public void delete(@PathVariable String tenantId,
   public MultiValueMap<String, String> getMonitorLabelSelectors(@PathVariable String tenantId) {
     return monitorManagement.getTenantMonitorLabelSelectors(tenantId);
   }
+
+  @GetMapping("/tenant/{tenantId}/search/{searchCriteria}")
+  @ApiOperation("Finds all monitors that match the searchCriteria either in the monitorName or the ID")
+  public PagedContent<Monitor> getMonitorLabelSelectors(@PathVariable String tenantId, @PathVariable String searchCriteria, Pageable pageable) {
```
itzg, line 337: Same thing as https://github.com/racker/salus-telemetry-resource-management/pull/82#discussion_r451057125, so I'd suggest:
```suggestion
  @GetMapping("/tenant/{tenantId}/search")
  @ApiOperation("Finds all monitors that match the searchCriteria either in the monitorName or the ID")
  public PagedContent<Monitor> getMonitorLabelSelectors(@PathVariable String tenantId, @QueryParam("q") String searchCriteria, Pageable pageable) {
```

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2245,6 +2246,10 @@ private String getRenderedContent(String template, ResourceDTO resourceDTO)
 
   private String getEnvoyIdForResource(Resource resource) {
     ResourceInfo info = envoyResourceManagement.getOne(resource.getTenantId(), resource.getResourceId()).join();
-    return info == null ? null : info.getEnvoyId() ;
+    return info == null ? null : info.getEnvoyId();
+  }
+
+  public Page<Monitor> monitorSearch(String tenantId, String searchCriteria, Pageable page) {
```
jjbuchan, line 2252:
```suggestion
  public Page<Monitor> getMonitorBySearchString(String tenantId, String searchCriteria, Pageable page) {
```
maybe?

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
jjbuchan, line 337: `getMonitorLabelSelectors` doesn't sound right to me?  This doesn't have anything to do with label selectors, does it?  Leftover from a copy/paste?

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4238,4 +4239,88 @@ public void testGetTenantMonitorLabelSelectors() {
...
+        .setLabelSelector(Collections.singletonMap("os", "LINUX"))
+        .setLabelSelectorMethod(LabelSelectorMethod.AND)
+        .setContent("content1")
+        .setAgentType(AgentType.FILEBEAT)
+        .setInterval(Duration.ofSeconds(60)));
+    Pageable page = PageRequest.of(0, 1);
+    Page<Monitor> value = monitorManagement.monitorSearch("t-1", "mon", page);
+    assertThat(value.getTotalElements(), equalTo(2L));
+    assertThat(value.getTotalPages(), equalTo(2));
+    assertThat(value.getNumberOfElements(), equalTo(1));
```
jjbuchan, line 4279: Can you add a comment to state this is the number of entries in the "page".  For a min I was confused how TotalElements and NumberOfElements was different.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4238,4 +4239,88 @@ public void testGetTenantMonitorLabelSelectors() {
...
+        .setMonitorName("otherMonitor")
+        .setMonitorType(MonitorType.cpu)
+        .setLabelSelector(Collections.singletonMap("os", "LINUX"))
+        .setLabelSelectorMethod(LabelSelectorMethod.AND)
+        .setContent("content1")
+        .setAgentType(AgentType.FILEBEAT)
+        .setInterval(Duration.ofSeconds(60)));
+    Pageable page = PageRequest.of(0, 1);
+    Page<Monitor> returnedMonitors = monitorRepository.findByTenantId("t-1", Pageable.unpaged());
+    assertThat(returnedMonitors.getTotalElements(), greaterThan(1L));
```
jjbuchan, line 4316: Why use `greaterThan` instead of the exact number?

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4238,4 +4239,88 @@ public void testGetTenantMonitorLabelSelectors() {
...
+    UUID searchId = returnedMonitors.get().findFirst().get().getId();
+    String searchIdSubString = searchId.toString().substring(10, 16);
```
jjbuchan, line 4318: I'm not sure why you're doing all these steps vs. just using the id returned from the `.save` call?

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
harageth: Yeah I can definitely do that. Since the paging required its own separate query I wanted to make sure it was properly getting tested.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
harageth: Looking at my test I just wanted to make sure that there was at least one that existed. Which makes this even more confusing. I will add a comment here too.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
harageth: I forgot I could

src/test/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiControllerTest.java
```diff
@@ -1015,6 +1015,36 @@ public void testCloneMonitor() throws Exception {
...
+    String searchCriteria = "ping";
+    mockMvc.perform(get(url, tenantId)
+        .param("q", searchCriteria)
+        .contentType(MediaType.APPLICATION_JSON)
+        .characterEncoding(StandardCharsets.UTF_8.name()))
+        .andExpect(status().isOk())
+        .andExpect(content()
+            .contentTypeCompatibleWith(MediaType.APPLICATION_JSON));
+
+    verify(monitorManagement).getMonitorsBySearchString(any(), any(), any());
```
itzg, line 1043: The any's are typical with `when` but for `verify` please assert the specific, expected values.

harageth: I had some problems with using an any for the pages Callie with specific values. But I see that the default paged object is created in a few other tests

itzg: Ah, I see. It's ok to use `any()` on a troublesome parameter or two, but then use `eq(...)` to do arg matching on the others...or else mockito complains :)

harageth: ahhhh yeah thats the part I was missing/forgetting

src/test/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiControllerTest.java
```diff
@@ -1015,6 +1016,36 @@ public void testCloneMonitor() throws Exception {
...
+  @Test
+  public void testSearchMonitor() throws Exception {
+    String tenantId = RandomStringUtils.randomAlphabetic(8);
+
+    Monitor monitor = podamFactory.manufacturePojo(Monitor.class);
+    monitor.setTenantId(tenantId);
+    monitor.setSelectorScope(ConfigSelectorScope.LOCAL);
+    monitor.setAgentType(AgentType.TELEGRAF);
+    monitor.setContent("{\"type\":\"mem\"}");
+    monitor.setLabelSelectorMethod(LabelSelectorMethod.OR);
```
jjbuchan, line 1028: Doesn't look like you need to set any of these fields?  Could just leave it as `Monitor monitor = podamFactory.manufacturePojo(Monitor.class);` with none of the extra stuff?

PR #186, opened 2020-07-27
added getTestMonitor functionality

shabd67:
# Resolves

[SALUS-569](https://jira.rax.io/browse/SALUS-569)

# What

Added getTestMonitor() functionality to MonitorApi to chain test-monitor and test-task operations via single API

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApi.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApi.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Rackspace US, Inc.
+ * Copyright 2020 Rackspace US, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,7 +16,11 @@
 
 package com.rackspace.salus.monitor_management.web.client;
 
-import com.rackspace.salus.monitor_management.web.model.*;
+import com.rackspace.salus.monitor_management.web.model.BoundMonitorDTO;
+import com.rackspace.salus.monitor_management.web.model.DetailedMonitorInput;
+import com.rackspace.salus.monitor_management.web.model.DetailedMonitorOutput;
+import com.rackspace.salus.monitor_management.web.model.TestMonitorInput;
+import com.rackspace.salus.monitor_management.web.model.TestMonitorOutput;
 import com.rackspace.salus.telemetry.model.AgentType;
 import java.util.List;
 import java.util.Map;
--- a/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApi.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApi.java
@@ -31,5 +31,5 @@ public interface MonitorApi {
 
   DetailedMonitorOutput createMonitor(String tenantId, DetailedMonitorInput input, MultiValueMap<String, String> headers);
 
-  TestMonitorOutput getTestMonitor(String tenantId, TestMonitorInput input);
+  TestMonitorOutput performTestMonitor(String tenantId, TestMonitorInput input, MultiValueMap<String, String> headers);
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApi.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApi.java
@@ -16,9 +16,7 @@
 
 package com.rackspace.salus.monitor_management.web.client;
 
-import com.rackspace.salus.monitor_management.web.model.BoundMonitorDTO;
-import com.rackspace.salus.monitor_management.web.model.DetailedMonitorInput;
-import com.rackspace.salus.monitor_management.web.model.DetailedMonitorOutput;
+import com.rackspace.salus.monitor_management.web.model.*;
 import com.rackspace.salus.telemetry.model.AgentType;
 import java.util.List;
 import java.util.Map;
@@ -32,4 +30,6 @@ public interface MonitorApi {
   DetailedMonitorOutput getPolicyMonitorById(String monitorId);
 
   DetailedMonitorOutput createMonitor(String tenantId, DetailedMonitorInput input, MultiValueMap<String, String> headers);
+
+  TestMonitorOutput getTestMonitor(String tenantId, TestMonitorInput input);
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApi.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApi.java
@@ -31,5 +31,5 @@ public interface MonitorApi {
 
   DetailedMonitorOutput createMonitor(String tenantId, DetailedMonitorInput input, MultiValueMap<String, String> headers);
 
-  TestMonitorOutput performTestMonitor(String tenantId, TestMonitorInput input, MultiValueMap<String, String> headers);
+  TestMonitorOutput performTestMonitor(String tenantId, TestMonitorInput input);
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Rackspace US, Inc.
+ * Copyright 2020 Rackspace US, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
```diff
@@ -18,10 +18,7 @@
 
 import static com.rackspace.salus.common.web.RemoteOperations.mapRestClientExceptions;
 
-import com.rackspace.salus.monitor_management.web.model.BoundMonitorDTO;
-import com.rackspace.salus.monitor_management.web.model.BoundMonitorsRequest;
-import com.rackspace.salus.monitor_management.web.model.DetailedMonitorInput;
-import com.rackspace.salus.monitor_management.web.model.DetailedMonitorOutput;
+import com.rackspace.salus.monitor_management.web.model.*;
```
jjbuchan, line 21: If you're using IntelliJ and select `Code -> Optimize Imports` I believe it will reset this to the previous way?

I actually have it configured to automatically optimize imports on the fly, so I never have to add/remove anything up here.

src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
```diff
@@ -140,4 +137,23 @@ public DetailedMonitorOutput createMonitor(String tenantId, DetailedMonitorInput
...
+
+  @Override
+  public TestMonitorOutput getTestMonitor(String tenantId, TestMonitorInput input) {
+    String uriString = UriComponentsBuilder
+            .fromUriString("/api/tenant/{tenantId}/test-monitor")
+            .buildAndExpand(tenantId)
+            .toUriString();
+
+    HttpHeaders reqHeaders = new HttpHeaders();
+    reqHeaders.setContentType(MediaType.APPLICATION_JSON);
```
jjbuchan, line 149: Since you're creating the header manually here, you'll probably want to do similar to the `createMonitor` method and allow for additional headers to be passed through and included in the request.

Although, I can't quite remember why `createMonitor` had to set the content-type in that way.

src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApi.java
```diff
@@ -32,4 +30,6 @@
   DetailedMonitorOutput getPolicyMonitorById(String monitorId);
 
   DetailedMonitorOutput createMonitor(String tenantId, DetailedMonitorInput input, MultiValueMap<String, String> headers);
+
+  TestMonitorOutput getTestMonitor(String tenantId, TestMonitorInput input);
```
jjbuchan, line 34: One more comment.  Maybe this is better named `performTestMonitor`?

src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
itzg: I need to add this to the onboarding guide, but Google's Java coding style gets pretty close to the style we all use. So you can import this into IntelliJ

https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml

using 

image

src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
itzg: Was it because of supporting the patch operation perhaps?

src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApi.java
```diff
@@ -16,9 +16,7 @@
 
 package com.rackspace.salus.monitor_management.web.client;
 
-import com.rackspace.salus.monitor_management.web.model.BoundMonitorDTO;
-import com.rackspace.salus.monitor_management.web.model.DetailedMonitorInput;
-import com.rackspace.salus.monitor_management.web.model.DetailedMonitorOutput;
+import com.rackspace.salus.monitor_management.web.model.*;
```
itzg, line 19: ...and this one needs the imports "optimized"

src/main/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClient.java
jjbuchan: Probably not for the create.  I kinda wonder if this is masking another problem.

I know I've hit issues in Insomnia where it gives some kind of error about the content type.  Whatever those were I was leaving until later where we deal with all our error handling.

jjbuchan: So @shabd67 unless you know that it's required to set the header here, I'd actually say we should leave it out (like the `getBoundMonitors` method).

Probably better to leave it out than include it without knowing why.  Same goes for the event-management PR.

src/test/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClientTest.java
```diff
@@ -130,4 +145,18 @@ public void testGetPolicyMonitor_doesntExist() throws JsonProcessingException {
...
   }
+
+  @Test
+  public void testPerformTestMonitor() throws IOException {
+    String expectedReqJson = readContent("MonitorApiClientTest/testPerformTestMonitor_req.json");
+    String tenantId = RandomStringUtils.randomAlphabetic(8);
+    String url = String.format("/api/tenant/%s/test-monitor", tenantId);
+    TestMonitorOutput monitor = podamFactory.manufacturePojo(TestMonitorOutput.class);
+
+    mockServer.expect(requestTo(url))
```
itzg, line 156: FYI, we haven't switch this unit test over to it, but there's a `requestToUriTemplate` that allows for templated URLs like controller declarations. Keep yours as-is since it is consistent with others in this class.

src/test/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClientTest.java
```diff
@@ -130,4 +145,18 @@ public void testGetPolicyMonitor_doesntExist() throws JsonProcessingException {
     DetailedMonitorOutput result = monitorApiClient.getPolicyMonitorById("id");
     assertThat(result, nullValue());
   }
+
+  @Test
+  public void testPerformTestMonitor() throws IOException {
+    String expectedReqJson = readContent("MonitorApiClientTest/testPerformTestMonitor_req.json");
+    String tenantId = RandomStringUtils.randomAlphabetic(8);
+    String url = String.format("/api/tenant/%s/test-monitor", tenantId);
+    TestMonitorOutput monitor = podamFactory.manufacturePojo(TestMonitorOutput.class);
```
itzg, line 154: Can you call this `testMonitorOutput` to avoid confusion? It's a small unit test, but good practice for bigger unit tests where the variable naming can help a lot.

src/test/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClientTest.java
```diff
@@ -130,4 +145,18 @@ public void testGetPolicyMonitor_doesntExist() throws JsonProcessingException {
...
+
+    mockServer.expect(requestTo(url))
+        .andExpect(method(HttpMethod.POST))
+        .andExpect(content().json(expectedReqJson, true))
+        .andRespond(
+            withSuccess(objectMapper.writeValueAsString(monitor), MediaType.APPLICATION_JSON));
+  }
```
itzg, line 161: You're missing a call to `monitorApiClient`, so it turns out the unit test isn't actually being exercised.

src/test/resources/MonitorApiClientTest/testPerformTestMonitor_req.json
```diff
@@ -0,0 +1,9 @@
+{
+  "resourceId": "{% prompt 'Resource ID', '', 'development:0', '', false, true %}",
```
itzg, line 2: This Insomnia template placeholder shouldn't be here. This JSON content will need to line up with the POJO that pass into the missing `monitorApiClient` call in your unit test.

src/test/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClientTest.java
```diff
@@ -149,14 +142,18 @@ public void testGetPolicyMonitor_doesntExist() throws JsonProcessingException {
...
     String expectedReqJson = readContent("MonitorApiClientTest/testPerformTestMonitor_req.json");
+    TestMonitorInput testMonitorInput = objectMapper.readValue(expectedReqJson, TestMonitorInput.class);
+
     String tenantId = RandomStringUtils.randomAlphabetic(8);
-    String url = String.format("/api/tenant/%s/test-monitor", tenantId);
-    TestMonitorOutput monitor = podamFactory.manufacturePojo(TestMonitorOutput.class);
+    TestMonitorOutput testMonitorOutput = podamFactory.manufacturePojo(TestMonitorOutput.class);
 
-    mockServer.expect(requestTo(url))
+    mockServer.expect(requestToUriTemplate("/api/tenant/{tenantId}/test-monitor",tenantId))
```
jjbuchan: nit

```suggestion
    mockServer.expect(requestToUriTemplate("/api/tenant/{tenantId}/test-monitor", tenantId))
```

PR #189, opened 2020-08-06
Added validation support for configurable properties classes

rohitsngh27:
# Resolves
https://jira.rax.io/browse/SALUS-977

# What
Added validation support for configurable properties classes

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/config/ZonesProperties.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/config/ZonesProperties.java
@@ -16,9 +16,6 @@
 
 package com.rackspace.salus.monitor_management.config;
 
-import java.util.Collections;
-import java.util.List;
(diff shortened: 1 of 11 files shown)


src/main/java/com/rackspace/salus/monitor_management/config/ZonesProperties.java
```diff
@@ -16,19 +16,17 @@
...
 import org.springframework.stereotype.Component;
+import org.springframework.validation.annotation.Validated;
 
 @ConfigurationProperties("salus.zones")
 @Component
 @Data
+@Validated
 public class ZonesProperties {
 
-  List<String> defaultZones = Collections.emptyList();
```
itzg, line 30: Just wanted to note for other reviewers that I discovered this wasn't being used any more. I'm guessing after policy-management was added, this was no longer used.

PR #195, opened 2020-08-14
added code to delete a resource and its associated bound-monitor.

shabd67:
# Resolves

[SALUS-980](https://jira.rax.io/browse/SALUS-980)

# What

Resolved - Deleting resource also deletes bound monitors for other resources.

# Depends on

https://github.com/racker/salus-telemetry-model/pull/148

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1681,7 +1681,8 @@ public class MonitorManagement {
       return;
     }
 
-    final List<Monitor> monitorsWithResourceId = monitorRepository.findByTenantIdAndResourceId(tenantId, resourceId);
+    final List<Monitor> monitorsWithResourceId = monitorRepository
+        .findByTenantIdAndResourceId(tenantId, resourceId);
 
     final List<UUID> boundMonitorIds =
         boundMonitorRepository.findMonitorIdsBoundToTenantAndResource(tenantId, resourceId);
@@ -1691,7 +1692,8 @@ public class MonitorManagement {
     final Set<UUID> monitorIdsToUnbind = new HashSet<>(boundMonitorIds);
 
     List<Monitor> selectedMonitors;
-    final Optional<Resource> resource = resourceRepository.findByTenantIdAndResourceId(tenantId, resourceId);
+    final Optional<Resource> resource = resourceRepository
+        .findByTenantIdAndResourceId(tenantId, resourceId);
     if (resource.isPresent()) {
       // resource created or updated
 
@@ -1718,16 +1720,20 @@ public class MonitorManagement {
 
       // ...the setminus operation upon monitorIdsToUnbind
       monitorIdsToUnbind.removeAll(selectedMonitorIds);
-    }
-    else {
+    } else {
       // resource deleted
 
       selectedMonitors = Collections.emptyList();
       // ...and monitorIdsToUnbind remains ALL of the currently bound
     }
 
+    Set<String> affectedEnvoys = null;
     // this needs to be updated to only unbind my tenant and resourceId
-    final Set<String> affectedEnvoys = unbindByTenantAndResourceId(tenantId, event.getResourceId(), monitorIdsToUnbind);
+    if (monitorIdsToUnbind.isEmpty()) {
+      affectedEnvoys = new HashSet<>();
+    } else  {
+      affectedEnvoys = unbindByTenantAndResourceId(tenantId, event.getResourceId());
+    }
 
     if (!selectedMonitors.isEmpty()) {
       affectedEnvoys.addAll(
@@ -1912,13 +1918,9 @@ public class MonitorManagement {
    * @return affected envoy IDs
    */
   Set<String> unbindByTenantAndResourceId(String tenantId,
-      String resourceId, Collection<UUID> monitorIdsToUnbind) {
-    if (monitorIdsToUnbind.isEmpty()) {
-      return new HashSet<>();
-    }
-
+      String resourceId) {
     final List<BoundMonitor> boundMonitors = boundMonitorRepository
-        .findMonitorsBoundToResourceAndTenant(tenantId, resourceId);
+        .findMonitorsBoundToTenantAndResource(tenantId, resourceId);
 
     log.debug("Unbinding {} from resourceId={}",
         boundMonitors, resourceId);
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1684,7 +1684,7 @@ public class MonitorManagement {
     final List<Monitor> monitorsWithResourceId = monitorRepository.findByTenantIdAndResourceId(tenantId, resourceId);
 
     final List<UUID> boundMonitorIds =
-        boundMonitorRepository.findMonitorsBoundToResource(tenantId, resourceId);
+        boundMonitorRepository.findMonitorIdsBoundToTenantAndResource(tenantId, resourceId);
 
     // monitorIdsToUnbind := boundMonitorIds \setminus selectedMonitorIds
     // ...so start with populating with boundMonitorIds
@@ -1727,7 +1727,7 @@ public class MonitorManagement {
     }
 
     // this needs to be updated to only unbind my tenant and resourceId
-    final Set<String> affectedEnvoys = unbindByTenantAndResourceId(tenantId, event.getResourceId());
+    final Set<String> affectedEnvoys = unbindByTenantAndResourceId(tenantId, event.getResourceId(), monitorIdsToUnbind);
 
     if (!selectedMonitors.isEmpty()) {
       affectedEnvoys.addAll(
@@ -1912,7 +1912,11 @@ public class MonitorManagement {
    * @return affected envoy IDs
    */
   Set<String> unbindByTenantAndResourceId(String tenantId,
-      String resourceId) {
+      String resourceId, Collection<UUID> monitorIdsToUnbind) {
+    if (monitorIdsToUnbind.isEmpty()) {
+      return new HashSet<>();
+    }
+
     final List<BoundMonitor> boundMonitors = boundMonitorRepository
         .findMonitorsBoundToResourceAndTenant(tenantId, resourceId);
 
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1726,8 +1726,8 @@ public class MonitorManagement {
       // ...and monitorIdsToUnbind remains ALL of the currently bound
     }
 
-    // this needs to be updated to only unbind my tenant and monitor id
-    final Set<String> affectedEnvoys = unbindByTenantAndMonitorId(tenantId, monitorIdsToUnbind);
+    // this needs to be updated to only unbind my tenant and resourceId
+    final Set<String> affectedEnvoys = unbindByTenantAndResourceId(tenantId, event.getResourceId());
 
     if (!selectedMonitors.isEmpty()) {
       affectedEnvoys.addAll(
@@ -1900,6 +1900,24 @@ public class MonitorManagement {
 
     log.debug("Unbinding {} from monitorIds={}",
         boundMonitors, monitorIdsToUnbind);
+    boundMonitorRepository.deleteAll(boundMonitors);
+    decrementBoundCounts(boundMonitors);
+
+    return extractEnvoyIds(boundMonitors);
+  }
+
+  /**
+   * Removes all bindings associated with the given monitor IDs and resource id.
+   *
+   * @return affected envoy IDs
+   */
+  Set<String> unbindByTenantAndResourceId(String tenantId,
+      String resourceId) {
+    final List<BoundMonitor> boundMonitors = boundMonitorRepository
+        .findMonitorsBoundToResourceAndTenant(tenantId, resourceId);
+
+    log.debug("Unbinding {} from resourceId={}",
+        boundMonitors, resourceId);
     boundMonitorRepository.deleteAll(boundMonitors);
     decrementBoundCounts(boundMonitors);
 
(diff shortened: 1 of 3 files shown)


itzg:
FYI in the PR description I like to note when a PR depends on changes from another, such as https://github.com/racker/salus-telemetry-model/pull/147 in this case. 

I add a section at the end like the following (which we should probably add to the template).

# Depends on

- https://github.com/racker/salus-telemetry-model/pull/147

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -3922,19 +3922,13 @@ public void testhandleResourceEvent_removedResource() {
...
-    );
+    verify(boundMonitorRepository).findMonitorIdsBoundToTenantAndResource("t-1", "r-1");
 
     verify(boundMonitorRepository).deleteAll(
         Collections.singletonList(boundMonitor)
     );
 
     verifyNoInteractions(policyApi);
-    verifyNoMoreInteractions(boundMonitorRepository, envoyResourceManagement,
-        zoneStorage, monitorEventProducer, resourceApi, resourceRepository);
```
GeorgeJahad, line 3937: did this need to be removed?

shabd67: have to remove boundMonitorRepository from verifyNoMoreInteractions inputs because there would be an interaction on boundMonitorRepository for resource deletion. hence, pushing that commit.

harageth: It would be better to verify that interaction rather than removing this line.

shabd67: added "verifyNoMoreInteractions(envoyResourceManagement,zoneStorage, monitorEventProducer, resourceApi, resourceRepository);"

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1906,6 +1906,28 @@ private void handleReattachedEnvoy(String tenantId, String resourceId, String en
     return extractEnvoyIds(boundMonitors);
   }
 
+  /**
+   * Removes all bindings associated with the given monitor IDs and resource id.
+   *
+   * @return affected envoy IDs
+   */
+  Set<String> unbindByTenantAndResourceId(String tenantId,
+      String resourceId, Collection<UUID> monitorIdsToUnbind) {
```
harageth, line 1915: It doesn't appear that `monitorIdsToUnbind` is really necessary in this function. It should probably be removed and understood in the previous function what it is actually trying to accomplish.

src/test/java/com/rackspace/salus/monitor_management/services/ResourceEventListenerTest.java
```diff
@@ -79,4 +79,17 @@ public void testReattachedEnvoyResourceEvent() throws InterruptedException {
...
+  public void testResourceDeletedEvent() throws InterruptedException {
+    final ResourceEvent event = new ResourceEvent()
+        .setLabelsChanged(false)
+        .setTenantId("t-1")
+        .setResourceId("r-1")
+        .setDeleted(true);
+
+    kafkaTemplate.send(TOPIC, "t-1:r-1", event);
+
+    verify(monitorManagement, after(5000)).handleResourceChangeEvent(event);
```
GeorgeJahad, line 93: I recently discovered mockito "timeout()" which seems better than "after()" in most cases because it doesn't force you to wait the whole time if not necessary.  Take a look: https://site.mockito.org/javadoc/current/org/mockito/verification/VerificationWithTimeout.html

src/test/java/com/rackspace/salus/monitor_management/services/ResourceEventListenerTest.java
```diff
@@ -79,4 +79,17 @@ public void testReattachedEnvoyResourceEvent() throws InterruptedException {
...
+  }
```
GeorgeJahad, line 94: is it worth verifying this was called as well: "unbindByTenantAndResourceId()"?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
shabd67: actually this function has been invoked from other classes as well where a check on `monitorIdsToUnbind` is there. So, have used `monitorIdsToUnbind ` with `if (monitorIdsToUnbind.isEmpty()) {
      return new HashSet<>();
    }`

src/test/java/com/rackspace/salus/monitor_management/services/ResourceEventListenerTest.java
shabd67: okay. let me have a look onto this.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
harageth: I guess my point is that that over complicates this function. It creates an impetus that the previous function has found the list of ids to unbind already while this function is doing the same thing in a different way that is not going to find the same id's.

src/test/java/com/rackspace/salus/monitor_management/services/ResourceEventListenerTest.java
shabd67: Not really, but added this as method I changed was invoked from `ResourceEventListener`.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1718,16 +1720,20 @@ void handleResourceChangeEvent(ResourceEvent event) {
...
     }
 
-    // this needs to be updated to only unbind my tenant and monitor id
-    final Set<String> affectedEnvoys = unbindByTenantAndMonitorId(tenantId, monitorIdsToUnbind);
+    Set<String> affectedEnvoys = null;
+    // this needs to be updated to only unbind my tenant and resourceId
+    if (monitorIdsToUnbind.isEmpty()) {
+      affectedEnvoys = new HashSet<>();
+    } else  {
+      affectedEnvoys = unbindByTenantAndResourceId(tenantId, event.getResourceId());
```
itzg, line 1735: Looking closer at the existing code it looks like this still needs to consider `monitorIdsToUnbind` since in the case of

```
    if (resource.isPresent()) {
      // resource created or updated
```

then it's possible

1. resource was modified, such as a metadata change
2. labels of that resource are still selected by previously bound monitors, which means bound monitors should remain in place
3. with this new change to `unbindByTenantAndResourceId(tenantId, event.getResourceId())` those bound monitors are now getting removed regardless of unchanged binding-ness

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1906,6 +1912,24 @@ private void handleReattachedEnvoy(String tenantId, String resourceId, String en
     return extractEnvoyIds(boundMonitors);
   }
 
+  /**
+   * Removes all bindings associated with the given monitor IDs and resource id.
+   *
+   * @return affected envoy IDs
+   */
+  Set<String> unbindByTenantAndResourceId(String tenantId,
+      String resourceId) {
```
itzg, line 1921: My comment above would be easy to address by
```suggestion
  Set<String> unbindByTenantAndResourceId(String tenantId,
      String resourceId, Set<UUID> monitorIdsToUnbind) {
```

and then a stream-filter on the `boundMonitors` retrieved by the findMonitorsBoundToTenantAndResource to include only those in the given set. Could even enhance the JPA query itself to add an `AND IN(:monitorIdsToUnbind)` but I don't feel strongly either way. @GeorgeJahad @harageth any preference to filter via JPA/SQL query or streams?

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
itzg: Agreed, this needs to be put back and the proper verification added. The point of `verifyNoMoreInteractions` to catch any escapes in the unit test code.

src/test/java/com/rackspace/salus/monitor_management/services/ResourceEventListenerTest.java
```diff
@@ -77,6 +78,19 @@ public void testReattachedEnvoyResourceEvent() throws InterruptedException {
 
     kafkaTemplate.send(TOPIC, "t-1:r-1", event);
 
-    verify(monitorManagement, after(5000)).handleResourceChangeEvent(event);
+    verify(monitorManagement, timeout(5000)).handleResourceChangeEvent(event);
```
itzg, line 81: It does seem more correct, but was there a reason that prompted this change?

src/test/java/com/rackspace/salus/monitor_management/services/ResourceEventListenerTest.java
itzg: Since `MonitorManagement` is mocked, there won't be any call to `unbindByTenantAndResourceId`. Speaking of which, I don't think this test suite needed to change at all since `ResourceEventListener` wasn't changed and everything around its test is mocked.

src/test/java/com/rackspace/salus/monitor_management/services/ResourceEventListenerTest.java
shabd67: it was George's idea so that we shouldn't wait for whole time if not necessary. 
[George's comment](https://github.com/racker/salus-telemetry-monitor-management/pull/195#discussion_r471601464)

shabd67: And verifyNoMoreInteractions has been put back with `verifyNoMoreInteractions(envoyResourceManagement,
        zoneStorage, monitorEventProducer, resourceApi, resourceRepository);`

itzg: Ah, github had hidden that comment from me.

itzg: Can you double check since the PR diff still shows the removal 

image

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
shabd67: yes, I was thinking to put `AND IN(:monitorIdsToUnbind)` in JPA query but was not aware of that context you explained in above comments.

src/test/java/com/rackspace/salus/monitor_management/services/ResourceEventListenerTest.java
shabd67: So, should I go for removing this test case then as there has been no changes in ResourceEventListener ?

itzg: Yes, I think you can remove this test case.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
itzg: Yeah, let's go with the JPA query change then.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -3922,18 +3922,14 @@ public void testhandleResourceEvent_removedResource() {
...
-    );
+    verify(boundMonitorRepository).findMonitorIdsBoundToTenantAndResource("t-1", "r-1");
 
     verify(boundMonitorRepository).deleteAll(
         Collections.singletonList(boundMonitor)
     );
 
     verifyNoInteractions(policyApi);
-    verifyNoMoreInteractions(boundMonitorRepository, envoyResourceManagement,
+    verifyNoMoreInteractions(envoyResourceManagement,
```
shabd67, line 3932: here, it is. @itzg

itzg: Yes, that's the change I'm talking about. You need to revert your change and put back `boundMonitorRepository`.

itzg: I'm really confused about what github is doing. Your code does seem fine when I check out your branch.

itzg: And confirmed the diff with IntelliJ by checking out the PR's (correct) branch

image

shabd67: but in this case while removing a resource, there will be interaction to boundMonitorRepository. isn't it?

itzg: Yes, so please fix by addressing @harageth 's comment

https://github.com/racker/salus-telemetry-monitor-management/pull/195/files#r471596025

itzg:
After merging https://github.com/racker/salus-telemetry-model/pull/148 can you re-run this PR's check to confirm that build passes?

shabd67:
> After merging [racker/salus-telemetry-model#148](https://github.com/racker/salus-telemetry-model/pull/148) can you re-run this PR's check to confirm that build passes?

Sure.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1718,16 +1720,15 @@ void handleResourceChangeEvent(ResourceEvent event) {
...
+    } else {
       // resource deleted
 
       selectedMonitors = Collections.emptyList();
       // ...and monitorIdsToUnbind remains ALL of the currently bound
     }
 
-    // this needs to be updated to only unbind my tenant and monitor id
-    final Set<String> affectedEnvoys = unbindByTenantAndMonitorId(tenantId, monitorIdsToUnbind);
+    // this needs to be updated to only unbind my tenant and resourceId
```
jjbuchan, line 1730: Is this comment relevant?  Doesn't this PR fix what seemed to be the original reason of the comment?

PR #196, opened 2020-08-19
Added inter-service admin endpoint to translate monitor content

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-779

# What

Test-monitor operations were not going through monitor translation processing, which meant that telegraf was sometimes being given Salus-specific monitor/plugin configs that it didn't comprehend.

# How

Generalized and exposed an existing method of `MonitorContentTranslationService` so that it can be used for both bulk BoundMonitor translations and one-off content translations. In turn a new inter-service, admin REST API endpoint was added to enable calls to this endpoint, such as from the Ambassador building config instruction.

# How to test

Unit tests updated to cover the new method/endpoint.

### DIFF
--- a/src/test/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClientTest.java
+++ b/src/test/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClientTest.java
@@ -17,9 +17,7 @@
 package com.rackspace.salus.monitor_management.web.client;
 
 import static com.rackspace.salus.common.util.SpringResourceUtils.readContent;
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.nullValue;
-import static org.junit.Assert.assertThat;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.springframework.test.web.client.match.MockRestRequestMatchers.content;
 import static org.springframework.test.web.client.match.MockRestRequestMatchers.method;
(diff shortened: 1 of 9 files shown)


src/test/java/com/rackspace/salus/monitor_management/web/client/MonitorApiClientTest.java
```diff
@@ -115,7 +114,7 @@ public void testQueryBoundMonitors() throws IOException {
     final Map<AgentType, String> installedAgentVersions = Map.of(AgentType.TELEGRAF, "1.12.0");
     final List<BoundMonitorDTO> boundMonitors =
         monitorApiClient.getBoundMonitors("e-1", installedAgentVersions);
-    assertThat(boundMonitors, equalTo(givenBoundMonitors));
+    assertThat(boundMonitors).isEqualTo(givenBoundMonitors);
```
GeorgeJahad, line 117: for my education, how is this better?

itzg: I changed it because I just happened to notice IntelliJ was reporting a deprecation warning. Apparently there's an "assertThat" in JUnit 4, Hamcrest, and AssertJ. The end result here and what I'm trying to use in new unit tests is the AssertJ style, which I have grown to really like:

https://assertj.github.io/doc/

PR #198, opened 2020-08-21
Added monitor type and scope for test-monitor content translation

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-779

# What

The previous PR https://github.com/racker/salus-telemetry-monitor-management/pull/196 didn't take into consideration the monitor type and scope. As a result, the content translation for test-monitors was trying to apply ALL of the translation operations, which failed every time due to conversion of fields that didn't exist.

# How

Propagate the monitor type and scope of the original test-monitor request and use those in the same way as translations for bound-monitors.

While I was in there, I noticed that the translation operators retrieval was sorting by `agentVersions` and then later using the `order` field. The latter is what we are really using for deterministic ordering. 

# How to test

Added unit tests for the newly elevated methods.

# Depends on

https://github.com/racker/salus-telemetry-model/pull/149

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/web/model/TranslateMonitorContentRequest.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/model/TranslateMonitorContentRequest.java
@@ -17,6 +17,8 @@
 package com.rackspace.salus.monitor_management.web.model;
 
 import com.rackspace.salus.telemetry.model.AgentType;
+import com.rackspace.salus.telemetry.model.ConfigSelectorScope;
+import com.rackspace.salus.telemetry.model.MonitorType;
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
 import lombok.Data;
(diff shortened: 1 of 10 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/model/TranslateMonitorContentRequest.java
```diff
@@ -29,4 +31,8 @@
   AgentType agentType;
   @NotBlank
   String agentVersion;
+  @NotNull
+  MonitorType monitorType;
+  @NotNull
+  ConfigSelectorScope scope;
```
jjbuchan, line 37: Makes sense for now, but I could imagine use-cases appear where we'd want to pass down null for these two.

Good with it as is, just wanted to state the point.

itzg: I'm not understanding your comment since the monitor translation requires monitor type and scope to narrow the applicable operators. That was the aspect I previously missed that led to this current PR.

jjbuchan: Yes, ignore that.  I think my mind was going in the direction of retrieving policies not translations.

PR #199, opened 2020-08-26
Updated patch api when resource does not exists

rohitsngh27:
# Resolves
https://jira.rax.io/browse/SALUS-938

# What
Updated patch api to give 200 response when resource does not exists

# How
Added a null check which was the cause of failure before

# How to test
This can be verified via REST call

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1236,7 +1236,7 @@ public class MonitorManagement {
     // If a new one is to be bound, bind it
     if (StringUtils.isNotBlank(updatedResourceId)) {
       ResourceDTO resource = resourceApi.getByResourceId(monitor.getTenantId(), updatedResourceId);
-      if (!ObjectUtils.isEmpty(resource)) {
+      if (resource != null) {
         affectedEnvoys.addAll(
             upsertBindingToResource(
                 Collections.singletonList(monitor),
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -110,6 +110,7 @@ import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
+import org.springframework.util.ObjectUtils;
 
 
 @Slf4j
@@ -1234,13 +1235,15 @@ public class MonitorManagement {
 
     // If a new one is to be bound, bind it
     if (StringUtils.isNotBlank(updatedResourceId)) {
-      ResourceDTO resource  = resourceApi.getByResourceId(monitor.getTenantId(), updatedResourceId);
-      affectedEnvoys.addAll(
-          upsertBindingToResource(
-              Collections.singletonList(monitor),
-              resource,
-              null
-          ));
+      ResourceDTO resource = resourceApi.getByResourceId(monitor.getTenantId(), updatedResourceId);
+      if (!ObjectUtils.isEmpty(resource)) {
+        affectedEnvoys.addAll(
+            upsertBindingToResource(
+                Collections.singletonList(monitor),
+                resource,
+                null
+            ));
+      }
     }
 
     return affectedEnvoys;
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -110,7 +110,6 @@ import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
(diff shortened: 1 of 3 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1234,13 +1235,15 @@ private static boolean labelSelectorMethodChanged(Monitor original, MonitorCU ne
...
     if (StringUtils.isNotBlank(updatedResourceId)) {
-      ResourceDTO resource  = resourceApi.getByResourceId(monitor.getTenantId(), updatedResourceId);
-      affectedEnvoys.addAll(
-          upsertBindingToResource(
-              Collections.singletonList(monitor),
-              resource,
-              null
-          ));
+      ResourceDTO resource = resourceApi.getByResourceId(monitor.getTenantId(), updatedResourceId);
+      if (!ObjectUtils.isEmpty(resource)) {
```
itzg, line 1239: Just this would work, right?
```suggestion
      if (resource != null) {
```

rohitsngh27: Yes This would work too

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -110,6 +110,7 @@
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
+import org.springframework.util.ObjectUtils;
```
jjbuchan, line 113: Doesn't look like this import is needed.

If you haven't already it can be a good idea to enable this optimize imports feature in IntelliJ.  It'll clean up any unused imports as you are working on the file.

image

rohitsngh27: Sure. Will resolve this

rohitsngh27:
> Per my last comment, I'd still like you to add a unit test to this PR for the original bug.
> 
> Once we have that, I'd be good with approving this to be merged.

I have added the unit test case as suggested by you

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4378,4 +4378,71 @@ public void deleteAllMonitorsForTenant_noEvents() {
...
+        .setInterval(monitor.getInterval())
+        .setSelectorScope(monitor.getSelectorScope())
+        .setPluginMetadataFields(monitor.getPluginMetadataFields());
+
+    final Monitor updatedMonitor = monitorManagement
+        .updateMonitor("t-1", monitor.getId(), update, true);
+
+    // VERIFY
+    // confirm monitor is updated
+    org.assertj.core.api.Assertions.assertThat(Collections.singleton(updatedMonitor))
```
itzg, line 4429: A static import of `assertThat` would be good.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4378,4 +4378,71 @@ public void deleteAllMonitorsForTenant_noEvents() {
...
+        .thenReturn(null);
+    when(resourceRepository.findByTenantIdAndResourceId("t-1", "r-1"))
+        .thenReturn(Optional.empty());
+
+    // Called when unbinding old resources
+    when(boundMonitorRepository.findResourceIdsBoundToMonitor(any()))
+        .thenReturn(Collections.emptySet());
+
+    // EXECUTE
+    String newResourceIdBinding = "r-1";
```
jjbuchan, line 4408:
```suggestion
    String newResourceIdBinding = RandomStringUtils.randomAlphanumeric(10);
```
One small suggestion.  Even though we have many tests that do it, I'd recommend not hardcoding values unless there's a specific reason for it.  We've seen issues in the past where setting things like `r-1` for resources all the time has hidden bugs that would have been exposed by randomizing the value instead.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4378,4 +4378,71 @@ public void deleteAllMonitorsForTenant_noEvents() {
...
+                .setContent("static content")
+                .setMonitorMetadataFields(List.of("monitorName"))
+                .setTenantId("t-1")
+                .setSelectorScope(ConfigSelectorScope.LOCAL)
+                .setResourceId(newResourceIdBinding)
+                .setLabelSelector(null)
+                .setLabelSelectorMethod(LabelSelectorMethod.AND)
+                .setInterval(Duration.ofSeconds(60)));
+
+    verifyNoMoreInteractions(monitorEventProducer);
```
jjbuchan, line 4445: It's a good idea to `verify` all of (or at least the most important `when`s from above).

It helps verifying the methods you are mocking are actually getting hit.

e.g. the first one would be
```
verify(resourceApi).getByResourceId("t-1", "r-1");
```

also keep in mind that in `verify` you do not want to use `any()` like you might in the `when`.  Instead you should be verifying the exact value you expect to be passed to the method call.  I think in this case `monitor.getId()` might be the right thing for your final `verify`.

PR #200, opened 2020-08-28
Added api to get agent config details

rohitsngh27:
# Resolves
https://jira.rax.io/browse/SALUS-485

# What
API to get agent config details for particular monitor

# How
A POST REST endpoint is created which returns the agent config details

# How to test
This can be tested via REST call and unit test cases.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -32,6 +32,7 @@ import com.rackspace.salus.monitor_management.errors.DeletionNotAllowedException
 import com.rackspace.salus.monitor_management.errors.InvalidTemplateException;
 import com.rackspace.salus.monitor_management.utils.MetadataUtils;
 import com.rackspace.salus.monitor_management.web.model.MonitorCU;
+import com.rackspace.salus.monitor_management.web.model.RenderedMonitorTemplate;
 import com.rackspace.salus.monitor_management.web.model.ZoneAssignmentCount;
 import com.rackspace.salus.monitor_management.web.validator.ValidUpdateMonitor;
 import com.rackspace.salus.policy.manage.web.client.PolicyApi;
@@ -2287,5 +2288,37 @@ public class MonitorManagement {
               .collect(Collectors.toList()));
       monitorRepository.deleteAllByTenantId(tenantId);
     }
+  }
+
+  public RenderedMonitorTemplate renderMonitorTemplate(UUID monitorId, String resourceId,
+      String tenantId) {
+    //Get existing monitor
+    Monitor monitor = getMonitor(tenantId, monitorId).orElseThrow(
+        () -> new NotFoundException(String.format("No policy monitor found for %s", monitorId)));
+
+    //Set Render monitor template
+    RenderedMonitorTemplate renderedMonitorTemplate = new RenderedMonitorTemplate();
+    renderedMonitorTemplate.setMonitor(monitor);
+    renderedMonitorTemplate.setRenderedContent(monitor.getContent());
+
+    if (StringUtils.isBlank(resourceId)) {
+      return renderedMonitorTemplate;
+    }
+    ResourceDTO resource = resourceApi.getByResourceId(tenantId, resourceId);
+
+    if (resource != null) {
+      try {
+        final String renderedContent = getRenderedContent(monitor.getContent(), resource);
+        renderedMonitorTemplate.setRenderedContent(renderedContent);
+      } catch (InvalidTemplateException e) {
+        log.warn("Unable to render content='{}' of monitor={} for resource={}",
+            monitor.getContent(), monitor, resource, e);
+        invalidTemplateErrors.increment();
+      }
+    } else {
+      log.warn("Failed to find resourceId={} during processing of monitor={}",
+          resourceId, monitor);
+    }
+    return renderedMonitorTemplate;
   }
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1234,13 +1234,15 @@ public class MonitorManagement {
 
     // If a new one is to be bound, bind it
     if (StringUtils.isNotBlank(updatedResourceId)) {
-      ResourceDTO resource  = resourceApi.getByResourceId(monitor.getTenantId(), updatedResourceId);
-      affectedEnvoys.addAll(
-          upsertBindingToResource(
-              Collections.singletonList(monitor),
-              resource,
-              null
-          ));
+      ResourceDTO resource = resourceApi.getByResourceId(monitor.getTenantId(), updatedResourceId);
+      if (resource != null) {
+        affectedEnvoys.addAll(
+            upsertBindingToResource(
+                Collections.singletonList(monitor),
+                resource,
+                null
+            ));
+      }
     }
 
     return affectedEnvoys;
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -2294,7 +2294,7 @@ public class MonitorManagement {
       String tenantId) {
     //Get existing monitor
     Monitor monitor = getMonitor(tenantId, monitorId).orElseThrow(
-        () -> new NotFoundException(String.format("No policy monitor found for %s", monitorId)));
+        () -> new NotFoundException(String.format("No monitor found for %s", monitorId)));
 
     //Set Render monitor template
     RenderedMonitorTemplate renderedMonitorTemplate = new RenderedMonitorTemplate();
@@ -2311,13 +2311,18 @@ public class MonitorManagement {
         final String renderedContent = getRenderedContent(monitor.getContent(), resource);
         renderedMonitorTemplate.setRenderedContent(renderedContent);
       } catch (InvalidTemplateException e) {
-        log.warn("Unable to render content='{}' of monitor={} for resource={}",
+        log.error("Unable to render content='{}' of monitor={} for resource={}",
             monitor.getContent(), monitor, resource, e);
-        invalidTemplateErrors.increment();
+        throw new IllegalArgumentException(
+            String.format("Unable to render content=%s of monitor=%s for resource=%s",
+                monitor.getContent(), monitor, resource));
       }
     } else {
-      log.warn("Failed to find resourceId={} during processing of monitor={}",
+      log.error("Failed to find resourceId={} during processing of monitor={}",
           resourceId, monitor);
+      throw new NotFoundException(
+          String.format("Failed to find resourceId=%s during processing of monitor=%s",
+              resourceId, monitorId));
     }
     return renderedMonitorTemplate;
   }
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -2314,8 +2314,8 @@ public class MonitorManagement {
         log.error("Unable to render content='{}' of monitor={} for resource={}",
             monitor.getContent(), monitor, resource, e);
         throw new IllegalArgumentException(
-            String.format("Unable to render content=%s of monitor=%s for resource=%s",
-                monitor.getContent(), monitor, resource));
+            String.format("Unable to render content=%s for resource=%s",
+                monitor.getContent(), resource));
       }
     } else {
       log.error("Failed to find resourceId={} during processing of monitor={}",
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -2318,11 +2318,11 @@ public class MonitorManagement {
                 monitor.getContent(), resource));
       }
     } else {
-      log.error("Failed to find resourceId={} during processing of monitor={}",
-          resourceId, monitor);
+      log.info("Invalid resourceId={} provided when rendering monitorId={} template for tenantId={}",
+          resourceId, monitorId, tenantId);
       throw new NotFoundException(
-          String.format("Failed to find resourceId=%s during processing of monitor=%s",
-              resourceId, monitorId));
+          String.format("Invalid resourceId=%s provided when rendering monitorId=%s template for tenantId=%s",
+              resourceId, monitorId, tenantId));
     }
     return renderedMonitorTemplate;
   }
--- a/src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
@@ -43,6 +43,7 @@ import io.swagger.annotations.Authorization;
 import io.swagger.annotations.AuthorizationScope;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
 import javax.json.JsonPatch;
 import lombok.extern.slf4j.Slf4j;
@@ -113,9 +114,9 @@ public class MonitorApiController {
     );
   }
 
-  @PostMapping(value="/admin/translate-monitor-content", produces = MediaType.TEXT_PLAIN_VALUE)
+  @PostMapping(value="/admin/translate-monitor-content", produces = MediaType.APPLICATION_JSON_VALUE)
   @ApiOperation("Translate monitor content for a specific agent type and version")
-  public String translateMonitorContent(@RequestBody @Validated TranslateMonitorContentRequest request)
+  public String translateMonitorContent(@RequestBody @Validated(ValidationGroups.TranslateMonitorContent.class) TranslateMonitorContentRequest request)
       throws MonitorContentTranslationException {
     final Map<AgentType, List<MonitorTranslationOperator>> operatorsByType =
         monitorContentTranslationService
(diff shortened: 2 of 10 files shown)


src/main/java/com/rackspace/salus/monitor_management/web/model/ValidationGroups.java
```diff
@@ -33,5 +33,6 @@
   public interface Create extends Default { }
   public interface Update extends Default { }
   public interface Patch extends Default { }
+  public interface TranslateMonitorContent extends Default { }
```
jjbuchan, line 36: I'm a little torn between using a validation group for this vs. creating a new class with only the fields we care about.  

In either case I'm struggling to come up with good names.  If we use the group I think it should be a more specific operation like `TranslationWithoutMonitorProvided` since both uses of `TranslateMonitorContentRequest` are to translate monitor content. We'd want to be able to easily distinguish between the use-cases.

```suggestion
  public interface TranslationWithoutMonitorProvided extends Default { }
```

src/main/java/com/rackspace/salus/monitor_management/web/model/TranslateMonitorContentRequest.java
```diff
@@ -22,17 +22,19 @@
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
 import lombok.Data;
+import org.springframework.validation.annotation.Validated;
 
 @Data
+@Validated
 public class TranslateMonitorContentRequest {
-  @NotBlank
+  @NotBlank(groups = {ValidationGroups.TranslateMonitorContent.class})
```
itzg, line 30: With my comment about creating a new class for the new request, that'll simplify away the group specifics here.

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
```diff
@@ -373,4 +374,29 @@ public void delete(@PathVariable String tenantId,
   public void deleteAllTenantMonitors(@PathVariable String tenantId, @RequestParam(defaultValue = "true") boolean sendEvents) {
     monitorManagement.removeAllTenantMonitors(tenantId, sendEvents);
   }
+
+  @PostMapping(value = "/tenant/{tenantId}/monitors/{monitorId}/agent-config", produces = MediaType.APPLICATION_JSON_VALUE)
+  @ApiOperation("Translate monitor content for a specific agent type and version")
+  public String getAgentConfig(@PathVariable String tenantId, @PathVariable UUID monitorId,
+      @RequestBody @Validated TranslateMonitorContentRequest request)
```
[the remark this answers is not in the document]
jjbuchan: I'm +1 to using a new class.  I was leaning that way but couldn't convince myself it was definitely the best option.

jjbuchan:
> whether the template variable rendering

How about allowing an optional `resourceId` to be provided in the payload?  It would then handle both cases.

itzg:
> > whether the template variable rendering
> 
> How about allowing an optional `resourceId` to be provided in the payload? It would then handle both cases.

I like that idea.

@rohitsngh27 , to avoid adding even more business logic and dependencies into the `MonitorApiController`, I would suggest adding a new public method to `MonitorManagement` that is something like `renderMonitorTemplate(UUID monitorId, String resourceId)` that returns a new class called `RenderedMonitorTemplate`. It could take care of retrieving the monitor and optionally retrieving the resource and doing template rendering. The returned `RenderedMonitorTemplate` would just contain `Monitor monitor` and `String renderedContent`. If the `resourceId` was null, then `renderedContent` would just be the `content` of the `Monitor`.

(Hopefully that all made sense...got more complicated the longer I typed.)

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2288,4 +2289,36 @@ public void removeAllTenantMonitors(String tenantId, boolean sendEvents) {
       monitorRepository.deleteAllByTenantId(tenantId);
     }
   }
+
+  public RenderedMonitorTemplate renderMonitorTemplate(UUID monitorId, String resourceId,
+      String tenantId) {
+    //Get existing monitor
+    Monitor monitor = getMonitor(tenantId, monitorId).orElseThrow(
+        () -> new NotFoundException(String.format("No policy monitor found for %s", monitorId)));
```
itzg, line 2297: Sorry, I missed this typo before
```suggestion
        () -> new NotFoundException(String.format("No monitor found for %s", monitorId)));
```

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2288,4 +2289,36 @@ public void removeAllTenantMonitors(String tenantId, boolean sendEvents) {
...
+
+    if (resource != null) {
+      try {
+        final String renderedContent = getRenderedContent(monitor.getContent(), resource);
+        renderedMonitorTemplate.setRenderedContent(renderedContent);
+      } catch (InvalidTemplateException e) {
+        log.warn("Unable to render content='{}' of monitor={} for resource={}",
+            monitor.getContent(), monitor, resource, e);
+        invalidTemplateErrors.increment();
+      }
```
itzg, line 2317: In this context, I'm thinking this should be wrapped and thrown as an `IllegalArgumentException`. The other call location could re-throw since I think it was within an kafka listener.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2288,4 +2289,36 @@ public void removeAllTenantMonitors(String tenantId, boolean sendEvents) {
...
+    } else {
+      log.warn("Failed to find resourceId={} during processing of monitor={}",
+          resourceId, monitor);
```
itzg, line 2320: ...and similarly this probably needs to throw an exception. Can't decide on `NotFoundException` or `IllegalArgumentException`.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4454,4 +4457,92 @@ public void testPatchExistingMonitor_nonExistingResource() {
...
+        .setTenantId(tenantId);
+
+    when(resourceApi.getByResourceId(tenantId, resourceId))
+        .thenReturn(r1);
+
+    final RenderedMonitorTemplate renderedMonitorTemplate = monitorManagement
+        .renderMonitorTemplate(monitor.getId(), resourceId, tenantId);
+
+    assertEquals(renderedMonitorTemplate.getMonitor().getId(), monitor.getId());
+    assertNotNull(renderedMonitorTemplate.getRenderedContent());
```
itzg, line 4493: The actual rendered content should be asserted. I would suggest borrowing some input ideas from a unit test like `MonitorContentRendererTest#testRenderTypical`

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4454,4 +4457,92 @@ public void testPatchExistingMonitor_nonExistingResource() {
...
+
+    String tenantId = RandomStringUtils.randomAlphanumeric(10);
+    String resourceId = RandomStringUtils.randomAlphanumeric(10);
+    UUID monitorId = UUID.randomUUID();
+
+    monitorManagement.renderMonitorTemplate(monitorId, resourceId, tenantId);
+  }
+
+  @Test
+  public void testRenderedMonitorTemplate_ResourceNotPresent() throws IOException {
```
itzg, line 4509: Just noting that this unit test will change if going with an exception for resource not found.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4454,4 +4457,92 @@ public void testPatchExistingMonitor_nonExistingResource() {
...
+        .setLabels(Collections.singletonMap("os", "linux"))
+        .setMetadata(new HashMap<>())
+        .setResourceId(resourceId)
+        .setTenantId(tenantId);
+
+    when(resourceApi.getByResourceId(tenantId, resourceId))
+        .thenReturn(r1);
+
+    final RenderedMonitorTemplate renderedMonitorTemplate = monitorManagement
+        .renderMonitorTemplate(monitor.getId(), resourceId, tenantId);
```
itzg, line 4490: Please also add a positive unit test where `resourceId` is null. In that case you'd assert the "rendered" content is the same as the monitor's.

rohitsngh27: I have added that test cases in MonitorManagementTest:- testRenderedMonitorTemplate_ResourceNotPresent

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
rohitsngh27: Should I also then change the log from warn to error if we are throwing an exception?
What should be the error message in that case?

itzg: Good question. It probably should be an error level log then. The message itself still seems ok.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2311,13 +2311,18 @@ public RenderedMonitorTemplate renderMonitorTemplate(UUID monitorId, String reso
         final String renderedContent = getRenderedContent(monitor.getContent(), resource);
         renderedMonitorTemplate.setRenderedContent(renderedContent);
       } catch (InvalidTemplateException e) {
-        log.warn("Unable to render content='{}' of monitor={} for resource={}",
+        log.error("Unable to render content='{}' of monitor={} for resource={}",
             monitor.getContent(), monitor, resource, e);
-        invalidTemplateErrors.increment();
+        throw new IllegalArgumentException(
+            String.format("Unable to render content=%s of monitor=%s for resource=%s",
+                monitor.getContent(), monitor, resource));
```
itzg, line 2318: The whole monitor in the exception message will be too much, so I'd suggest just:
```suggestion
            String.format("Unable to render content=%s for resource=%s",
                monitor.getContent(), resource));
```

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4486,11 +4486,17 @@ public void testRenderedMonitorTemplate() throws IOException {
     when(resourceApi.getByResourceId(tenantId, resourceId))
         .thenReturn(r1);
 
+    final MonitorContentProperties properties = new MonitorContentProperties();
+    final MonitorContentRenderer renderer = new MonitorContentRenderer(properties);
```
itzg, line 4490: When `new`ing a service/component that's usually a sign Spring is being circumvented. In this case, `MonitorContentRenderer` is already autowired via the `@Import` on the test class so you shouldn't manually create one.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4454,4 +4457,147 @@ public void testPatchExistingMonitor_nonExistingResource() {
     verifyNoMoreInteractions(monitorEventProducer);
   }
 
+  @Test
+  public void testRenderedMonitorTemplate() throws IOException, InvalidTemplateException {
+
+    String tenantId = RandomStringUtils.randomAlphanumeric(10);
+    String resourceId = RandomStringUtils.randomAlphanumeric(10);
+
+    String monitorContent = readContent("/MonitorConversionServiceTest_ping_with_policy.json");
```
itzg, line 4466: What I meant more specifically was that you should just use some literal content like what you did in the failed template case with

```
    String monitorContent = "value=${resource.wrong.reference}";
```

but instead something valid like
```
    String monitorContent = "value=${resource.labels.os}";
```

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4454,4 +4457,147 @@ public void testPatchExistingMonitor_nonExistingResource() {
...
+        .setTenantId(tenantId)
+        .setResourceId(null)
+        .setSelectorScope(ConfigSelectorScope.REMOTE)
+        .setLabelSelectorMethod(LabelSelectorMethod.AND)
+        .setZones(Collections.emptyList())
+        .setInterval(Duration.ofSeconds(60));
+    entityManager.persist(monitor);
+
+    final ResourceDTO r1 = new ResourceDTO()
+        .setLabels(Collections.singletonMap("os", "linux"))
```
itzg, line 4481: ...where the template is referencing this

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4454,4 +4457,147 @@ public void testPatchExistingMonitor_nonExistingResource() {
...
+    final MonitorContentProperties properties = new MonitorContentProperties();
+    final MonitorContentRenderer renderer = new MonitorContentRenderer(properties);
+
+    final String rendered = renderer.render(monitor.getContent(), r1);
+
+    final RenderedMonitorTemplate renderedMonitorTemplate = monitorManagement
+        .renderMonitorTemplate(monitor.getId(), resourceId, tenantId);
+
+    assertEquals(renderedMonitorTemplate.getMonitor().getId(), monitor.getId());
+    assertNotNull(renderedMonitorTemplate.getRenderedContent());
```
itzg, line 4498: ...and rendered content can be assert to ensure `${resource.labels.os}` was replaced with `linux`

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4486,17 +4489,15 @@ public void testRenderedMonitorTemplate() throws IOException, InvalidTemplateExc
     when(resourceApi.getByResourceId(tenantId, resourceId))
         .thenReturn(r1);
 
-    final MonitorContentProperties properties = new MonitorContentProperties();
-    final MonitorContentRenderer renderer = new MonitorContentRenderer(properties);
-
-    final String rendered = renderer.render(monitor.getContent(), r1);
+    final String rendered = monitorContentRenderer.render(monitor.getContent(), r1);
```
itzg, line 4492: I think you can now remove this and line 4499 below.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2288,4 +2289,41 @@ public void removeAllTenantMonitors(String tenantId, boolean sendEvents) {
...
+
+    //Set Render monitor template
+    RenderedMonitorTemplate renderedMonitorTemplate = new RenderedMonitorTemplate();
+    renderedMonitorTemplate.setMonitor(monitor);
+    renderedMonitorTemplate.setRenderedContent(monitor.getContent());
+
+    if (StringUtils.isBlank(resourceId)) {
+      return renderedMonitorTemplate;
+    }
+    ResourceDTO resource = resourceApi.getByResourceId(tenantId, resourceId);
```
jjbuchan, line 2307: This could be a db request instead of api.  You'd get back a `Resource` but can still easily convert it to a DTO if needed.

I don't think we cleaned up all these api requests once we decided to use the shared db.  That could be a quick issue for you to work on in another PR once this is merged - convert any basic api lookups to db lookups instead.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2288,4 +2289,41 @@ public void removeAllTenantMonitors(String tenantId, boolean sendEvents) {
...
+        renderedMonitorTemplate.setRenderedContent(renderedContent);
+      } catch (InvalidTemplateException e) {
+        log.error("Unable to render content='{}' of monitor={} for resource={}",
+            monitor.getContent(), monitor, resource, e);
+        throw new IllegalArgumentException(
+            String.format("Unable to render content=%s for resource=%s",
+                monitor.getContent(), resource));
+      }
+    } else {
+      log.error("Failed to find resourceId={} during processing of monitor={}",
```
jjbuchan, line 2321: I don't think this should be an `error`.  i.e. we don't want errors showing up in our logs when it might just have been an invalid value provided by the customer and not a failure in the system.

I'd be good with either `warn` or `info` but am slightly leaning towards `info`.

jjbuchan: Since I see `processMonitorContentModified` has the same log message but as `warn` I'd recommend this being `info` and potentially even changing the message a little to make it clearer that it's from this method and not the other.

Maybe `Invalid resourceId={} provided when rendering monitorId={} template for tenantId={}`.

Based on recent discussions I'm unclear whether logging objects will be parsed correctly in our logs which is why I suggest the monitorId instead of the full monitor object.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4454,4 +4457,142 @@ public void testPatchExistingMonitor_nonExistingResource() {
...
+        .renderMonitorTemplate(monitor.getId(), resourceId, tenantId);
+
+    assertEquals(renderedMonitorTemplate.getMonitor().getId(), monitor.getId());
+    assertNotNull(renderedMonitorTemplate.getRenderedContent());
+    assertEquals(renderedMonitorTemplate.getRenderedContent(), "value=linux");
+
+    verify(resourceApi).getByResourceId(tenantId, resourceId);
+  }
+
+  @Test(expected = NotFoundException.class)
```
jjbuchan, line 4499: I'd recommend just using `@Test` here and instead wrapping the render request in `assertThatThrownBy(() -> {`

I think the `expected` format does not always catch problems / by not asserting on the error status it can hide issues.

Here's an example copied from elsewhere similar to what I'd expect

```
    assertThatThrownBy(() -> monitorMetadataPolicyManagement.removeZonePolicy(region))
        .isInstanceOf(NotFoundException.class)
        .hasMessage(
            String.format("No zone policy found for region %s", region)
        );
```

jjbuchan: Same for all the other exception tests too.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -4454,4 +4457,142 @@ public void testPatchExistingMonitor_nonExistingResource() {
...
+        .setAgentType(AgentType.TELEGRAF)
+        .setMonitorType(MonitorType.ping)
+        .setContent(monitorContent)
+        .setTenantId(tenantId)
+        .setResourceId(null)
+        .setSelectorScope(ConfigSelectorScope.REMOTE)
+        .setLabelSelectorMethod(LabelSelectorMethod.AND)
+        .setZones(Collections.emptyList())
+        .setInterval(Duration.ofSeconds(60));
+    entityManager.persist(monitor);
```
jjbuchan, line 4478: @itzg what do you think about changing these to `monitorRepository.save()` instead?

I'd like to limit the use of entity manager where possible, and then when we know we need to use it we can document why it is needed.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
itzg: As I'm refactoring this zone monitor binding logic I'm noticing that `MonitorManagementTest` still uses the JUnit `assertThat`. Therefore, there's only a couple AssertJ patterns in there so far.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
itzg: Agreed. I'm switching some of those as I refactor the zone monitor binding.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2318,11 +2318,11 @@ public RenderedMonitorTemplate renderMonitorTemplate(UUID monitorId, String reso
                 monitor.getContent(), resource));
       }
     } else {
-      log.error("Failed to find resourceId={} during processing of monitor={}",
-          resourceId, monitor);
+      log.info("Invalid resourceId={} provided when rendering monitorId={} template for tenantId={}",
```
itzg, line 2321: I think it was good as an error
```suggestion
      log.error("Invalid resourceId={} provided when rendering monitorId={} template for tenantId={}",
```

jjbuchan: @itzg I suggested it be `info`.  It's an error to the customer making a request but not to us.  The system processed things as expected.

itzg: Sorry, I missed that and that makes sense.

PR #202, opened 2020-08-31
updated config from tools update and fixed up cloud upload

harageth:
# Resolves

https://jira.rax.io/browse/SALUS-990

# What

Make sure that we are generating the documentation in a way that google cloud storage will actually display the HTML.

# How

Simplifies the generation process by removing the mustache templates, generates the html with a different tool that isn't downloading js from some other site and uploads each of the public and admin html files on just a deploy. 

# How to test

Delete your entire targets directory then run the maven `package` command (after having updated and `mvn deploy`ed the new tools changes [linked at the bottom])

# Why

The biggest problem from this side was that we were generating multiple swagger html documents that needed to be uploaded and the tools used for that wouldn't generate each one with different names. Because of that the artifacts approach in the cloudbuilds file wouldn't work because it wanted to upload every file to one directory and uploading multiple files of the same name just overwrote that file. 

Ultimately using gsutils allowed us to upload each file to its own location without having to worry about what the filename was in the first place. 

# Depends on

https://github.com/racker/salus-tools/pull/41


pom.xml
```diff
@@ -186,11 +187,6 @@
       <artifactId>spring-kafka-test</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-starter-test</artifactId>
-      <scope>test</scope>
-    </dependency>
```
jjbuchan, line 193: Why?

jjbuchan: Nevermind, I see it's there twice.

harageth: Yep and it weirdly caused an error for me so I decided to remove it.

PR #203, opened 2020-09-01
Refactored zone unit tests in order to use non-mock BoundMonitorRepository

itzg:
# Prep work for

https://jira.rax.io/browse/SALUS-358

# What

The zone binding changes I'll make will involve using new named queries on the `BoundMonitor` entity, so the zone assignment related unit tests really need to be using a non-mock `BoundMonitorRepository`

# How

Moved almost all of the unit test methods that were zone related out of `MonitorManagementTest` and into a new test class that `@Autowire`s the `BoundMonitorRepository` to be backed by testcontainer's mysql.

The biggest change was replacing when/verify on what was the mock repository, it instead does retrieval of the bound monitors by monitor ID and asserts the expected results. I introduced a custom hamcrest matcher `BoundMonitorMatcher` to simplify the assertion declarations and avoid all the awkward things like not being able to directly assert JPA-bags to lists, skipping timestamps, etc.

In the process I did some cleanup of the unit tests, such as using randomized tenantIds and better utility method consolidation to improve readability.

# How to test

All the unit tests should still pass.


src/test/java/com/rackspace/salus/monitor_management/services/BoundMonitorMatcher.java
```diff
@@ -0,0 +1,53 @@
...
+  final String resourceId;
+  final String envoyId;
+
+  @Override
+  public boolean matches(Object o) {
+    if (!(o instanceof BoundMonitor)) {
+      return false;
+    }
+    final BoundMonitor boundMonitor = (BoundMonitor) o;
+    return Objects.equals(boundMonitor.getTenantId(), tenantId) &&
```
GeorgeJahad, line 39: does this need to compare the monitor id?

itzg: It probably should for completeness. I hadn't yet since it was only used within the `assertBindings` context so far:
```
  private void assertBindings(UUID monitorId, BoundMonitorMatcher... matchers) {
    final List<BoundMonitor> results = boundMonitorRepository
        .findAllByMonitor_Id(monitorId);

    assertThat(results, hasSize(matchers.length));

    assertThat(results, containsInAnyOrder(matchers));
  }
```

GeorgeJahad: I would either add it, or add a comment saying it was intentionally left out, because otherwise people will wonder if it is a bug.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -1761,68 +1564,6 @@ public void testRemoveMonitor() {
 
     verify(boundMonitorRepository).deleteAll(Collections.singletonList(boundMonitor));
 
-    verify(zoneStorage).decrementBoundCount(
-        ResolvedZone.createPrivateZone("t-1", "z-1"),
-        "r-gone"
-    );
-    verify(zoneStorage).getEnvoyIdToResourceIdMap(ResolvedZone.createPrivateZone("t-1", "z-1"));
```
GeorgeJahad, line 1768: I understand removing the decrementBoundCount() but why getEnvoyIdToResourceIdMap() and sendMonitorEvent()

GeorgeJahad: ah looks like maybe sendMonitorEvent() is still there and the diff is just wierd

itzg: I had already forgotten about this. In the process of refactoring out the zone related tests, I realized that left us without a `testRemoveMonitor` test case for local monitors. So I repurposed this one for local monitors and as such the zoneStorage operations were no longer applicable.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -2340,526 +2081,42 @@ public void testDistributeNewMonitor_agent() {
...
-            .setResourceId("r-3")
-    );
-
-    // but only one envoy is available
-    Queue<EnvoyResourcePair> availableEnvoys = new LinkedList<>();
-    availableEnvoys.add(new EnvoyResourcePair().setEnvoyId("e-1").setResourceId("r-e-1"));
-    // ...same envoy again to verify de-duping
-    availableEnvoys.add(new EnvoyResourcePair().setEnvoyId("e-1").setResourceId("r-e-1"));
-
-    when(zoneStorage.findLeastLoadedEnvoy(any()))
```
GeorgeJahad, line 2495: i can't quite tell what this block of code was doing but there appears to be nothing analogous in the new version of the code.  is it not needed or done somewhere else now?

itzg: Honestly, I couldn't remember either why the original unit test bothered to simulate that the "e-1" envoy was reported as available the first two times and not on the third. We don't any allocate limits on envoy-pollers, so the behavior didn't seem realistic. I decided to remove that confusing, complexity when bringing the test over.

PR #205, opened 2020-09-03
Calculate and maintain zone binding counts in SQL

itzg:
# Resolves

https://jira.rax.io/browse/SALUS-358

# What

Rather than use a brittle, DIY counter in etcd, we'll instead use etcd to simply retrieve a list of the active poller's per remote monitoring zone and use the `BoundMonitor` table as the single source of truth for bindings per envoy-poller.

The motivation to make this change was observing my local system having troubles binding a new monitor to the 1000 resources I had left over from envoy-ambassador stress testing. The `ZoneStorage` code was getting itself into a tight loop of operations like the following with numerous collisions trying to increment the binding count:

```
2020-08-28 13:22:10.821 DEBUG [salus-telemetry-monitor-management,,,] 78771 --- [         etcd22] c.r.s.t.etcd.services.ZoneStorage        : Putting new bound count=894 for resource=development:0 in zone=ResolvedZone(name=dev, tenantId=aaaaaa)
2020-08-28 13:22:10.821 DEBUG [salus-telemetry-monitor-management,,,] 78771 --- [         etcd27] c.r.s.t.etcd.services.ZoneStorage        : Changing bound count of resource=development:0 in zone=ResolvedZone(name=dev, tenantId=aaaaaa) by amount=-1
2020-08-28 13:22:10.827 DEBUG [salus-telemetry-monitor-management,,,] 78771 --- [         etcd30] c.r.s.t.etcd.services.ZoneStorage        : Re-trying incrementing bound count of resource=development:0 in zone=ResolvedZone(name=dev, tenantId=aaaaaa) due to collision
```

# How

[The first commit](https://github.com/racker/salus-telemetry-monitor-management/commit/965b6cace8a3923034b6ae8cd0b5087b74410d47) only a refactoring to consolidate onto a single method of `MonitorManagement` for locating a least-loaded envoy-poller.

The second, big commit includes a new `ZoneAllocationResolver` component, which is allocated and re-used before each iterative operation done in monitor management. For example, if a monitor is about to get bound to 10,000 tenant resources, then a single instance of `ZoneAllocationResolver` is utilized for that. Each subsequent operation gets a fresh instance of that component though.

As you'll see in the javadoc, the `ZoneAllocationResolver` uses a minimal, cheap query into etcd to zero out the known, active envoy-pollers. Then it uses the new named queries on `BoundMonitor` to retrieve an aggregate count of bound monitors per envoy-poller resourceid for a requested zone.

The solution should scale to handle large numbers of tenant resources. The etcd query is small and constant. The `BoundMonitor` monitor table becomes the single source of truth and the remaining "expensive" operation is done within SQL, optimized by new indices, and the snapshot of those counts is used in-memory for any size of monitor-resource-zone binding operation.

# How to test

Updated existing unit tests of remote monitor binding to confirm the `pollerResourceId` was set as expected.

Added a new unit class for the new `ZoneAllocationResolver`.

# Depends on

- https://github.com/racker/salus-telemetry-model/pull/151
- https://github.com/racker/salus-telemetry-etcd-adapter/pull/67

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -2042,14 +2042,16 @@ public class MonitorManagement {
    */
   @SuppressWarnings("WeakerAccess")
   public void handleExpiredEnvoy(@Nullable String zoneTenantId, String zoneName, String envoyId) {
-    log.debug("Reassigning bound monitors for disconnected envoy={} with zoneName={} and zoneTenantId={}",
-        envoyId, zoneName, zoneTenantId);
     List<BoundMonitor> boundMonitors = getAllBoundMonitorsByEnvoyId(envoyId);
     if (boundMonitors.isEmpty()) {
       return;
     }
+    log.info("Reassigning bound monitors for disconnected envoy={} with zoneName={} and zoneTenantId={}",
+        envoyId, zoneName, zoneTenantId);
     for (BoundMonitor boundMonitor : boundMonitors) {
-      boundMonitor.setEnvoyId(null);
+      boundMonitor
+          .setEnvoyId(null)
+          .setPollerResourceId(null);
     }
 
     saveBoundMonitors(boundMonitors);
@@ -2083,7 +2085,7 @@ public class MonitorManagement {
   public CompletableFuture<Integer> rebalanceZone(@Nullable String zoneTenantId, String zoneName) {
     final ResolvedZone zone = resolveZone(zoneTenantId, zoneName);
 
-    log.debug("Rebalancing zone={}", zone);
+    log.info("Rebalancing zone={}", zone);
 
     return zoneAllocationResolverFactory.create()
         .getZoneBindingCounts(zone)
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -19,7 +19,6 @@ package com.rackspace.salus.monitor_management.services;
 import static com.google.common.collect.Collections2.transform;
 import static com.rackspace.salus.telemetry.entities.Monitor.POLICY_TENANT;
 import static com.rackspace.salus.telemetry.entities.Resource.REGION_METADATA;
-import static com.rackspace.salus.telemetry.etcd.types.ResolvedZone.createPrivateZone;
 import static com.rackspace.salus.telemetry.etcd.types.ResolvedZone.resolveZone;
 import static org.springframework.util.CollectionUtils.isEmpty;
 
@@ -49,7 +48,6 @@ import com.rackspace.salus.telemetry.entities.Resource;
 import com.rackspace.salus.telemetry.entities.Zone;
 import com.rackspace.salus.telemetry.errors.AlreadyExistsException;
 import com.rackspace.salus.telemetry.etcd.services.EnvoyResourceManagement;
-import com.rackspace.salus.telemetry.etcd.services.ZoneStorage;
 import com.rackspace.salus.telemetry.etcd.types.EnvoyResourcePair;
 import com.rackspace.salus.telemetry.etcd.types.ResolvedZone;
 import com.rackspace.salus.telemetry.messaging.MetadataPolicyEvent;
@@ -107,7 +105,6 @@ import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.stereotype.Service;
-import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
@@ -120,11 +117,11 @@ public class MonitorManagement {
   private final ResourceRepository resourceRepository;
   private final MonitorPolicyRepository monitorPolicyRepository;
   private final BoundMonitorRepository boundMonitorRepository;
-  private final ZoneStorage zoneStorage;
   private final MonitorEventProducer monitorEventProducer;
   private final MonitorContentRenderer monitorContentRenderer;
   private final PolicyApi policyApi;
   private final ResourceApi resourceApi;
+  private final ZoneAllocationResolverFactory zoneAllocationResolverFactory;
   private final ZoneManagement zoneManagement;
   private final ZonesProperties zonesProperties;
   private final String labelMatchQuery;
@@ -156,11 +153,12 @@ public class MonitorManagement {
       MonitorRepository monitorRepository, EntityManager entityManager,
       EnvoyResourceManagement envoyResourceManagement,
       BoundMonitorRepository boundMonitorRepository,
-      ZoneStorage zoneStorage,
       MonitorEventProducer monitorEventProducer,
       MonitorContentRenderer monitorContentRenderer,
       PolicyApi policyApi, ResourceApi resourceApi,
-      ZoneManagement zoneManagement, ZonesProperties zonesProperties,
+      ZoneAllocationResolverFactory zoneAllocationResolverFactory,
+      ZoneManagement zoneManagement,
+      ZonesProperties zonesProperties,
       MonitorConversionService monitorConversionService,
       MetadataUtils metadataUtils, MeterRegistry meterRegistry,
       JdbcTemplate jdbcTemplate)
@@ -171,11 +169,11 @@ public class MonitorManagement {
     this.entityManager = entityManager;
     this.envoyResourceManagement = envoyResourceManagement;
     this.boundMonitorRepository = boundMonitorRepository;
-    this.zoneStorage = zoneStorage;
     this.monitorEventProducer = monitorEventProducer;
     this.monitorContentRenderer = monitorContentRenderer;
     this.policyApi = policyApi;
     this.resourceApi = resourceApi;
+    this.zoneAllocationResolverFactory = zoneAllocationResolverFactory;
     this.zoneManagement = zoneManagement;
     this.zonesProperties = zonesProperties;
     this.monitorConversionService = monitorConversionService;
@@ -197,14 +195,6 @@ public class MonitorManagement {
         "operation", "handleMonitorPolicyEvent", "reason", "orphanedPolicyMonitor");
     policyIntegrityErrors = meterRegistry.counter("policy_integrity",
         "operation", "handleMonitorPolicyEvent", "reason", "tooManyClones");
-  }
-
-  /**
-   * FOR UNIT TESTING, provides the zone properties
-   * @return the {@link ZonesProperties} used by this service
-   */
-  ZonesProperties getZonesProperties() {
-    return zonesProperties;
   }
 
   /**
@@ -525,6 +515,8 @@ public class MonitorManagement {
     } else {
       // REMOTE MONITOR
 
+      final ZoneAllocationResolver zoneAllocationResolver = zoneAllocationResolverFactory.create();
+
       for (ResourceDTO resource : resources) {
         List<String> zonesForResource = zones;
 
@@ -535,7 +527,7 @@ public class MonitorManagement {
         for (String zone : zonesForResource) {
           try {
             boundMonitors.add(
-                bindRemoteMonitor(monitor, resource, zone)
+                bindRemoteMonitor(monitor, resource, zone, zoneAllocationResolver)
             );
           } catch (InvalidTemplateException e) {
             log.warn("Unable to render monitor={} onto resource={}",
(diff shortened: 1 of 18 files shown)


GeorgeJahad:
In your comments on pr creation you mention a couple of refactorings.  Was that to point to some commits that can be safely ignored, or should we be looking at all the commits here?

itzg:
> In your comments on pr creation you mention a couple of refactorings. Was that to point to some commits that can be safely ignored, or should we be looking at all the commits here?

Looking at all of the commits will be good. I just wanted to clarify that the first commit was a refactoring that didn't change functionality.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementExcludeResourceIdsTest.java
```diff
@@ -121,6 +122,9 @@
   @MockBean
   ResourceApi resourceApi;
 
+  @MockBean
+  ZoneAllocationResolver zoneAllocationResolver;
```
GeorgeJahad, line 126: This doesn't seem to get used in this class.  Is it required?

itzg: It's not exercised by the unit tests, but I was thinking the `MonitorManagement` bean dependencies would require it; however, I commented that out and the service initialized fine, but apparently there is a usage that was quietly getting mocked:

```
org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'com.rackspace.salus.monitor_management.services.ZoneAllocationResolver' available

	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBean(DefaultListableBeanFactory.java:352)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.getBean(DefaultListableBeanFactory.java:343)
	at org.springframework.beans.factory.support.CglibSubclassingInstantiationStrategy$LookupOverrideMethodInterceptor.intercept(CglibSubclassingInstantiationStrategy.java:246)
	at com.rackspace.salus.monitor_management.services.ZoneAllocationResolverFactory$$EnhancerBySpringCGLIB$$5bf6f0f8.create(<generated>)
	at com.rackspace.salus.monitor_management.services.MonitorManagement.upsertBindingToResource(MonitorManagement.java:1775)
```

Now I'm wondering if some assert/verify's are needed.

itzg: ...oh, now I see what's going on with that unit test. It's only using local monitors; however, the `zoneAllocationResolver` gets created prior to the loop since it's not known yet if remote monitors are involved:

https://github.com/racker/salus-telemetry-monitor-management/blob/4ea57c183999ed0587221e350e350f32586fbc37/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java#L1775-L1784

So, yes, it's needed but for a different reason than I originally anticipated.

GeorgeJahad: Interesting, I'm guessing it is related to the @Lookup annotation:
``` 
 @Lookup
  public abstract ZoneAllocationResolver create();
```

which I've never seen used before.

itzg: Yes, it's pretty rare that prototype-scoped beans are needed, so that `@Lookup` feature was new to me too:

https://docs.spring.io/spring/docs/current/spring-framework-reference/core.html#beans-factory-lookup-method-injection

itzg: ...the `@Lookup` method probably should have been added to `MonitorManagement` itself, but didn't like the feeling of flipping that into an abstract class.

GeorgeJahad: hmm,  something more for me to learn about

src/main/java/com/rackspace/salus/monitor_management/services/ZoneAllocationResolver.java
```diff
@@ -0,0 +1,176 @@
...
+    return getPollerResourceCounts(zone)
+        .thenCompose(pollerResourceCounts ->
+            // ...and get a mapping of poller resourceIds to current envoyIds
+            getResourceIdToEnvoyIdMap(zone)
+                .thenApply(resourceIdToEnvoyId ->
+                    pollerResourceCounts.entrySet().stream()
+                        // Pick the smallest, least-assigned entry
+                        .min(Comparator.comparingLong(Entry::getValue))
+                        .map(entry -> {
+                          // In our cache, count a binding to that poller-envoy
```
GeorgeJahad, line 77: So to be clear this comment means that resource count is being incremented on the assumption the the caller of findLeastLoadedEnvoy() is going to create a binding to that envoy?

src/main/java/com/rackspace/salus/monitor_management/services/ZoneAllocationResolver.java
```diff
@@ -0,0 +1,176 @@
...
+  private CompletableFuture<Map<String, Integer>> getPollerResourceCounts(ResolvedZone zone) {
+    return pollerCountCache.computeIfAbsent(
+        zone,
+        this::retrievePollerResourceCounts
+    );
+  }
+
+  /**
+   * Combines tracking of active poller-envoys with the currently bound monitors into a count of
+   * of bound monitors per poller-envoy
```
GeorgeJahad, line 100: This comment seems unclear to me.

Is this not clearer:
"Returns map of active poller-envoys to number of monitors bound to each"

The way it reads now sounds to me like it is doing more than just creating a map.  Or am I misunderstanding something?

itzg: I like your wording better. I might still re-use the other as a follow-up, implementation note.

src/main/java/com/rackspace/salus/monitor_management/services/ZoneAllocationResolver.java
itzg: I like the general idea of the parameter...however, I can't really think of a reason to call `findLeastLoadedEnvoy` with that behavior disabled. In fact, it actually seems more system impacting to not increment the tracking since repeated calls to this method would keep giving the same answer, potentially creating a large imbalance towards the one envoy-poller.

src/main/java/com/rackspace/salus/monitor_management/services/ZoneAllocationResolver.java
```diff
@@ -0,0 +1,176 @@
...
+ * This stateful component encapsulates and caches the resolving of poller-envoys to allocate for
+ * monitor binding. It is recommended to use {@link ZoneAllocationResolverFactory} to create
+ * instances of this component prior to iterating over monitor binding changes.
+ * <p>
+ *   The caching optimizes for the common scenario where a remote monitor is binding to a potentially
+ *   large number of resources. Rather than query etcd and SQL for the poller-envoy loads repeatedly,
+ *   a snapshot of each zone is retrieved as needed and tracks across repeated use.
+ *   The downside of using mutable snapshots is that concurrent use of these resolvers across the
+ *   system can result in potential imbalance in poller-envoy allocations; however, some amount of
+ *   imbalance is fine and can be rectified by a rebalance operation.
```
GeorgeJahad, line 46: So even though there will be multiple monitoring management nodes storing their own copy of this cache, the entries do not expire.

Is the belief that on average the ratio of load across the pollers will remain relatively accurate so that even though the actually counts will be off, findLeastLoadedEnvoy() will return a correct choice?

itzg: Perhaps the use of the term "cache" (and this caveat in this whole paragraph) was misleading. I should probably call those variables "snapshot". These resolver instances are thrown away at the end of each user/event operation and are re-calculated on the next. For example, two create monitor operations will each get a freshly computed instance of this resolver.

itzg:
@GeorgeJahad I changed the wording of the javadoc and fields to avoid implying that the `ZoneAllocationResolver` instances were meant for long term caching.

src/test/java/com/rackspace/salus/monitor_management/services/ZoneAllocationResolverTest.java
```diff
@@ -0,0 +1,412 @@
...
+                new BoundMonitor()
+                    .setMonitor(monitor)
+                    .setTenantId(monitor.getTenantId())
+                    .setEnvoyId(pollerEnvoyId)
+                    .setPollerResourceId(pollerResourceId)
+                    .setZoneName(zoneName)
+                    .setResourceId(randomAlphanumeric(5))
+            )
+        )
+        .collect(Collectors.toList());
```
GeorgeJahad, line 392: what happens if you remove this line?

GeorgeJahad (approved):

jjbuchan (commented):
Looks good!

src/main/java/com/rackspace/salus/monitor_management/services/TestMonitorService.java
```diff
@@ -194,15 +196,15 @@ private String resolveRemoteEnvoy(String tenantId,
           "test-monitor requires only one monitoring zone to be given");
     }
 
-    final String envoyId = monitorManagement
-        .findLeastLoadedEnvoyInZone(tenantId, monitoringZones.get(0));
+    final Optional<EnvoyResourcePair> leastLoaded = monitorManagement
+        .findLeastLoadedEnvoyInZone(resolveZone(tenantId, monitoringZones.get(0)));
 
-    if (envoyId == null) {
+    if (leastLoaded.isEmpty()) {
```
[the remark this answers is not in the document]
itzg: It sure is. Thanks.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
```diff
@@ -117,7 +116,10 @@
     MonitorContentProperties.class,
     MonitorConversionService.class,
     MetadataUtils.class,
-    DatabaseConfig.class})
+    DatabaseConfig.class,
+    SimpleMeterRegistry.class,
```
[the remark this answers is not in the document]
itzg: Agreed, I'm pretty sure this didn't work before, but for whatever reason it was happy with this now.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -139,7 +136,12 @@
     MonitorContentRenderer.class,
     MonitorContentProperties.class,
     MetadataUtils.class,
-    DatabaseConfig.class})
+    DatabaseConfig.class,
+    ZoneAllocationResolverFactory.class,
+    ZonesProperties.class,
+    SimpleMeterRegistry.class,
+
```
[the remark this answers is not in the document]
itzg: but I'll delete that empty line

src/main/java/com/rackspace/salus/monitor_management/services/ZoneAllocationResolver.java
```diff
@@ -0,0 +1,187 @@
...
+ *   potentially large number of resources.
+ *   For example, rather than repeatedly query etcd and SQL for every resource of a monitor
+ *   to be bound, a snapshot of each zone is retrieved as needed and locally incremented.
+ *   The downside of using a mutable snapshot is that each monitor operation can slightly
+ *   diverge in tracking poller-envoy allocations; however, some amount of imbalance is fine
+ *   and can be rectified by a rebalance operation.
+ * </p>
+ */
+@Component
+@Scope("prototype")
```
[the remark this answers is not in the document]
itzg: As mentioned earlier on PR, it's very rarely helpful but is nice when you do need a short-lived component with regular dependency injection.

src/main/java/com/rackspace/salus/monitor_management/services/ZoneAllocationResolver.java
```diff
@@ -0,0 +1,187 @@
...
+public class ZoneAllocationResolver {
+
+  private final ZoneStorage zoneStorage;
+  private final EntityManager entityManager;
+
+  private final Map<ResolvedZone, CompletableFuture<Map<String, String>>> resourceIdToEnvoyIdSnapshot = new HashMap<>();
+  private final Map<ResolvedZone, CompletableFuture<Map<String, Integer>>> pollerCountSnapshot = new HashMap<>();
```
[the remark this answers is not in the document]
itzg: It allowed for the snapshot resolving to pass through either an in-progress etcd lookup or the pre-completed future.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagement_ZoneBindingsTest.java
```diff
@@ -104,20 +106,27 @@
...
+      MonitorContentRenderer.class,
+      MonitorContentProperties.class,
+      MetadataUtils.class,
+      DatabaseConfig.class,
+      ServicesProperties.class,
+      ZonesProperties.class,
+      SimpleMeterRegistry.class,
+      ZoneAllocationResolverFactory.class,
+  })
+  public static class TestConfig { }
```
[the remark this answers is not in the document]
itzg: For some reason the components in etcd-adapter were also getting auto-registered and complaining about missing dependencies. I move these imports back to the usual spot and now it works. So, next commit it'll be consistent with others.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagement_ZoneBindingsTest.java
```diff
@@ -1266,4 +1175,11 @@ private BoundMonitor persistBoundMonitor(String resourceId, String zoneName,
     return counts;
   }
 
+  private void assertBindings(UUID monitorId, BoundMonitorMatcher... matchers) {
+    final List<BoundMonitor> results = boundMonitorRepository
+        .findAllByMonitor_Id(monitorId);
+
+    assertThat(results, containsInAnyOrder(matchers));
```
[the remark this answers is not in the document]
itzg: I since confirmed that  Hamcrest's`containsInAnyOrder` also asserts the length of the expected collection with the actual collection.

src/main/java/com/rackspace/salus/monitor_management/web/model/BoundMonitorDTO.java
```diff
@@ -54,6 +54,7 @@
   AgentType agentType;
   String renderedContent;
   String envoyId;
+  String pollerResourceId;
```
jjbuchan, line 57: Need to add `@JsonView(View.Admin.class)`.  We wouldn't want resourceIds exposed to customers (at least not for public pollers).

jjbuchan: Probably worth adding to `BoundMonitorDTOTest` like what we have in ZoneDTO

```
    objectAsString = objectMapper.writerWithView(View.Admin.class).writeValueAsString(dto);
    convertedDto = objectMapper.readValue(objectAsString, ZoneDTO.class);
    assertThat(convertedDto.getState(), notNullValue());
```
to verify this field is not returned for customers but is for us.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2042,14 +2042,16 @@ private void handleReattachedEnvoy(String tenantId, String resourceId, String en
    */
   @SuppressWarnings("WeakerAccess")
   public void handleExpiredEnvoy(@Nullable String zoneTenantId, String zoneName, String envoyId) {
-    log.debug("Reassigning bound monitors for disconnected envoy={} with zoneName={} and zoneTenantId={}",
-        envoyId, zoneName, zoneTenantId);
     List<BoundMonitor> boundMonitors = getAllBoundMonitorsByEnvoyId(envoyId);
     if (boundMonitors.isEmpty()) {
       return;
     }
+    log.info("Reassigning bound monitors for disconnected envoy={} with zoneName={} and zoneTenantId={}",
```
jjbuchan, line 2049: Now you've got this here I think adding in the count of boundMonitors being reassigned would be good

src/main/java/com/rackspace/salus/monitor_management/web/model/BoundMonitorDTO.java
itzg: Thank you, I forgot about the admin view that you mentioned at the start. Hmm, it would be nice to have it reveal the poller resource ID for their private zones though...

Speaking of which, I wonder `envoyId` should also be limited to admin view?

jjbuchan: > it would be nice
Agreed.  I figure this is a quick solution for now.

> if `envoyId` should also be limited 
I'm good with that.  Maybe `Internal` for it?

jjbuchan (approved):

PR #206, opened 2020-09-07
Converted resourceApi requests to DB requests

rohitsngh27:
# Resolves
https://jira.rax.io/browse/SALUS-1014

# What
Convert api requests to DB requests

# How
Changes in MonitorManagement class to use DB requests instead of API call.

# How to test
This can be tested via REST call

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -2279,14 +2279,8 @@ public class MonitorManagement {
 
   public ResourceDTO findResourceByTenantIdAndResourceId(String tenantId, String resourceId) {
     Resource resource = resourceRepository.findByTenantIdAndResourceId(tenantId, resourceId)
-        .orElseGet(() -> null);
+        .orElse(null);
 
-    if (resource == null) {
-      return null;
-    }
-    ResourceInfo resourceInfo = envoyResourceManagement
-        .getOne(resource.getTenantId(), resource.getResourceId()).join();
-    return new ResourceDTO(resource,
-        resourceInfo == null ? null : resourceInfo.getEnvoyId());
+    return resource == null ? null : new ResourceDTO(resource, null);
   }
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1153,7 +1153,7 @@ public class MonitorManagement {
     for (Entry<String, List<BoundMonitor>> resourceEntry : groupedByResourceId.entrySet()) {
 
       final String resourceId = resourceEntry.getKey();
-      final ResourceDTO resource = resourceApi.getByResourceId(tenantId, resourceId);
+      final ResourceDTO resource = findResourceByTenantIdAndResourceId(tenantId, resourceId);
 
       if (resource != null) {
         try {
@@ -1212,7 +1212,7 @@ public class MonitorManagement {
 
     // If a new one is to be bound, bind it
     if (StringUtils.isNotBlank(updatedResourceId)) {
-      ResourceDTO resource = resourceApi.getByResourceId(monitor.getTenantId(), updatedResourceId);
+      ResourceDTO resource = findResourceByTenantIdAndResourceId(monitor.getTenantId(), updatedResourceId);
       if (resource != null) {
         affectedEnvoys.addAll(
             upsertBindingToResource(
@@ -2254,7 +2254,7 @@ public class MonitorManagement {
     if (StringUtils.isBlank(resourceId)) {
       return renderedMonitorTemplate;
     }
-    ResourceDTO resource = resourceApi.getByResourceId(tenantId, resourceId);
+    ResourceDTO resource = findResourceByTenantIdAndResourceId(tenantId, resourceId);
 
     if (resource != null) {
       try {
@@ -2275,5 +2275,18 @@ public class MonitorManagement {
               resourceId, monitorId, tenantId));
     }
     return renderedMonitorTemplate;
+  }
+
+  public ResourceDTO findResourceByTenantIdAndResourceId(String tenantId, String resourceId) {
+    Resource resource = resourceRepository.findByTenantIdAndResourceId(tenantId, resourceId)
+        .orElseGet(() -> null);
+
+    if (resource == null) {
+      return null;
+    }
+    ResourceInfo resourceInfo = envoyResourceManagement
+        .getOne(resource.getTenantId(), resource.getResourceId()).join();
+    return new ResourceDTO(resource,
+        resourceInfo == null ? null : resourceInfo.getEnvoyId());
   }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2326,4 +2326,17 @@ public RenderedMonitorTemplate renderMonitorTemplate(UUID monitorId, String reso
     }
     return renderedMonitorTemplate;
   }
+
+  public ResourceDTO findResourceByTenantIdAndResourceId(String tenantId, String resourceId) {
+    Resource resource = resourceRepository.findByTenantIdAndResourceId(tenantId, resourceId)
+        .orElseGet(() -> null);
```
jjbuchan, line 2332: In this case you can just do

```suggestion
        .orElse(null);
```

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2326,4 +2326,17 @@ public RenderedMonitorTemplate renderMonitorTemplate(UUID monitorId, String reso
...
+
+    if (resource == null) {
+      return null;
+    }
+    ResourceInfo resourceInfo = envoyResourceManagement
+        .getOne(resource.getTenantId(), resource.getResourceId()).join();
```
jjbuchan, line 2338: After a quick look, I don't think you need to do this.  It should be safe to return `ResourceDTO(resource, null`.  The methods that depend on the envoyId end up looking it up later themselves. e.g. in `upsertBindingToResource`it also calls `envoyResourceManagement`.

jjbuchan (commented):
I think you'll have to update some tests as well.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1176,7 +1176,7 @@ private static boolean labelSelectorMethodChanged(Monitor original, MonitorCU ne
     for (Entry<String, List<BoundMonitor>> resourceEntry : groupedByResourceId.entrySet()) {
 
       final String resourceId = resourceEntry.getKey();
-      final ResourceDTO resource = resourceApi.getByResourceId(tenantId, resourceId);
```
itzg, line 1179: In a `resource-management` PR it would be good to remove that `getByResourceId` method to prevent us from accidentally re-introducing the call. In fact, removing that in your local workspace will flush out any other uses.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2326,4 +2326,11 @@ public RenderedMonitorTemplate renderMonitorTemplate(UUID monitorId, String reso
     }
     return renderedMonitorTemplate;
   }
+
+  public ResourceDTO findResourceByTenantIdAndResourceId(String tenantId, String resourceId) {
+    Resource resource = resourceRepository.findByTenantIdAndResourceId(tenantId, resourceId)
+        .orElse(null);
+
+    return resource == null ? null : new ResourceDTO(resource, null);
```
itzg, line 2334: `Optional` has a handy [map](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html#map(java.util.function.Function)) method that lets safely remap the object only when it's present:
```suggestion
    return resourceRepository.findByTenantIdAndResourceId(tenantId, resourceId)
        .map(resource -> new ResourceDTO(resource,null));
```

itzg (commented):
Just the one small change.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
rohitsngh27: This will result in an Optional and then i have to make add condition to check if Optional is empty return null otherwise return Resource Object.

itzg: Sorry, yes, it still needs the `.orElse(null)`

itzg:
Github is being a little weird now since it is showing the merged commit

https://github.com/racker/salus-telemetry-monitor-management/pull/206/commits/f9891401f3124dda9d0d82fb0cedbac047b905ef

as part of the diffs; however, it does looks like merge conflicts in `MonitorManagementTest` that weren't resolved quite right. 

I'll see what a local merge does for me.

itzg:
Trying a reset and then rebase worked for me locally. Here's what I did in IntelliJ after cloning your branch:

https://streamable.com/2lkuu7

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
rohitsngh27: Done that in resource management PR

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2276,4 +2278,11 @@ public RenderedMonitorTemplate renderMonitorTemplate(UUID monitorId, String reso
...
```
itzg, line 2286: With my clarification, I meant:
```suggestion
    return resourceRepository.findByTenantIdAndResourceId(tenantId, resourceId)
        .map(resource -> new ResourceDTO(resource,null))
        .orElse(null);
```

itzg (approved):
One optional change, but all looks good. Please wait for @jjbuchan 's re-review before merging.

jjbuchan (commented):
Looks good.  Will approve once the final `Optional` change is done.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -105,6 +106,7 @@
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.stereotype.Service;
+import org.springframework.util.Assert;
```
jjbuchan, line 109: This isn't used either.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -48,6 +48,7 @@
 import com.rackspace.salus.telemetry.entities.Zone;
 import com.rackspace.salus.telemetry.errors.AlreadyExistsException;
 import com.rackspace.salus.telemetry.etcd.services.EnvoyResourceManagement;
+import com.rackspace.salus.telemetry.etcd.services.ZoneStorage;
```
jjbuchan, line 51: Doesn't look like these two new imports are being used.

I'm guessing it was introduced when resolving previous conflicts.

jjbuchan (commented):
Some minor comments.  Also, per a comment on another PR, we can maybe clean up the pom.  If resourceApi isn't used at all anywhere we can remove this block (but I've not checked to see if all uses are removed - I'll let you confirm).

```
    <dependency>
      <groupId>com.rackspace.salus</groupId>
      <artifactId>salus-telemetry-resource-management</artifactId>
      <version>0.0.1-SNAPSHOT</version>
      <classifier>client</classifier>
      <exclusions>
        <exclusion>
          <groupId>*</groupId>
          <artifactId>*</artifactId>
        </exclusion>
      </exclusions>
    </dependency>
```

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
rohitsngh27: Yes...Will resolve that

itzg (approved):

PR #207, opened 2020-09-08
added success counter metrics in MonitorContentTranslationService

shabd67:
# Resolves

[SALUS-1008](https://jira.rax.io/browse/SALUS-1008)

# What

Adding Success/Failure Metrics count to be displayed onto Grafana

# Depends on

https://github.com/racker/salus-common/pull/74


itzg (commented):
@jjbuchan @GeorgeJahad @harageth , seeing these metrics in context of success vs failure makes me wondering if we've been "inverting" our usage of metric names and discriminating tags. Or more specifically, we've been perhaps over-using discriminating tags. What we have tended to do actually looks similar to the "Bad approach" [shown in the Micrometer docs](https://micrometer.io/docs/concepts#_tag_naming).

Having explored several TSDBs lately I am liking the verbosity and expressiveness of naming conventions like [Prometheus'](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels). Micrometer even points to [this Graphite article](https://matt.aimonetti.net/posts/2013-06-practical-guide-to-graphite-monitoring/) for advice on naming conventions.

shabd67:
Sample Metric:

[
    "service_operations_succeeded,app=salus-telemetry-monitor-management,host=MCR872MD6M,metric_type=counter,operation=createMonitor"
],
[
    "service_operations_succeeded,app=salus-telemetry-monitor-management,host=MCR872MD6M,metric_type=counter,operation=removeMonitor"
],
[
    "service_operations_succeeded,app=salus-telemetry-monitor-management,host=MCR872MD6M,metric_type=counter,operation=updateMonitor"
]

src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
```diff
@@ -211,8 +220,10 @@ public String translateMonitorContent(List<MonitorTranslationOperator> operators
     try {
       contentTree = (ObjectNode) objectMapper.readTree(monitorContent);
     } catch (IOException e) {
+      translateMonitorContentFailed.tags("operation", "monitorContentTranslate").register(meterRegistry).increment();
```
itzg, line 223: With comment above
```suggestion
      translateMonitorContentFailed.tags("operation", "translate").register(meterRegistry).increment();
```

itzg (commented):
A few comments so far

src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
```diff
@@ -83,8 +87,9 @@ public MonitorContentTranslationService(
     this.monitorTranslationOperatorRepository = monitorTranslationOperatorRepository;
     this.objectMapper = objectMapper;
 
-    this.monitorTranslateErrors = meterRegistry.counter("errors",
-        "operation", "monitorTranslate");
+    this.meterRegistry = meterRegistry;
+    translateMonitorContentSuccess = Counter.builder(MetricsName.SERVICE_OPERATION_SUCCEEDED_METRIC_NAME);
+    translateMonitorContentFailed  = Counter.builder(MetricsName.SERVICE_OPERATION_FAILED_METRIC_NAME);
```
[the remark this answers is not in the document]
shabd67: yes, that would be good and efficient for debugging as well.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -181,20 +180,11 @@ public MonitorManagement(
...
-        "objectType", "boundMonitor");
-    orphanedPolicyMonitors = meterRegistry.counter("policy_integrity",
-        "operation", "handleMonitorPolicyEvent", "reason", "orphanedPolicyMonitor");
-    policyIntegrityErrors = meterRegistry.counter("policy_integrity",
-        "operation", "handleMonitorPolicyEvent", "reason", "tooManyClones");
+
+    this.meterRegistry = meterRegistry;
+    createMonitorSuccess = Counter.builder(MetricsName.SERVICE_OPERATION_SUCCEEDED_METRIC_NAME);
+    createMonitorFailed  = Counter.builder(MetricsName.SERVICE_OPERATION_FAILED_METRIC_NAME);
+    orphanedBoundMonitorRemoved = Counter.builder("orphaned");
```
itzg, line 187: It looks like this one can be created like normal rather than using a builder; however, the consistency of builders makes sense too.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -296,6 +286,11 @@ public Monitor createMonitor(String tenantId, @Valid MonitorCU newMonitor) throw
 
     final Set<String> affectedEnvoys = bindMonitor(tenantId, monitor, monitor.getZones());
     sendMonitorBoundEvents(affectedEnvoys);
+    if (meterRegistry != null) {
```
itzg, line 289: When would `meterRegistry` be null?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -399,6 +395,7 @@ public Monitor createPolicyMonitor(@Valid MonitorCU newMonitor) {
     metadataUtils.setMetadataFieldsForMonitor(POLICY_TENANT, monitor, false);
 
     monitor = monitorRepository.save(monitor);
+    createMonitorSuccess.tags("operation","createMonitor","objectType","policyMonitor").register(meterRegistry).increment();
```
itzg, line 398: Assuming the existence of a "service" tag and then the "objectType" here, then similarly "operation" can be just "create".

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -569,14 +566,16 @@ private void saveBoundMonitors(List<BoundMonitor> boundMonitors) {
...
+      createMonitorFailed.tags("operation","boundMonitorSaveAll").register(meterRegistry).increment();
+
       for (BoundMonitor bound : boundMonitors) {
         try {
           // save does not fail if the entry already exists, it will act like an update instead.
           boundMonitorRepository.save(bound);
         } catch (Exception ex) {
           log.error("Failed to save boundMonitor={}", bound, ex);
-          boundMonitorSaveErrors.increment();
+          createMonitorFailed.tags("operation","boundMonitorSave").register(meterRegistry).increment();
```
itzg, line 578: Metrics on every bound monitor operation could get inefficient and tedious to cover in all the places. There's a version of `increment` that takes a value to increment in total.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -793,7 +792,7 @@ public Monitor updateMonitor(String tenantId, UUID id, @Valid MonitorCU updatedV
     monitor = monitorRepository.save(monitor);
 
     sendMonitorBoundEvents(affectedEnvoys);
-
+    createMonitorSuccess.tags("operation","updateMonitor").register(meterRegistry).increment();
```
itzg, line 795: To be consistent with others, should include an "objectType". And with that "operation" can be simply "update".

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1055,7 +1054,7 @@ private static boolean labelSelectorMethodChanged(Monitor original, MonitorCU ne
       if (resource.isEmpty()) {
         // remove any orphaned bound monitors
         log.warn("Removing orphaned bound monitor for resourceId={} and monitor={}", resourceId, monitor);
-        orphanedBoundMonitorRemoved.increment();
+        orphanedBoundMonitorRemoved.tags("objectType", "boundMonitor").register(meterRegistry).increment();
```
itzg, line 1057: For this particular metric, the "objectType" doesn't seem to be necessary.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1097,7 +1096,7 @@ private static boolean labelSelectorMethodChanged(Monitor original, MonitorCU ne
         } catch (InvalidTemplateException e) {
           log.warn("Unable to render monitor={} onto resource={}",
               monitor, resource.get(), e);
-          invalidTemplateErrors.increment();
+          createMonitorFailed.tags("operation","renderMonitorTemplate").register(meterRegistry).increment();
```
itzg, line 1099: On failure metrics it would be good to have a "reason" or "exception" tag. An "exception" tag would be logical to add and can be simply the name of the caught exception, like "InvalidTemplateException" in this case or `e.getClass().getSimpleName()` to save typing out specific names each time.

itzg (commented):
A few more comments that you can use as a pattern for the other files.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
shabd67: now, it will be look like:
`createMonitorSuccess = Counter.builder(MetricsName.SERVICE_OPERATION_SUCCEEDED_METRIC_NAME).tag("service","MonitorManagement");
createMonitorFailed  = Counter.builder(MetricsName.SERVICE_OPERATION_FAILED_METRIC_NAME).tag("service","MonitorManagement");
orphanedBoundMonitorRemoved = Counter.builder("orphaned").tag("service","MonitorManagement");`

itzg: 👍

jjbuchan: @shabd67 fyi, three ` form a multi-line code-block.  You can see by adding two more at the beginning and end of your example.  The functionality of that in github and slack is the same.
```
for example this was started with ``` and ended with ```
```

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
jjbuchan: Agreed that metrics for exception types is good.  Log messages should handle any more detail than that.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
jjbuchan: I don't think it should.  You may have copied this from the KafkaErrorConfig?  If so, the only reason it needed this check is because it relies on an `Optional` autowired dependency.  Any of our services should not have that same problem as they are the ones that are initializing that dependency.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
jjbuchan, line 92: I'd shorten these to remove the trailing `_METRIC_NAME`.  The class being called `MetricsName` helps provide enough context.

jjbuchan: and I guess a nit on that... I'd prefer `MetricName` or `MetricNames`.  Currently it feels like the `s` is in the wrong place :)

src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
```diff
@@ -99,7 +104,9 @@ public MonitorTranslationOperator create(MonitorTranslationOperatorCreate in) {
         .setOrder(in.getOrder());
 
     log.info("Creating new monitorTranslationOperator={}", operator);
-    return monitorTranslationOperatorRepository.save(operator);
+    MonitorTranslationOperator monitorTranslationOperator = monitorTranslationOperatorRepository.save(operator);
+    translateMonitorContentSuccess.tags("operation", "CreateMonitorTranslationOperator").register(meterRegistry).increment();
```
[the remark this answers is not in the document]
jjbuchan: I'm good with that.  We do lose the natural grouping based on individual tags if `create` means a variety of things across services but the benefit of grouping by a selection of services and having simpler names outweighs the crossover that may happen imo.

src/main/java/com/rackspace/salus/monitor_management/services/TestMonitorService.java
```diff
@@ -108,6 +122,7 @@ public TestMonitorService(MonitorConversionService monitorConversionService,
     final TestMonitorRequestEvent event;
     try {
       if (!SUPPORTED_AGENT_TYPES.contains(monitorCU.getAgentType())) {
+        testMonitorFailed.tags("operation", "performTestMonitor").register(meterRegistry).increment();
```
jjbuchan, line 125: Just noting that all these could use the previously mentioned exception logging too.

jjbuchan (commented):
I agree with Geoff's points and think this is a great start.  The most important thing is producing metrics to begin with.  As time goes on and we're looking at graphs more this'll help make it easy to tweak things as we need and better see what might be missing.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
shabd67: yes, it was copied from there and post that, I just forgot to remove that piece of code.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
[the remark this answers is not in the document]
shabd67: I think `"operation", "create","objectType","monitorTranslationOperator"` as tags would be better as it demonstrates that the operation type is "create" and objectType which has been created is "monitorTranslationOperator". @jjbuchan and @itzg , your views on it?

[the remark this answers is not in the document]
jjbuchan: I do like the idea of including a `service` tag for each of these.  Would you include that in addition to `objectType`?

I'd assume most services would only have one `objectType` but some may have multiple?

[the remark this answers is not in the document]
jjbuchan: ...although saying that if there is a common tag that includes the application name (such as monitoring-management) then maybe the service tag is essentially redundant.

I'm willing to +1 either of the options suggested.  I really do think once we have the metrics out there and are using them it'll be easier to determine whether we need to tweak them.  In reality, either option will most likely work fine.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
shabd67: instead of having name of caught exception, I think it's better to have toString() of caught exception as that would cover exception class name along with exception message.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
shabd67: Noted.

shabd67:
@jjbuchan / @itzg , if you notice I have written metrics generation logic to all the service classes which contains main business logic like MonitorManagement, MonitorContentTranslationService, TestMonitorService and ZoneManagement only. Do I need to write the same logic for other service classes like Kafka producer and consumer classes, ZoneResolver, MonitorContentRenderer ?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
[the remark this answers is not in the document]
itzg: I'd recommend including "service" since I was thinking that would be one level finer grained than "app" to indicate which Spring **service** specifically did the operation.

So, looks like we've arrived at three standard tags (for most cases): `operation`, `service`, `objectType`

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
itzg: Using the `toString` value is dangerous. Tag value cardinality needs to be well bounded since the number of metric names and unique combination of tag key-values drives the number of metrics our TSDB needs to track. Right now in UMB we're tracking about 5 million metric-sets...which is already a lot :)

src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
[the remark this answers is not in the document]
shabd67: while for Success cases: 
`[
                            1599733529204000000,
                            "salus-telemetry-monitor-management",
                            null,
                            "MCR872MD6M",
                            "counter",
                            "monitor",
                            "create",
                            "MonitorManagement",
                            1
                        ],
                        [
                            1599733529205000000,
                            "salus-telemetry-monitor-management",
                            null,
                            "MCR872MD6M",
                            "counter",
                            "monitor",
                            "remove",
                            "MonitorManagement",
                            1
                        ]`

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
shabd67: yes, that might not be helpful then.

itzg (commented):
I like the "operation", "service", "objectType" changes.

@jjbuchan and others, I'd like to get your thought on the comment I left below.

(I'm going to pause my review until we ponder the error metric strategy.)

shabd67:
@itzg / @jjbuchan , will change lambda and functional optional usage in MonitorManagement once you approve current RestExceptionHandler approach.

src/main/java/com/rackspace/salus/monitor_management/web/controller/RestExceptionHandler.java
```diff
@@ -38,26 +41,34 @@
...
+    public RestExceptionHandler(ErrorAttributes errorAttributes, MeterRegistry meterRegistry) {
         super(errorAttributes);
+        this.meterRegistry = meterRegistry;
+        monitorManagementErrorCounter = Counter.builder(MetricNames.SERVICE_OPERATION_FAILED);
     }
 
     @ExceptionHandler({NotFoundException.class})
     public ResponseEntity<?> handleNotFound(HttpServletRequest request, Exception e) {
         logRequestFailure(request, e);
+        monitorManagementErrorCounter.tags("exception","NotFoundException").register(meterRegistry).increment();
```
itzg, line 57: I think it would be good to include `request.getServletPath()` as a tagged value similar to what Spring MVC metrics do. That way we will have a pretty good indication of what logs to find when debugging.

itzg, line 57: Can use `e.getClass().getSimpleName()` here to simplify copy-paste across methods and modules.

itzg (commented):
I like this approach with RestExceptionHandler with servlet path added. 

I just confirmed that @jjbuchan  had already added [similar error metric handling in our kafka listener error handler](https://github.com/racker/salus-common/blob/bcf53c9f841961c10fac4c6a88bf0679b2ff9fa0/src/main/java/com/rackspace/salus/common/messaging/KafkaErrorConfig.java#L75)...now that I think about it, that's probably the counter builder example that was used :)

src/main/java/com/rackspace/salus/monitor_management/services/TestMonitorService.java
```diff
@@ -176,6 +188,7 @@ public TestMonitorService(MonitorConversionService monitorConversionService,
                 .setErrors(List.of(String
                     .format("An unexpected internal error occurred: %s", throwable.getMessage())));
```
itzg, line 189: Would be good to increment an error counter here

src/main/java/com/rackspace/salus/monitor_management/services/TestMonitorService.java
```diff
@@ -196,12 +209,13 @@ private String resolveRemoteEnvoy(String tenantId,
...
-            "No envoys were available in the given monitoring zone"));
-
-    return leastLoaded.getEnvoyId();
+    final Optional<EnvoyResourcePair> leastLoadedOptional = monitorManagement
+        .findLeastLoadedEnvoyInZone(resolveZone(tenantId, monitoringZones.get(0)));
+    if(leastLoadedOptional.isEmpty()) {
+      throw new MissingRequirementException(
+          "No envoys were available in the given monitoring zone");
+    }
+    return leastLoadedOptional.get().getEnvoyId();
```
itzg, line 218: Can be reverted

src/main/java/com/rackspace/salus/monitor_management/services/ZoneManagement.java
```diff
@@ -329,5 +349,6 @@ public void removeAllTenantZones(String tenantId, boolean sendEvents) {
       }else {
         zoneRepository.deleteAllByTenantId(tenantId);
       }
+      zoneManagementSuccess.tags("operation","removeAll","objectType","zone").register(meterRegistry).increment();
```
itzg, line 352: To align with others
```suggestion
      zoneManagementSuccess.tags("operation","removeAll","objectType","privateZone").register(meterRegistry).increment();
```

itzg (changes requested):
Just small changes now

src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
```diff
@@ -103,7 +105,7 @@ public MonitorTranslationOperator create(MonitorTranslationOperatorCreate in) {
 
     log.info("Creating new monitorTranslationOperator={}", operator);
     MonitorTranslationOperator monitorTranslationOperator = monitorTranslationOperatorRepository.save(operator);
-    translateMonitorContentSuccess.tags("operation", "create","objectType","monitorTranslationOperator").register(meterRegistry).increment();
+    translateMonitorContentSuccess.tags(MetricTags.OPERATION_METRIC_TAG, "create",MetricTags.OBJECT_TYPE_METRIC_TAG,"monitorTranslationOperator").register(meterRegistry).increment();
```
itzg, line 108: Were you going to go ahead and create a common metric tag value for "create"?

shabd67: yes

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1168,7 +1174,6 @@ private static boolean labelSelectorMethodChanged(Monitor original, MonitorCU ne
         } catch (InvalidTemplateException e) {
           log.warn("Unable to render updatedContent='{}' of monitor={} for resource={}",
               updatedContent, monitor, resource, e);
-          invalidTemplateErrors.increment();
```
itzg, line 1171: This one needs `invalidTemplateErrors` incremented

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1427,7 +1434,6 @@ private void handlePolicyAdditionEvent(MonitorPolicyEvent event, Monitor clonedM
     List<Monitor> existingPolicyMonitor = existingPolicyIds.stream()
         .map(id -> monitorPolicyRepository.findById(id).orElseGet(() -> {
           log.warn("Monitor is tied to non-existent policy={}", id);
-          orphanedPolicyMonitors.increment();
```
itzg, line 1430: This one needs `orphanedBoundMonitorRemoved` incremented

src/main/java/com/rackspace/salus/monitor_management/services/MonitorContentTranslationService.java
```diff
@@ -163,9 +171,9 @@ public void delete(UUID operatorId) {
                         boundMonitor.getRenderedContent()
                     )
                 );
+            return boundMonitorDTO;
           } catch (MonitorContentTranslationException e) {
             log.error("Failed to translate boundMonitor={}", boundMonitor, e);
-            monitorTranslateErrors.increment();
```
itzg, line 168: Just noticed this one needs a specific error counter added back.

src/main/java/com/rackspace/salus/monitor_management/services/TestMonitorService.java
itzg: Still would like a specific error counter here

src/main/java/com/rackspace/salus/monitor_management/web/controller/RestExceptionHandler.java
```diff
@@ -38,26 +42,37 @@
...
+    public RestExceptionHandler(ErrorAttributes errorAttributes, MeterRegistry meterRegistry) {
         super(errorAttributes);
+        this.meterRegistry = meterRegistry;
+        monitorManagementErrorCounter = Counter.builder(MetricNames.SERVICE_OPERATION_FAILED);
     }
 
     @ExceptionHandler({NotFoundException.class})
     public ResponseEntity<?> handleNotFound(HttpServletRequest request, Exception e) {
         logRequestFailure(request, e);
+        monitorManagementErrorCounter.tags(MetricTags.URI_METRIC_TAG,request.getServletPath(),
```
itzg, line 58: Can you double check that the servlet path includes just the path variables and not specific values, like the actual tenant ID?

itzg (changes requested):
Just a couple more spots.

src/main/java/com/rackspace/salus/monitor_management/web/controller/RestExceptionHandler.java
shabd67: yes, it like `1600093980109000000 salus-telemetry-monitor-management NotFoundException MCR872MD6M counter     /api/tenant/aaaaaa/monitors/38633258-701b-46ee-b88b-a0804d826dfa 2`

itzg: Uh oh, that's not good. I wonder which field of the request has the equivalent of the `@RequestMapping` path?

shabd67: its like 

> time                app                                exception         host       metric_type uri                                                              value
----                ---                                ---------         ----       ----------- ---                                                              -----
1600093980109000000 salus-telemetry-monitor-management NotFoundException MCR872MD6M counter     /api/tenant/aaaaaa/monitors/38633258-701b-46ee-b88b-a0804d826dfa 2

itzg: Yes, I understand that. I would like the path to be `/tenant/{tenantId}/monitors/{uuid}` to avoid an explosion of distinct tag values.

itzg: From https://github.com/spring-projects/spring-boot/blob/76bf89ecd2db8031658cd716ea13206c738dd7f6/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/servlet/WebMvcTags.java#L156 it looks like `request.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE)` will give that

shabd67: yes, its working fine. thank you

itzg (changes requested):
Just needs the templated URL path mentioned in https://github.com/racker/salus-telemetry-monitor-management/pull/207#discussion_r487994017

src/main/java/com/rackspace/salus/monitor_management/web/controller/RestExceptionHandler.java
```diff
@@ -55,23 +56,23 @@ public RestExceptionHandler(ErrorAttributes errorAttributes, MeterRegistry meter
     @ExceptionHandler({NotFoundException.class})
     public ResponseEntity<?> handleNotFound(HttpServletRequest request, Exception e) {
         logRequestFailure(request, e);
-        monitorManagementErrorCounter.tags(MetricTags.URI_METRIC_TAG,request.getServletPath(),
+        monitorManagementErrorCounter.tags(MetricTags.URI_METRIC_TAG,request.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE).toString(),
```
itzg, line 59: Did this work as expected?

shabd67: yes. 

> 1600095461417000000 salus-telemetry-monitor-management NotFoundException MCR872MD6M counter     /api/tenant/{tenantId}/monitors/{uuid} 2

itzg: Excellent!

itzg (dismissed):

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementExcludeResourceIdsTest.java
```diff
@@ -81,6 +84,7 @@
 @AutoConfigureDataJpa
 @AutoConfigureTestEntityManager
 @Transactional
+@Import({MetricsAutoConfiguration.class, CompositeMeterRegistryAutoConfiguration.class})
```
itzg, line 87: I usually just import `SimpleMeterRegistry` to get both a real instance and not all of the auto-configured exports that we might happen to have in our apps (however, they're disabled by default).
```suggestion
@Import({SimpleMeterRegistry.class})
```

itzg (changes requested):
Use SimpleMeterRegistry since it is much...simpler.

src/test/java/com/rackspace/salus/monitor_management/services/TestMonitorServiceTest.java
```diff
@@ -78,7 +82,7 @@
   private static final Duration resultsTimeout = Duration.ofMillis(500);
 
   @Configuration
-  @Import({TestMonitorService.class})
+  @Import({TestMonitorService.class, MetricsAutoConfiguration.class, CompositeMeterRegistryAutoConfiguration.class})
```
itzg, line 85: This one had the inner config class, which threw off your pattern :)
```suggestion
  @Import({TestMonitorService.class, SimpleMeterRegistry.class})
```

src/test/java/com/rackspace/salus/monitor_management/services/TestMonitorServiceTest.java
```diff
@@ -70,6 +73,7 @@
 import org.springframework.test.context.junit4.SpringRunner;
 
 @RunWith(SpringRunner.class)
+@Import({SimpleMeterRegistry.class})
```
itzg, line 76: Remove this one and change `@Import` below.

itzg (changes requested):
Just one super small change.

itzg (approved):

PR #213, opened 2020-10-07
Updated logs which was causing OOM error

rohitsngh27:
# Resolves
https://jira.rax.io/browse/SALUS-1031

# What
Resolve OOM error caused by delete tenant for large number of bound monitors

# How
Logs need to be updated

# How to test
This can be tested via Rest Api call

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1897,7 +1897,7 @@ public class MonitorManagement {
     final List<BoundMonitor> boundMonitors =
         boundMonitorRepository.findAllByTenantIdAndMonitor_IdIn(tenantId, monitorIdsToUnbind);
 
-    log.debug("Unbinding boundMonitorCount {} from {} monitorCount",
+    log.debug("Unbinding boundMonitorCount={} from monitorCount={}",
         boundMonitors.size(), monitorIdsToUnbind.size());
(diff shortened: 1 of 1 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1897,8 +1897,8 @@ private void handleReattachedEnvoy(String tenantId, String resourceId, String en
     final List<BoundMonitor> boundMonitors =
         boundMonitorRepository.findAllByTenantIdAndMonitor_IdIn(tenantId, monitorIdsToUnbind);
 
-    log.debug("Unbinding {} from monitorIds={}",
-        boundMonitors, monitorIdsToUnbind);
+    log.debug("Unbinding {} from {} monitors",
```
itzg, line 1900: To give some more context to the new log structure
```suggestion
    log.debug("Unbinding boundMonitorCount={} from monitorCount={}",
```

itzg (changes requested):
Are there not any other log statements that take `List<BoundMonitor>` parameters?

rohitsngh27:
> Are there not any other log statements that take `List<BoundMonitor>` parameters?

Added few more

itzg (changes requested):
Other spots also need the `field={}` syntax

itzg (approved):

jjbuchan (commented):
Good find!

PR #215, opened 2020-10-08
Using Projection to resolve OOM error for large BMs

rohitsngh27:
# Resolves
https://jira.rax.io/browse/SALUS-1031

# What
Using Projection to resolve OOM error for large BMs

# How
Using BoundMonitorProjection for retrieval of BMs.

# How to test
This can be tested via Rest API.

# Why
This limits the data to what is only required to process bound monitor retrieval.

# Depends on
https://github.com/racker/salus-telemetry-model/pull/155


itzg (approved):
Looks good. Wait on another review before merging please.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
```diff
@@ -689,8 +690,6 @@ public void testRefreshPolicyMonitorsForTenant_removePolicy() {
 
     // policy monitor no longer exists on tenant
     assertTrue(monitorRepository.findByTenantIdAndPolicyId(tenantId, policyId).isEmpty());
-
-    verify(boundMonitorRepository).deleteAll(anyIterable());
```
jjbuchan, line 693: If I'm following this correctly you should replace this with a new `verify` for `deleteById`.

If for any tests there are more than one being removed you can use the `times()` method.

e.g. `verify(boundMonitorRepository, times(5)).deleteById(any());`

jjbuchan (changes requested):
Looks good other than the missing assertions.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
itzg: Since the "findAll" is returning an empty list
```
    when(boundMonitorRepository.findAllByTenantIdAndMonitor_IdIn(tenantId, List.of(clonedMonitor.getId())))
        .thenReturn(Collections.emptyList());
```

I think it turns out that `deleteById` doesn't get called.

jjbuchan: Oh, I see it would do the delete before even if it was empty.  I should have looked harder.

jjbuchan (approved):
All good but I'd like to see the response to my question in https://github.com/racker/salus-telemetry-model/pull/155 before this is merged.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
itzg: I had started to write the same comment as you :) , but happened to expand the diff above the change.

PR #220, opened 2020-10-16
Added handling for such in effect policies whose monitor is not found while deleting cloned monitor

shabd67:
# Resolves

[SALUS-1024](https://jira.rax.io/browse/SALUS-1024)

# What

Added handling for such in effect policies whose monitor is not found while deleting cloned monitor

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1490,32 +1490,32 @@ public class MonitorManagement {
     List<MonitorPolicyDTO> effectivePolicies = policyApi.getEffectiveMonitorPoliciesForTenant(tenantId, false);
 
     // if the old policy had been overriding another one using the same monitorId, find the other one
-      List<UUID>  newPolicyIds = effectivePolicies.stream()
-            .filter(p -> event.getMonitorId() != null)
-            .filter(p -> p.getMonitorId().equals(monitorId))
-            .map(PolicyDTO::getId)
-            .collect(Collectors.toList());
+    List<UUID>  newPolicyIds = effectivePolicies.stream()
+        .filter(p -> event.getMonitorId() != null)
+        .filter(p -> p.getMonitorId().equals(monitorId))
+        .map(PolicyDTO::getId)
+        .collect(Collectors.toList());
 
-      if (newPolicyIds.isEmpty()) {
-        log.debug("Removing cloned policy monitor={} for event={}", clonedMonitor.getId(), event);
-        unbindAndRemoveMonitor(clonedMonitor);
-        return;
-      }
+    if (newPolicyIds.isEmpty()) {
+      log.debug("Removing cloned policy monitor={} for event={}", clonedMonitor.getId(), event);
+      unbindAndRemoveMonitor(clonedMonitor);
+      return;
+    }
 
-      if (newPolicyIds.size() > 1) {
-        log.error(
-            "More than one effective policy configured to use the same monitorId={} for tenant={}",
-            monitorId, tenantId);
-        policyIntegrityErrors
-            .tags(MetricTags.OPERATION_METRIC_TAG, "handleMonitorPolicyEvent", "reason",
-                "tooManyClones")
-            .register(meterRegistry).increment();
-      }
-      UUID newPolicyId = newPolicyIds.get(0);
-      log.debug("Updating cloned policy monitor={} with new policyId={}", clonedMonitor,
-          newPolicyId);
-      clonedMonitor.setPolicyId(newPolicyId);
-      monitorRepository.save(clonedMonitor);
+    if (newPolicyIds.size() > 1) {
+      log.error(
+          "More than one effective policy configured to use the same monitorId={} for tenant={}",
+          monitorId, tenantId);
+      policyIntegrityErrors
+          .tags(MetricTags.OPERATION_METRIC_TAG, "handleMonitorPolicyEvent", "reason",
+              "tooManyClones")
+          .register(meterRegistry).increment();
+    }
+    UUID newPolicyId = newPolicyIds.get(0);
+    log.debug("Updating cloned policy monitor={} with new policyId={}", clonedMonitor,
+        newPolicyId);
+    clonedMonitor.setPolicyId(newPolicyId);
+    monitorRepository.save(clonedMonitor);
   }
 
   /**
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -22,7 +22,6 @@ import static com.rackspace.salus.telemetry.entities.Resource.REGION_METADATA;
 import static com.rackspace.salus.telemetry.etcd.types.ResolvedZone.resolveZone;
 import static org.springframework.util.CollectionUtils.isEmpty;
 
-import brave.Tracer;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.google.common.collect.Streams;
 import com.google.common.math.Stats;
@@ -61,8 +60,8 @@ import com.rackspace.salus.telemetry.messaging.MonitorPolicyEvent;
 import com.rackspace.salus.telemetry.messaging.ResourceEvent;
 import com.rackspace.salus.telemetry.messaging.TenantPolicyChangeEvent;
 import com.rackspace.salus.telemetry.model.ConfigSelectorScope;
-import com.rackspace.salus.telemetry.model.JobType;
 import com.rackspace.salus.telemetry.model.JobStatus;
+import com.rackspace.salus.telemetry.model.JobType;
 import com.rackspace.salus.telemetry.model.LabelSelectorMethod;
 import com.rackspace.salus.telemetry.model.MonitorType;
 import com.rackspace.salus.telemetry.model.NotFoundException;
@@ -1395,14 +1394,18 @@ public class MonitorManagement {
 
     String tenantId = event.getTenantId();
     UUID policyId = event.getPolicyId();
-    UUID monitorId = event.getMonitorId();
+    MonitorPolicy policy = monitorPolicyRepository.findById(policyId).orElse(null);
 
-    if (monitorId == null) {
+    if (policy != null) {
+      // if policy exists do not trust the monitorId in the event is up to date
+      // delays in processing can occur.
+      event.setMonitorId(policy.getMonitorId());
+    }
+    if (event.getMonitorId() == null) {
       handlePolicyOptOutEvent(event);
       return;
     }
 
-    MonitorPolicy policy = monitorPolicyRepository.findById(policyId).orElse(null);
     Monitor clonedMonitorForEventPolicy = monitorRepository.findByTenantIdAndPolicyId(tenantId, policyId)
         .orElse(null);
 
@@ -1909,7 +1912,7 @@ public class MonitorManagement {
       return new HashSet<>();
     }
     Pageable pageRequest = PageRequest.of(0, 1000);
-    Page<BoundMonitor> boundMonitors =boundMonitorRepository
+    Page<BoundMonitor> boundMonitors = boundMonitorRepository
         .findAllByTenantIdAndMonitor_IdIn(tenantId, monitorIdsToUnbind, pageRequest);
     Set<String> extractedEnvoyIds = new HashSet<>();
 
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1491,8 +1491,7 @@ public class MonitorManagement {
 
     // if the old policy had been overriding another one using the same monitorId, find the other one
     List<UUID>  newPolicyIds = effectivePolicies.stream()
-        .filter(p -> event.getMonitorId() != null)
-        .filter(p -> p.getMonitorId().equals(monitorId))
+        .filter(p -> Objects.equals(p.getMonitorId(), monitorId))
         .map(PolicyDTO::getId)
         .collect(Collectors.toList());
 
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1488,27 +1488,36 @@ public class MonitorManagement {
     List<MonitorPolicyDTO> effectivePolicies = policyApi.getEffectiveMonitorPoliciesForTenant(tenantId, false);
 
     // if the old policy had been overriding another one using the same monitorId, find the other one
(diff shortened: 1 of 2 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1488,27 +1490,47 @@ private void handlePolicyRemovalEvent(MonitorPolicyEvent event, Monitor clonedMo
...
+            .tags(MetricTags.OPERATION_METRIC_TAG, "handleMonitorPolicyEvent", "reason",
+                "tooManyClones")
+            .register(meterRegistry).increment();
+      }
+      UUID newPolicyId = newPolicyIds.get(0);
+      log.debug("Updating cloned policy monitor={} with new policyId={}", clonedMonitor,
+          newPolicyId);
+      clonedMonitor.setPolicyId(newPolicyId);
+      monitorRepository.save(clonedMonitor);
+    } catch(Exception e)  {
```
itzg, line 1527: Similar to other comment, catching `Exception` is too broad. Where was the exception happening and what exactly?

itzg (changes requested):

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1488,27 +1490,47 @@ private void handlePolicyRemovalEvent(MonitorPolicyEvent event, Monitor clonedMo
...
-        .collect(Collectors.toList());
+    try {
+      List<UUID> newPolicyIds = null;
+      //if policy is in effect and monitor not found
+      try {
+        newPolicyIds = effectivePolicies.stream()
+            .filter(p -> p.getMonitorId().equals(monitorId))
+            .map(PolicyDTO::getId)
+            .collect(Collectors.toList());
+      } catch(NullPointerException e)  {
```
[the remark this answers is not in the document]
jjbuchan: Yeah, this logic isn't quite right.  we want to ignore `null` monitorIds.  An exception should never be thrown/caught.

jjbuchan (changes requested):

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1488,27 +1490,32 @@ private void handlePolicyRemovalEvent(MonitorPolicyEvent event, Monitor clonedMo
...
+        policyIntegrityErrors
+            .tags(MetricTags.OPERATION_METRIC_TAG, "handleMonitorPolicyEvent", "reason",
+                "tooManyClones")
+            .register(meterRegistry).increment();
+      }
+      UUID newPolicyId = newPolicyIds.get(0);
+      log.debug("Updating cloned policy monitor={} with new policyId={}", clonedMonitor,
+          newPolicyId);
+      clonedMonitor.setPolicyId(newPolicyId);
+      monitorRepository.save(clonedMonitor);
```
jjbuchan, line 1518: With the recent change is all this idented wrong now?  In IntelliJ you can use the `Code->Auto Indent Lines` option to fix it.

jjbuchan (changes requested):
A test case should be added for this

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1488,7 +1490,8 @@ private void handlePolicyRemovalEvent(MonitorPolicyEvent event, Monitor clonedMo
     List<MonitorPolicyDTO> effectivePolicies = policyApi.getEffectiveMonitorPoliciesForTenant(tenantId, false);
 
     // if the old policy had been overriding another one using the same monitorId, find the other one
-    List<UUID> newPolicyIds = effectivePolicies.stream()
+    List<UUID>  newPolicyIds = effectivePolicies.stream()
+        .filter(p -> event.getMonitorId() != null)
```
itzg, line 1494: Since `event` is scoped at the method level, it seems more efficient and logical to assert `event.getMonitorId()` before this stream and just build an empty `newPolicyIds`.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1488,7 +1490,8 @@ private void handlePolicyRemovalEvent(MonitorPolicyEvent event, Monitor clonedMo
...
         .filter(p -> p.getMonitorId().equals(monitorId))
```
itzg, line 1495: ...or could use the null-safe approach and avoid the need for the added filter above
```suggestion
        .filter(p -> Objects.equals(p.getMonitorId(), monitorId))
```

itzg, line 1495: Comparing with your previous change...was this the line where the exception was thrown?

itzg (changes requested):
I'm still a little confused about the original issue. Speaking of which, please add a unit test that recreates the issue and confirms with code change that it is resolved.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
itzg: In that case `p` must be null since `.equals` evaluates to false when passed a null parameter. I just tested with this code where it prints `result=false`:

```
    String other = null;
    boolean result = "hello".equals(other);
    System.out.println("result="+result);
```

itzg: ...you could use [Objects.equals](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Objects.html#equals(java.lang.Object,java.lang.Object)) to be fully null-safe

shabd67: its p.getMonitorId() returning null. 

`String other = null;
	    boolean result = other.equals("hello");
	    System.out.println("result="+result);`

itzg: Why did you add the filter on `event.getMonitorId() != null` then? It seems like that should have been `p.getMonitorId() != null`. In any case, sounds like `Objects.equals` solve that bit.

jjbuchan (changes requested):
I'm holding off reviewing fully until a test case is added.

The test case should reveal if you have any new bugs.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
shabd67: yes, that was incorrect.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
```diff
@@ -966,6 +967,41 @@ public void testHandleMonitorPolicyEvent_removePolicy_butKeepMonitor() {
     assertTrue(monitorRepository.findByTenantIdAndPolicyId(tenantId, policyId).isEmpty());
   }
 
+  /**
+   * Receive a remove monitor policy event and process it for a tenant
+   * that was previously using the policy and its entry in monitor_policies having
+   * null as monitor id
```
itzg, line 973: It would be good to document here the root cause of why a monitor policy had a null monitor ID in the first place. James can probably help provide some input, if needed.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
```diff
@@ -966,6 +967,41 @@ public void testHandleMonitorPolicyEvent_removePolicy_butKeepMonitor() {
...
+    UUID policyMonitorId = currentMonitor.getId();
+
+    // store a monitor for the tenant that is tied to the policy in the event
+    Monitor clonedMonitor = createMonitorForPolicyForTenant(tenantId, policyId);
+
+    List<MonitorPolicyDTO> list = List.of((MonitorPolicyDTO) new MonitorPolicyDTO()
+        .setMonitorId(null)
+        .setId(policyId));
+
+    doReturn(list).when(policyApi).getEffectiveMonitorPoliciesForTenant(anyString(), anyBoolean());
```
itzg, line 989: Mockito recommends not using `doReturn` except in rare cases

https://javadoc.io/static/org.mockito/mockito-core/3.5.15/org/mockito/Mockito.html#doReturn-java.lang.Object-
```suggestion
    when(policyApi.getEffectiveMonitorPoliciesForTenant(anyString(), anyBoolean())
      .thenReturn(list);
```

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
```diff
@@ -966,6 +967,41 @@ public void testHandleMonitorPolicyEvent_removePolicy_butKeepMonitor() {
...
+
+    PageRequest pageRequest = PageRequest.of(0, 1000);
+
+    doReturn(Page.empty()).when(boundMonitorRepository).findAllByTenantIdAndMonitor_IdIn(tenantId, List.of(clonedMonitor.getId()), pageRequest);
```
itzg, line 993: and here

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
```diff
@@ -966,6 +967,41 @@ public void testHandleMonitorPolicyEvent_removePolicy_butKeepMonitor() {
...
+
+    MonitorPolicyEvent event = (MonitorPolicyEvent) new MonitorPolicyEvent()
+        .setMonitorId(policyMonitorId)
+        .setTenantId(tenantId)
+        .setPolicyId(policyId);
+    monitorManagement.handleMonitorPolicyEvent(event);
+
+    // policy monitor no longer exists on tenant
+    assertTrue(monitorRepository.findByTenantIdAndPolicyId(tenantId, policyId).isEmpty());
```
itzg, line 1002: Be sure to follow the pattern of assertions in the same test class. This class uses `assertThat` with hamcrest matchers, so this would be
```suggestion
    assertThat(monitorRepository.findByTenantIdAndPolicyId(tenantId, policyId), hasSize(0));
```

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
```diff
@@ -966,6 +967,41 @@ public void testHandleMonitorPolicyEvent_removePolicy_butKeepMonitor() {
...
+  }
```
itzg, line 1003: Please balance all `when`'s with a `verify`. Also needs a `verifyNoMoreInteractions`.

itzg (changes requested):

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
shabd67: @jjbuchan , have updated the description for this test case. kindly provide your input onto that.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
```diff
@@ -966,6 +966,45 @@ public void testHandleMonitorPolicyEvent_removePolicy_butKeepMonitor() {
...
+    String tenantId = RandomStringUtils.randomAlphanumeric(10);
+    UUID policyId = UUID.randomUUID();
+    UUID policyMonitorId = currentMonitor.getId();
+
+    // store a monitor for the tenant that is tied to the policy in the event
+    Monitor clonedMonitor = createMonitorForPolicyForTenant(tenantId, policyId);
+
+    List<MonitorPolicyDTO> list = List.of((MonitorPolicyDTO) new MonitorPolicyDTO()
+        .setMonitorId(null)
+        .setId(policyId));
```
jjbuchan, line 985: To test the issue you saw, this list should return multiple items.  The policy with the null monitorId should already not have any monitor stored for it.

The test case should verify that removing a previously configured policy monitor works despite the fact another unrelated policy has a null monitor.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
```diff
@@ -966,6 +966,45 @@ public void testHandleMonitorPolicyEvent_removePolicy_butKeepMonitor() {
     assertTrue(monitorRepository.findByTenantIdAndPolicyId(tenantId, policyId).isEmpty());
   }
 
+  /**
+   * Receive a remove monitor policy event and process it for a tenant
+   * that was previously using the policy but the monitor is not found
```
jjbuchan, line 971: This doesn't quite describe the scenario of this test case enough.

Rather than discussing it more with you right now, I would like you to dig in to try and understand this case and the original problem a little more.  I suspect the more you dig into it to understand things yourself the more you'll learn/remember vs. me just telling you the answer.

Since you'll be doing a few Policy related issues it really will be good to fully understand what's going on.

jjbuchan (changes requested):

itzg (commented):
Latest changes address most of my concerns, but I'll let James provide more specific feedback.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
```diff
@@ -966,6 +966,45 @@ public void testHandleMonitorPolicyEvent_removePolicy_butKeepMonitor() {
...
+   *
+   */
+  @Test
+  public void testHandleMonitorPolicyEvent_removePolicy_null_monitor_id() {
```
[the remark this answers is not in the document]
jjbuchan: With this comment I meant that it was good to keep this test as is (other than changing the comment) but then add a new test for what I described below.  Can you update this file to include the two new tests rather than just one.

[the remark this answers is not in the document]
shabd67: sure. I am adding that previous test case as well.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementPolicyTest.java
```diff
@@ -966,6 +966,47 @@ public void testHandleMonitorPolicyEvent_removePolicy_butKeepMonitor() {
     assertTrue(monitorRepository.findByTenantIdAndPolicyId(tenantId, policyId).isEmpty());
   }
 
+  /**
+   * Receive a remove monitor policy event and process it for a tenant
+   * that was previously using the policy which is not in effect and
+   * having a null monitor id
```
jjbuchan, line 972: This comment isn't very clear.  I'll wait until both tests are up before going into more detail as that'll make it easier to note the differences and if anything is still missing.

jjbuchan (changes requested):

jjbuchan (approved):

PR #221, opened 2020-10-16
Make delete all monitors for tenant async

rohitsngh27:
# Resolves
https://jira.rax.io/browse/SALUS-1051

# What
Make delete all monitors for tenant async

# How
Used @Aysnc annotation to make removeAllTenantMonitors async.

# How to test
This can be tested via rest call

# Depends on
https://github.com/racker/salus-telemetry-model/pull/161

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -45,7 +45,7 @@ import com.rackspace.salus.policy.manage.web.model.PolicyDTO;
 import com.rackspace.salus.resource_management.web.client.ResourceApi;
 import com.rackspace.salus.resource_management.web.model.ResourceDTO;
 import com.rackspace.salus.telemetry.entities.BoundMonitor;
-import com.rackspace.salus.telemetry.entities.Jobs;
+import com.rackspace.salus.telemetry.entities.Job;
 import com.rackspace.salus.telemetry.entities.MetadataPolicy;
 import com.rackspace.salus.telemetry.entities.Monitor;
 import com.rackspace.salus.telemetry.entities.MonitorPolicy;
@@ -61,7 +61,7 @@ import com.rackspace.salus.telemetry.messaging.MonitorPolicyEvent;
 import com.rackspace.salus.telemetry.messaging.ResourceEvent;
 import com.rackspace.salus.telemetry.messaging.TenantPolicyChangeEvent;
 import com.rackspace.salus.telemetry.model.ConfigSelectorScope;
-import com.rackspace.salus.telemetry.model.JobName;
+import com.rackspace.salus.telemetry.model.JobType;
 import com.rackspace.salus.telemetry.model.JobStatus;
 import com.rackspace.salus.telemetry.model.LabelSelectorMethod;
 import com.rackspace.salus.telemetry.model.MonitorType;
@@ -69,7 +69,7 @@ import com.rackspace.salus.telemetry.model.NotFoundException;
 import com.rackspace.salus.telemetry.model.ResourceInfo;
 import com.rackspace.salus.telemetry.model.TargetClassName;
 import com.rackspace.salus.telemetry.repositories.BoundMonitorRepository;
-import com.rackspace.salus.telemetry.repositories.JobsRepository;
+import com.rackspace.salus.telemetry.repositories.JobRepository;
 import com.rackspace.salus.telemetry.repositories.MonitorPolicyRepository;
 import com.rackspace.salus.telemetry.repositories.MonitorRepository;
 import com.rackspace.salus.telemetry.repositories.ResourceRepository;
@@ -96,7 +96,6 @@ import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import javax.persistence.EntityManager;
 import javax.persistence.PersistenceContext;
-import javax.transaction.Transactional;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 import lombok.extern.slf4j.Slf4j;
@@ -149,10 +148,9 @@ public class MonitorManagement {
 
   MeterRegistry meterRegistry;
 
-  @Autowired
   Tracer tracer;
 
-  JobsRepository jobsRepository;
+  JobRepository jobRepository;
 
   // metrics counters
   private final Counter.Builder orphanedBoundMonitorRemoved;
@@ -177,7 +175,8 @@ public class MonitorManagement {
       MonitorConversionService monitorConversionService,
       MetadataUtils metadataUtils, MeterRegistry meterRegistry,
       JdbcTemplate jdbcTemplate,
-      JobsRepository jobsRepository)
+      JobRepository jobRepository,
+      Tracer tracer)
       throws IOException {
     this.resourceRepository = resourceRepository;
     this.monitorPolicyRepository = monitorPolicyRepository;
@@ -197,8 +196,9 @@ public class MonitorManagement {
     this.jdbcTemplate = jdbcTemplate;
     this.labelMatchQuery = SpringResourceUtils.readContent("sql-queries/monitor_label_matching_query.sql");
     this.labelMatchOrQuery = SpringResourceUtils.readContent("sql-queries/monitor_label_matching_OR_query.sql");
-    this.jobsRepository = jobsRepository;
+    this.jobRepository = jobRepository;
     this.meterRegistry = meterRegistry;
+    this.tracer = tracer;
     createMonitorSuccess = Counter.builder(MetricNames.SERVICE_OPERATION_SUCCEEDED).tag(
         MetricTags.SERVICE_METRIC_TAG,"MonitorManagement");
     orphanedBoundMonitorRemoved = Counter.builder("orphaned").tag(MetricTags.SERVICE_METRIC_TAG,"MonitorManagement");
@@ -2328,18 +2328,18 @@ public class MonitorManagement {
    * Saves the Job results in jobs table.
    *
    * @param tenantId
-   * @param jobName
+   * @param jobType
    * @param jobStatus
    * @param description
    */
-  public void saveJobResults(String tenantId, JobName jobName, JobStatus jobStatus, String description) {
+  public void saveJobResults(String tenantId, JobType jobType, JobStatus jobStatus, String description) {
     String id = tracer.currentSpan().context().traceIdString();
-    Jobs job = new Jobs()
+    Job job = new Job()
         .setId(id)
-        .setName(jobName.getValue())
+        .setType(jobType)
         .setStatus(jobStatus)
         .setDescription(description)
         .setTenantId(tenantId);
-    jobsRepository.save(job);
+    jobRepository.save(job);
   }
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -22,6 +22,7 @@ import static com.rackspace.salus.telemetry.entities.Resource.REGION_METADATA;
 import static com.rackspace.salus.telemetry.etcd.types.ResolvedZone.resolveZone;
 import static org.springframework.util.CollectionUtils.isEmpty;
 
+import brave.Tracer;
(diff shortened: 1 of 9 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -142,6 +149,11 @@
 
   MeterRegistry meterRegistry;
 
+  @Autowired
```
jjbuchan, line 152: Is there a reason for this being here instead of being done in the constructor like everything else?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2308,4 +2323,23 @@ public ResourceDTO findResourceByTenantIdAndResourceId(String tenantId, String r
...
+   * @param tenantId
+   * @param jobName
+   * @param jobStatus
+   * @param description
+   */
+  public void saveJobResults(String tenantId, JobName jobName, JobStatus jobStatus, String description) {
+    String id = tracer.currentSpan().context().traceIdString();
+    Jobs job = new Jobs()
+        .setId(id)
+        .setName(jobName.getValue())
```
jjbuchan, line 2339: Looks like you must still be updating this so I'll hold off on the review until you request one again.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
rohitsngh27: No specific reason, will add this in constructor

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2308,4 +2323,23 @@ public ResourceDTO findResourceByTenantIdAndResourceId(String tenantId, String r
...
+   * Saves the Job results in jobs table.
+   *
+   * @param tenantId
+   * @param jobName
+   * @param jobStatus
+   * @param description
+   */
+  public void saveJobResults(String tenantId, JobName jobName, JobStatus jobStatus, String description) {
+    String id = tracer.currentSpan().context().traceIdString();
+    Jobs job = new Jobs()
```
GeorgeJahad, line 2337: Looks like this is out of date?  It's singular now, right?

rohitsngh27: Yes, Updating the PR aligning with latest model changes

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
rohitsngh27: Updated the PR

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2253,7 +2266,8 @@ private String getEnvoyIdForResource(Resource resource) {
     return monitorRepository.search(tenantId, searchCriteria, page);
   }
 
-  public void removeAllTenantMonitors(String tenantId, boolean sendEvents) {
+  @Async
+  public CompletableFuture<Void> removeAllTenantMonitors(String tenantId, boolean sendEvents) {
```
itzg, line 2270: Since there was request to add `JobStatus.IN_PROGRESS` it seems like the job entry should be created here and updated in `saveJobResults`.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2308,4 +2323,23 @@ public ResourceDTO findResourceByTenantIdAndResourceId(String tenantId, String r
...
+  /**
+   * Saves the Job results in jobs table.
+   *
+   * @param tenantId
+   * @param jobType
+   * @param jobStatus
+   * @param description
+   */
+  public void saveJobResults(String tenantId, JobType jobType, JobStatus jobStatus, String description) {
+    String id = tracer.currentSpan().context().traceIdString();
```
itzg, line 2336: Since `tracer.currentSpan().context()` probably depends on thread-local storage, I worry about this being unreliable in an async scenario. It also seems like `MonitorApiController` should pass the trace ID to ensure that web-related details like that are abstracted away from this service layer code.

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
```diff
@@ -372,8 +374,19 @@ public void delete(@PathVariable String tenantId,
 
   @DeleteMapping("/admin/tenant/{tenantId}/monitors")
   @ApiOperation("Deletes all monitors for a particular tenant")
-  public void deleteAllTenantMonitors(@PathVariable String tenantId, @RequestParam(defaultValue = "true") boolean sendEvents) {
-    monitorManagement.removeAllTenantMonitors(tenantId, sendEvents);
+  public void deleteAllTenantMonitors(
+      @PathVariable String tenantId, @RequestParam(defaultValue = "true") boolean sendEvents) {
+
+    monitorManagement.removeAllTenantMonitors(tenantId, sendEvents)
```
itzg, line 380: ...if traceId is retrieved prior to these calls then it would be guaranteed to grab the proper web request's thread context.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -342,7 +346,8 @@ public void testGetMonitor() {
   @Test
   public void testGetMonitorForUnauthorizedTenant() {
     String unauthorizedTenantId = RandomStringUtils.randomAlphanumeric(10);
-    Optional<Monitor> monitor = monitorManagement.getMonitor(unauthorizedTenantId, currentMonitor.getId());
+    Optional<Monitor> monitor = monitorManagement
+        .getMonitor(unauthorizedTenantId, currentMonitor.getId());
```
itzg, line 350: Try to avoid running a reformat on the entire file. 247 lines changed in this file and only the three or so for the mock `tracer` seem to be necessary.

rohitsngh27: Yes I noticed that but thought it would be good to keep the file aligned with our indentation style

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
rohitsngh27: So I can call `saveJobResults` method first and create another method as `updateJobResults` which will get the job details and update it or simply call the `saveJobResults` method again.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
itzg: Yes, your intention is good, but keeping diffs small is slightly more important. For one thing, when doing a git blame on a line of code, it avoids having to go back a couple of revisions to find the original change/author.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
itzg: Yes and yes, there could be some shared functionality between the two methods.

itzg (dismissed):
Looks good. Wait for at least @jjbuchan 's re-review before merging.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2308,4 +2319,22 @@ public ResourceDTO findResourceByTenantIdAndResourceId(String tenantId, String r
...
+   * @param description
+   */
+  public void saveJobResults(String id, String tenantId, JobType jobType, JobStatus jobStatus, String description) {
+    Job job = new Job()
+        .setId(id)
+        .setType(jobType)
+        .setStatus(jobStatus)
+        .setDescription(description)
+        .setTenantId(tenantId);
+    jobRepository.save(job);
```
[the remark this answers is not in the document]
rohitsngh27: Yes, I have tested the scenario and save method is taking care of update operation too.

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
```diff
@@ -372,8 +378,21 @@ public void delete(@PathVariable String tenantId,
 
   @DeleteMapping("/admin/tenant/{tenantId}/monitors")
   @ApiOperation("Deletes all monitors for a particular tenant")
-  public void deleteAllTenantMonitors(@PathVariable String tenantId, @RequestParam(defaultValue = "true") boolean sendEvents) {
-    monitorManagement.removeAllTenantMonitors(tenantId, sendEvents);
+  public void deleteAllTenantMonitors(
+      @PathVariable String tenantId, @RequestParam(defaultValue = "true") boolean sendEvents) {
+    String id = tracer.currentSpan().context().traceIdString();
```
jjbuchan, line 383: I looked into the chance of this ID having duplicates over time and found this thread - https://github.com/spring-cloud/spring-cloud-sleuth/issues/936

In summary we should be good, and if we happen to see any dupes we can set a property to increase the size of them.

jjbuchan (commented):
I just realized all your changes so far only affect monitor-management.  Will there be an admin API change as well to reflect that the initial call is async?  i.e. When making an admin api call I'd expect to get back the `202` response.

In the current context, naming the job `JobType.DELETE_TENANT` doesn't feel entirely accurate as it only really relates to deleting all monitors for a tenant.  If the job is being tracked at the individual service instead of the admin api, I'd suggest we change it to `DELETE_TENANT_MONITORS`.

I'd also expect some test to have to change in monitor-management to ensure the behavior is working as expected.  At minimum verifying a 202 is returned would be good, but verifying an async request actually happens would be better.  Can you look into that?

rohitsngh27:
> I just realized all your changes so far only affect monitor-management. Will there be an admin API change as well to reflect that the initial call is async? i.e. When making an admin api call I'd expect to get back the `202` response.
> 
> In the current context, naming the job `JobType.DELETE_TENANT` doesn't feel entirely accurate as it only really relates to deleting all monitors for a tenant. If the job is being tracked at the individual service instead of the admin api, I'd suggest we change it to `DELETE_TENANT_MONITORS`.
> 
> I'd also expect some test to have to change in monitor-management to ensure the behavior is working as expected. At minimum verifying a 202 is returned would be good, but verifying an async request actually happens would be better. Can you look into that?

Do we need to return 202 response from monitor-management api as well?

jjbuchan:
> Do we need to return 202 response from monitor-management api as well?

I'd say so.  It's the code that makes sense for async requests.

rohitsngh27:
Added few unit test cases in controller and also added new method to update job results as previous approach was deleting the created timestamp

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
```diff
@@ -378,21 +379,23 @@ public void delete(@PathVariable String tenantId,
...
+                .updateJobResults(id, JobStatus.SUCCESS, null);
           } else {
-            monitorManagement.saveJobResults(id, tenantId, JobType.DELETE_TENANT, JobStatus.FAILURE,
-                throwable.getMessage());
+            monitorManagement
+                .updateJobResults(id, JobStatus.FAILURE,
+                    throwable.getMessage());
           }
         });
+    return new ResponseEntity<>(HttpStatus.ACCEPTED);
```
itzg, line 398: `@ResponseStatus` on the method declaration also works.

itzg (dismissed):

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
jjbuchan: +1 to using the annotation for status instead of having it here

jjbuchan (commented):
One small change that would be nice, but all looks good otherwise.

jjbuchan (approved):

itzg:
@harageth , I remember now, this was part of what we hoped would fix that delete tenant issue that was found just before your leave.

PR #222, opened 2020-10-23
Added apis to get detached but pending timeout pollers

rohitsngh27:
# Resolves
https://jira.rax.io/browse/SALUS-992

# What
Added apis to get detached but pending timeout pollers

# How
Added 2 new apis for public and admin to get the list of detached but pending timeout pollers.

# How to test
This can be tested via Rest call.

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -22,7 +22,6 @@ import static com.rackspace.salus.telemetry.entities.Resource.REGION_METADATA;
 import static com.rackspace.salus.telemetry.etcd.types.ResolvedZone.resolveZone;
 import static org.springframework.util.CollectionUtils.isEmpty;
 
-import brave.Tracer;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.google.common.collect.Streams;
 import com.google.common.math.Stats;
@@ -73,6 +72,7 @@ import com.rackspace.salus.telemetry.repositories.JobRepository;
 import com.rackspace.salus.telemetry.repositories.MonitorPolicyRepository;
 import com.rackspace.salus.telemetry.repositories.MonitorRepository;
 import com.rackspace.salus.telemetry.repositories.ResourceRepository;
+import com.rackspace.salus.telemetry.repositories.ZoneRepository;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.MeterRegistry;
 import java.io.IOException;
@@ -91,6 +91,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
@@ -150,6 +151,8 @@ public class MonitorManagement {
 
   JobRepository jobRepository;
 
+  ZoneRepository zoneRepository;
+
   // metrics counters
   private final Counter.Builder orphanedBoundMonitorRemoved;
   private final Counter.Builder policyIntegrityErrors;
@@ -173,7 +176,8 @@ public class MonitorManagement {
       MonitorConversionService monitorConversionService,
       MetadataUtils metadataUtils, MeterRegistry meterRegistry,
       JdbcTemplate jdbcTemplate,
-      JobRepository jobRepository)
+      JobRepository jobRepository,
+      ZoneRepository zoneRepository)
       throws IOException {
     this.resourceRepository = resourceRepository;
     this.monitorPolicyRepository = monitorPolicyRepository;
@@ -195,6 +199,7 @@ public class MonitorManagement {
     this.labelMatchOrQuery = SpringResourceUtils.readContent("sql-queries/monitor_label_matching_OR_query.sql");
     this.jobRepository = jobRepository;
     this.meterRegistry = meterRegistry;
+    this.zoneRepository = zoneRepository;
     createMonitorSuccess = Counter.builder(MetricNames.SERVICE_OPERATION_SUCCEEDED).tag(
         MetricTags.SERVICE_METRIC_TAG,"MonitorManagement");
     orphanedBoundMonitorRemoved = Counter.builder("orphaned").tag(MetricTags.SERVICE_METRIC_TAG,"MonitorManagement");
@@ -2096,18 +2101,36 @@ public class MonitorManagement {
       @Nullable String zoneTenantId, String zoneName) {
 
     final ResolvedZone zone = resolveZone(zoneTenantId, zoneName);
+    ZoneAllocationResolver zoneAllocationResolver = zoneAllocationResolverFactory.create();
 
-    return zoneAllocationResolverFactory.create()
-        .getZoneBindingCounts(zone)
-        .thenApply(bindingCounts ->
-            bindingCounts.entrySet().stream()
-                .map(entry ->
-                    new ZoneAssignmentCount()
-                        .setResourceId(entry.getKey().getResourceId())
-                        .setEnvoyId(entry.getKey().getEnvoyId())
-                        .setAssignments(entry.getValue()))
-                .collect(Collectors.toList()));
+    return zoneAllocationResolver.getActiveZoneBindingCounts(zone)
+        .thenApply(bindingCounts -> getZoneAssingmentPollerCount(bindingCounts, true))
+        .thenApply(pollerCounts -> {
+          pollerCounts.addAll(getZoneAssingmentPollerCount(
+              zoneAllocationResolver.getExpiredZoneBindingCounts(zone).join(), false));
+          return pollerCounts;
+        });
   }
+
+  /**
+   * This method maps the bindingcount map to the ZoneAssignmentCount list.
+   *
+   * @param bindingCounts
+   * @param isConnected
+   * @return
+   */
+  public List<ZoneAssignmentCount> getZoneAssingmentPollerCount(
+      Map<EnvoyResourcePair, Integer> bindingCounts, boolean isConnected) {
+    return bindingCounts.entrySet().stream()
+        .map(entry ->
+            new ZoneAssignmentCount()
+                .setResourceId(entry.getKey().getResourceId())
+                .setEnvoyId(entry.getKey().getEnvoyId())
+                .setAssignments(entry.getValue())
+                .setConnected(isConnected))
+        .collect(Collectors.toList());
+  }
+
 
   /**
    * Rebalances a zone by re-assigning bound monitors from envoys that are over-assigned
@@ -2122,7 +2145,7 @@ public class MonitorManagement {
     log.info("Rebalancing zone={}", zone);
 
     return zoneAllocationResolverFactory.create()
-        .getZoneBindingCounts(zone)
+        .getActiveZoneBindingCounts(zone)
         .thenApply(bindingCounts ->
             rebalanceWithZoneBindingCounts(zone, bindingCounts)
         );
@@ -2351,5 +2374,27 @@ public class MonitorManagement {
     job.setStatus(jobStatus);
     job.setDescription(description);
     jobRepository.save(job);
+  }
+
+  /**
+   * Gets the zone assignment count for all zones of a tenant.
+   *
+   * @param tenantId
+   * @return
+   */
+  public CompletableFuture<Map<String, List<ZoneAssignmentCount>>> getAssignmentCountForTenant(
+      String tenantId) {
+    Page<String> page = zoneRepository.findAllZoneNameForTenant(tenantId, Pageable.unpaged());
+    if (page.isEmpty()) {
+      throw new NotFoundException("No zones present for tenant %s" + tenantId);
+    }
+    Map<String, List<ZoneAssignmentCount>> zoneAssignmentCountMap = new HashMap<>();
+    page.stream().forEach(zone -> {
+      List<ZoneAssignmentCount> assignmentCounts = getZoneAssignmentCounts(tenantId, zone).join();
+      if (!assignmentCounts.isEmpty()) {
+        zoneAssignmentCountMap.put(zone, assignmentCounts);
+      }
+    });
+    return CompletableFuture.completedFuture(zoneAssignmentCountMap);
   }
 }
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -2106,6 +2106,13 @@ public class MonitorManagement {
     handleNewEnvoyInZone(zoneTenantId, zoneName);
   }
 
+  /**
+   * Returns the list of ZoneAssignmentCount for a particular zone of a tenant.
+   *
+   * @param zoneTenantId
+   * @param zoneName
+   * @return
+   */
   public CompletableFuture<List<ZoneAssignmentCount>> getZoneAssignmentCounts(
       @Nullable String zoneTenantId, String zoneName) {
 
@@ -2113,23 +2120,24 @@ public class MonitorManagement {
     ZoneAllocationResolver zoneAllocationResolver = zoneAllocationResolverFactory.create();
 
     return zoneAllocationResolver.getActiveZoneBindingCounts(zone)
-        .thenApply(activePollerMap -> getZoneAssignmentPollerCount(activePollerMap, true))
+        .thenApply(activePollerMap -> mapToZoneAssignmentCount(activePollerMap, true))
         .thenCompose(assignmentCounts -> zoneAllocationResolver.getExpiringZoneBindingCounts(zone)
             .thenApply(envoyResourcePairIntegerMap -> {
               assignmentCounts
-                  .addAll(getZoneAssignmentPollerCount(envoyResourcePairIntegerMap, false));
+                  .addAll(mapToZoneAssignmentCount(envoyResourcePairIntegerMap, false));
               return assignmentCounts;
             }));
   }
 
   /**
-   * This method maps the bindingcount map to the ZoneAssignmentCount list.
+   * This method is used to convert the map of EnvoyResourcePair count to list of
+   * ZoneAssignmentCount.
    *
    * @param bindingCounts
    * @param isConnected
    * @return
    */
-  public List<ZoneAssignmentCount> getZoneAssignmentPollerCount(
+  public List<ZoneAssignmentCount> mapToZoneAssignmentCount(
       Map<EnvoyResourcePair, Integer> bindingCounts, boolean isConnected) {
     return bindingCounts.entrySet().stream()
         .map(entry ->
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -91,7 +91,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.annotation.Nullable;
@@ -2114,12 +2113,13 @@ public class MonitorManagement {
     ZoneAllocationResolver zoneAllocationResolver = zoneAllocationResolverFactory.create();
 
     return zoneAllocationResolver.getActiveZoneBindingCounts(zone)
-        .thenApply(bindingCounts -> getZoneAssingmentPollerCount(bindingCounts, true))
-        .thenApply(pollerCounts -> {
-          pollerCounts.addAll(getZoneAssingmentPollerCount(
-              zoneAllocationResolver.getExpiredZoneBindingCounts(zone).join(), false));
-          return pollerCounts;
-        });
+        .thenApply(activePollerMap -> getZoneAssignmentPollerCount(activePollerMap, true))
+        .thenCompose(assignmentCounts -> zoneAllocationResolver.getExpiringZoneBindingCounts(zone)
+            .thenApply(envoyResourcePairIntegerMap -> {
+              assignmentCounts
+                  .addAll(getZoneAssignmentPollerCount(envoyResourcePairIntegerMap, false));
+              return assignmentCounts;
+            }));
   }
 
   /**
@@ -2129,7 +2129,7 @@ public class MonitorManagement {
    * @param isConnected
    * @return
    */
-  public List<ZoneAssignmentCount> getZoneAssingmentPollerCount(
+  public List<ZoneAssignmentCount> getZoneAssignmentPollerCount(
       Map<EnvoyResourcePair, Integer> bindingCounts, boolean isConnected) {
     return bindingCounts.entrySet().stream()
         .map(entry ->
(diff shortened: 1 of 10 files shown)


itzg (changes requested):
Great changes especially with the unit tests. Speaking of which, can you add one more unit test for the "tenant wide" private zone call?

rohitsngh27:
> Great changes especially with the unit tests. Speaking of which, can you add one more unit test for the "tenant wide" private zone call?

I have added that in line number 794. I guess that's what you are talking about?

itzg:
> > Great changes especially with the unit tests. Speaking of which, can you add one more unit test for the "tenant wide" private zone call?
> 
> I have added that in line number 794. I guess that's what you are talking about?

I'm sorry, I didn't notice there were two cases in one. Thanks!

itzg (approved):

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
```diff
@@ -259,4 +263,19 @@ public void delete(HttpServletRequest request) {
   public void deleteTenantZones(@PathVariable String tenantId, @RequestParam(defaultValue = "true") boolean sendEvents) {
         zoneManagement.removeAllTenantZones(tenantId, sendEvents);
   }
+
+  @GetMapping("/admin/detached-pollers/**")
```
jjbuchan, line 267: Following on from this thread - https://github.com/racker/salus-telemetry-zone-watcher/pull/24#discussion_r510354354

I'm not sure why we're limiting the endpoint to detached pollers and not using it to query all pollers?

rohitsngh27: As mentioned by @itzg in here :- https://github.com/racker/salus-telemetry-monitor-management/blob/d241afc0d4be2e6269a41d9a3126730fe66c2086/src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java#L137 
This gives us the count of connected pollers with some more details I guess.

jjbuchan: I see `/zone-assignment-counts/{name}` returns items in this format:

```
[
  {
    "resourceId": "r-1",
    "envoyId": "e-1",
    "assignments": 3
  }
]
```

and `/detached-pollers/{name}` returns a list of resource strings.

I do still wonder if these two endpoints should be combined.

We could update the original object to contain a `status` field to show `connected` and `expiring` or something like that and be able to filter on what we want to see.

I think the `expiring` pollers would also have an `assignments` value until they have fully disconnected.  

@itzg wouldn't we want to know how many monitors are assigned to an `expiring` poller the same way we do for an `active` one?  Is that info not being available not part of the problem we have right now?

itzg: `/zone-assignment-counts` will include the pending-expiring ones also since it merges the information from both etcd (where they'd be absent) and the bound monitors table (where they'd be present).

> Is that info not being available not part of the problem we have right now?

I don't know of any remaining problem. Last known problems I knew about were fixed in https://github.com/racker/salus-telemetry-monitor-management/pull/205 and https://github.com/racker/salus-telemetry-monitor-management/pull/208

itzg: ...BTW the original purpose of this issue was simply to help QE verify the test case dealing with stopping/starting envoy-pollers and avoid having to query etcd directly to confirm the intermediate state.

itzg (approved):
I'm still good with the changes, but being the original requestor I might be "blind" to a use case that @jjbuchan sees is missing.

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
jjbuchan: > I don't know of any remaining problem

I really mean "missing feature" vs. "problem".

> the original purpose of this issue was simply to help QE verify the test case dealing with stopping/starting envoy-pollers and avoid having to query etcd directly to confirm the intermediate state.

The problem I'm having with these changes is that if an endpoint already exists that knows (or at least can know) all the info why are we adding a completely separate one vs. just enhancing the original?

It sounds like adding a `status`/`state` field to the existing endpoint and adding filtering via query parameters solves QE's need?

My concerns are from a usability/training/maintenance perspective.  Why have two endpoints that return things in different formats when we can have one that does everything and returns things in a consistent manner.

itzg:
> It sounds like adding a status/state field to the existing endpoint and adding filtering via query parameters solves QE's need?

This sounds like a reasonable compromise, but rather than keep dragging this out for @rohitsngh27 we should go back to the drawing board and write up clear requirements.

itzg:
...and by "we" I mean I'll add a comment to the Jira issue that we can iterate upon with the requirements.

rohitsngh27:
How should I proceed with this PR now? Do I need to make any changes or merge it as is?

itzg:
> How should I proceed with this PR now? Do I need to make any changes or merge it as is?

I think you can use this PR to convey the next iteration since the changes should remain mostly in the same areas of code.

@jjbuchan @rohitsngh27 , I was hoping gist's allowed for inline comments, but they don't so maybe adding comments at the bottom of https://gist.github.com/itzg/d334f9b3aea2e03e055fad71a5de2ff4 could be a way to review the new requirements.

rohitsngh27:
> I think you can use this PR to convey the next iteration since the changes should remain mostly in the same areas of code.

So, Should I make the changes in existing `getPrivateZoneAssignmentCounts` and `getPublicZoneAssignmentCounts` as mentioned by @jjbuchan ?

rohitsngh27:
Initial commit with new changes to check if it's fine. Will update/add unit test cases later

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
```diff
@@ -141,13 +142,19 @@ private String extractPublicZoneNameFromUri(HttpServletRequest request) {
   @GetMapping("/tenant/{tenantId}/zone-assignment-counts/{name}")
   @ApiOperation(value = "Gets assignment counts of monitors to poller-envoys in the private zone")
   public CompletableFuture<List<ZoneAssignmentCount>> getPrivateZoneAssignmentCounts(
-      @PathVariable String tenantId, @PathVariable @PrivateZoneName String name) {
+      @PathVariable String tenantId, @PathVariable @PrivateZoneName Optional<String> name) {
```
itzg, line 145: I think this shouldn't be `Optional` since there's now the separate `getPrivateZoneAssignmentCountsPerTenant`

itzg (changes requested):
Looking good overall

src/main/java/com/rackspace/salus/monitor_management/services/ZoneAllocationResolver.java
```diff
@@ -138,44 +172,14 @@ public ZoneAllocationResolver(ZoneStorage zoneStorage, EntityManager entityManag
...
-          if (zone.isPublicZone()) {
-            query = entityManager
-                .createNamedQuery("BoundMonitor.publicPollerLoading", Tuple.class)
-                .setParameter("zoneName", zone.getName());
-          } else {
-            query = entityManager
-                .createNamedQuery("BoundMonitor.privatePollerLoading", Tuple.class)
-                .setParameter("tenantId", zone.getTenantId())
-                .setParameter("zoneName", zone.getName());
+          if (!activePollerResources.isEmpty()) {
```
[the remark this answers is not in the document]
rohitsngh27: Actually for the case when lets say an envoy with Zone-> `zone1` and Resource-> `resource1`and the envoy is currently in expiring stage(pending timeout).
So, Ideally the response should only show the `connected=false` one but if I remove the empty check from here it gets the data from BoundMonitor where it would be present and set that into the map for activePollersCount.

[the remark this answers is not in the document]
itzg: That makes sense for `retrieveExpiredPollerResourceCounts` where it is returning only `connected=false` ones, but for `retrieveActivePollerResourceCounts` it seems like it should always consider the bound monitors counts via SQL.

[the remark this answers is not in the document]
rohitsngh27: So an envoy is connected it will show 2 same entry in final response with difference being connected = true and false like:-
`[
  {
    "resourceId": "test11",
    "envoyId": "1fa0b59e-193b-11eb-a876-acde48001122",
    "assignments": 20,
    "connected": true
  },
  {
    "resourceId": "test11",
    "envoyId": "1fa0b59e-193b-11eb-a876-acde48001122",
    "assignments": 20,
    "connected": false
  }
]`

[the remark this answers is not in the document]
jjbuchan: This conversation is confusing me.  I'm going to keep going through the rest of the code and come back to this, but so far I'm not sure if I agree with

> for `retrieveActivePollerResourceCounts` it seems like it should always consider the bound monitors counts via SQL.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2105,19 +2111,37 @@ public void handleExpiredEnvoy(@Nullable String zoneTenantId, String zoneName, S
...
-                    new ZoneAssignmentCount()
-                        .setResourceId(entry.getKey().getResourceId())
-                        .setEnvoyId(entry.getKey().getEnvoyId())
-                        .setAssignments(entry.getValue()))
-                .collect(Collectors.toList()));
+    return zoneAllocationResolver.getActiveZoneBindingCounts(zone)
+        .thenApply(bindingCounts -> getZoneAssingmentPollerCount(bindingCounts, true))
+        .thenApply(pollerCounts -> {
+          pollerCounts.addAll(getZoneAssingmentPollerCount(
+              zoneAllocationResolver.getExpiredZoneBindingCounts(zone).join(), false));
```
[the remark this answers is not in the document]
rohitsngh27: `thenAccept` will return void but I need the result for `zoneAllocationResolver.getExpiredZoneBindingCounts(zone)` to add to the final list response.

src/main/java/com/rackspace/salus/monitor_management/services/ZoneAllocationResolver.java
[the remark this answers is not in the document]
itzg: Thanks, that helped to clarify. I was mainly concerned about breaking the bound monitor processing, but I'm now thinking even for that it should only consider active envoy-pollers anyway.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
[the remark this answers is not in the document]
itzg: Sorry, I meant `thenCompose`

src/main/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiController.java
```diff
@@ -137,13 +142,19 @@ private String extractPublicZoneNameFromUri(HttpServletRequest request) {
...
 
-    if (!zoneManagement.exists(tenantId, name)) {
-      throw new NotFoundException(String.format("No private zone found named %s", name));
-    }
+    String zoneName = name.orElse("");
```
jjbuchan, line 147: Adding a note to remind myself to come back and look at this.

Based on the above comment, this logic should change.

src/test/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiControllerTest.java
```diff
@@ -741,4 +746,63 @@ public void testCreatePublicZoneInvalidPublicZoneName() throws Exception {
...
+
+    @Test
+    public void testGetExpiredPollerPublicZone() throws Exception {
+
+        when(zoneStorage.getExpiredPollerResourceIdsInZone(any()))
+            .thenReturn(CompletableFuture.completedFuture(List.of("r-1", "r-2")));
+
+        ResolvedZone resolvedZone = ResolvedZone.createPublicZone("public/testPublicZone");
+        final MvcResult result = mvc.perform(
+            get("/api/admin/detached-pollers/{name}",
```
jjbuchan, line 758: Looks like tests still need to be updated.

rohitsngh27: I will update all the tests. Right now only focusing on functionality first

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2361,4 +2385,26 @@ public void updateJobResults(String id, JobStatus jobStatus, String description)
...
+  public CompletableFuture<Map<String, List<ZoneAssignmentCount>>> getAssignmentCountForTenant(
+      String tenantId) {
+    Page<String> page = zoneRepository.findAllZoneNameForTenant(tenantId, Pageable.unpaged());
+    if (page.isEmpty()) {
+      throw new NotFoundException("No zones present for tenant %s" + tenantId);
+    }
+    Map<String, List<ZoneAssignmentCount>> zoneAssignmentCountMap = new HashMap<>();
+    page.stream().forEach(zone -> {
+      List<ZoneAssignmentCount> assignmentCounts = getZoneAssignmentCounts(tenantId, zone).join();
+      if (!assignmentCounts.isEmpty()) {
```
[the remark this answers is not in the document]
rohitsngh27: If I remove this it does show all the zones but ones having no envoys attached will come as empty array

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2361,4 +2385,26 @@ public void updateJobResults(String id, JobStatus jobStatus, String description)
...
     jobRepository.save(job);
   }
+
+  /**
+   * Gets the zone assignment count for all zones of a tenant.
+   *
+   * @param tenantId
+   * @return
+   */
+  public CompletableFuture<Map<String, List<ZoneAssignmentCount>>> getAssignmentCountForTenant(
```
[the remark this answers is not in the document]
rohitsngh27: Any reference on this one? 
I was thinking of first all all the completable future returning from `getZoneAssignmentCounts ` in a list of completable future and then using `CompletableFuture.allOf`.
But this approach is making code more complex as we also need to take care of the key(zone name) which will be used in final response map

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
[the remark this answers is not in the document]
jjbuchan: Cool.  I think that's what I'd expected.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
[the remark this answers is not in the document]
itzg: Yeah, merging all those completables into the final map will indeed get complicated. I'm definitely fine with how you have it.

rohitsngh27:
Resolved comments and added unit test cases too

src/main/java/com/rackspace/salus/monitor_management/services/ZoneAllocationResolver.java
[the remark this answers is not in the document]
rohitsngh27: Do I need to remove the empty check here in both active or expired one?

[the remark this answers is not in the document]
jjbuchan: From what I understand I think it should stay.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2114,12 +2113,13 @@ public void handleExpiredEnvoy(@Nullable String zoneTenantId, String zoneName, S
     ZoneAllocationResolver zoneAllocationResolver = zoneAllocationResolverFactory.create();
```
jjbuchan, line 2113: I can't add a comment high enough up, but can you add some javadoc to this method?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2129,7 +2129,7 @@ public void handleExpiredEnvoy(@Nullable String zoneTenantId, String zoneName, S
    * @param isConnected
    * @return
    */
-  public List<ZoneAssignmentCount> getZoneAssingmentPollerCount(
+  public List<ZoneAssignmentCount> getZoneAssignmentPollerCount(
```
jjbuchan, line 2132: This javadoc maybe needs to be expanded?

We have `getZoneAssignmentCounts` and `getZoneAssignmentPollerCount` and it's not immediately clear what the difference is without looking through the code in each.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagement_ZoneBindingsTest.java
```diff
@@ -1256,4 +1273,77 @@ private void assertBindings(UUID monitorId, BoundMonitorMatcher... matchers) {
...
+    Map<EnvoyResourcePair, Integer> activeZoneCountsForZone1 = new HashMap<>();
+    activeZoneCountsForZone1.put(new EnvoyResourcePair().setResourceId("r-1").setEnvoyId("e-1"), 5);
+    activeZoneCountsForZone1.put(new EnvoyResourcePair().setResourceId("r-2").setEnvoyId("e-2"), 6);
+
+    Map<EnvoyResourcePair, Integer> activeZoneCountsForZone2 = new HashMap<>();
+    activeZoneCountsForZone2.put(new EnvoyResourcePair().setResourceId("r-3").setEnvoyId("e-3"), 0);
+    activeZoneCountsForZone2.put(new EnvoyResourcePair().setResourceId("r-4").setEnvoyId("e-4"), 1);
+
+    when(zoneRepository.findAllZoneNameForTenant("t-1", Pageable.unpaged()))
+        .thenReturn(new PageImpl<>(List.of("z-1", "z-2")));
```
jjbuchan, line 1291: If we add a `z-3` in here will it simulate a zone connected with no pollers and therefore show as an empty list?

An assertion on `counts.get("z-3")` isEmpty or hasSize(0) would be good to cover that scenario.

src/test/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiControllerTest.java
```diff
@@ -748,61 +750,69 @@ public void testCreatePublicZoneInvalidPublicZoneName() throws Exception {
...
 
-        when(zoneStorage.getExpiredPollerResourceIdsInZone(any()))
-            .thenReturn(CompletableFuture.completedFuture(List.of("r-1", "r-2")));
+        Map<String, List<ZoneAssignmentCount>> expected = Map.of("z-1", Collections.singletonList(
+            new ZoneAssignmentCount().setResourceId("r-1").setEnvoyId("e-1").setAssignments(3)
+                .setConnected(true)),
+            "z-2", Collections.singletonList(
+                new ZoneAssignmentCount().setResourceId("r-2").setEnvoyId("e-2").setAssignments(1)
+                    .setConnected(false)),
+            "z-3", Collections.emptyList());
```
jjbuchan, line 761: Ah yeah, this is the scenario I'm talking about above.  I think the above changes might need to be done to ensure this mock is accurate.

jjbuchan (commented):
Sorry to drag this out a little more but if you can look through all the methods used here and tweak the javadoc where needed that'll help the next person who has to look at this code.

Thanks!

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagement_ZoneBindingsTest.java
rohitsngh27: I have used a json file where `z-3` is set to empty list. That would cover this I guess?

jjbuchan: Not quite.  That is testing a different layer of the service.

In that test you are mocking the response from monitorManagement.

```
        when(monitorManagement.getZoneAssignmentCountForTenant(any()))
            .thenReturn(CompletableFuture.completedFuture(expected));
```

I'm suggesting we add the empty `z-3 ` to this test as that will verify the above `when` assumption is correct.

rohitsngh27: Ohhhh Got it... You mean to add the case when zone has no pollers(active or expiring).

src/test/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiControllerTest.java
rohitsngh27: This one is fine. Right?

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagement_ZoneBindingsTest.java
jjbuchan: Yep, that's it

src/test/java/com/rackspace/salus/monitor_management/web/controller/ZoneApiControllerTest.java
jjbuchan: yeah, all good here.  just noticed it was related to the one i was talking about above.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2101,23 +2106,50 @@ public void handleExpiredEnvoy(@Nullable String zoneTenantId, String zoneName, S
...
 
+  /**
+   * This method is used to convert the map of EnvoyResourcePair count to list of
+   * ZoneAssignmentCount.
+   *
+   * @param bindingCounts
+   * @param isConnected
+   * @return
+   */
+  public List<ZoneAssignmentCount> mapToZoneAssignmentCount(
```
jjbuchan, line 2140: One final thing!  Looks like this method could be package private?  In addition to the updated javadoc I think that would help show this is more of a helper method than something that is called directly by other services.

```suggestion
  List<ZoneAssignmentCount> mapToZoneAssignmentCount(
```

rohitsngh27: Ohh yes, Sorry to miss out on that.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagement_ZoneBindingsTest.java
```diff
@@ -1256,4 +1273,84 @@ private void assertBindings(UUID monitorId, BoundMonitorMatcher... matchers) {
...
+    activeZoneCountsForZone2.put(new EnvoyResourcePair().setResourceId("r-3").setEnvoyId("e-3"), 0);
+    activeZoneCountsForZone2.put(new EnvoyResourcePair().setResourceId("r-4").setEnvoyId("e-4"), 1);
+
+    when(zoneRepository.findAllZoneNameForTenant("t-1", Pageable.unpaged()))
+        .thenReturn(new PageImpl<>(List.of("z-1", "z-2", "z-3")));
+
+    when(zoneAllocationResolver.getActiveZoneBindingCounts(any()))
+        .thenReturn(CompletableFuture.completedFuture(activeZoneCountsForZone1))
+        .thenReturn(CompletableFuture.completedFuture(activeZoneCountsForZone2))
+        .thenReturn(CompletableFuture.completedFuture(Collections.emptyMap()));
```
itzg, line 1297: This approach is fine since the ordering is ensured by previous mock response, but FYI you can always use exact argument matchers on `when` declarations to narrow the mock response for each.

jjbuchan (approved):
Thanks for putting up with all the comments!

itzg (approved):

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagement_ZoneBindingsTest.java
rohitsngh27: I was doing that way only first but somehow it was not working and giving null pointer in original code where the mocked method is called.

itzg: Strange. Maybe the argument matching equality wasn't quite as expected. Again, what you have is fine and thanks for trying the other way initially.

PR #224, opened 2020-10-28
Rename PolicyMonitor to MonitorTemplate

shabd67:
# Resolves

[SALUS-1035](https://jira.rax.io/browse/SALUS-1035)

# What

Renamed PolicyMonitor to MonitorTemplate

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1387,7 +1387,7 @@ public class MonitorManagement {
       throw new NotFoundException(String.format("No monitor template found for %s", id));
     }
 
-    if (monitorPolicyRepository.existsByMonitorId(id)) {
+    if (monitorPolicyRepository.existsByMonitorTemplateId(id)) {
       throw new DeletionNotAllowedException("Cannot remove monitor that is in use by a policy");
     }
     monitorRepository.deleteById(id);
@@ -1404,7 +1404,7 @@ public class MonitorManagement {
     if (policy != null) {
       // if policy exists do not trust the monitorId in the event is up to date
       // delays in processing can occur.
-      event.setMonitorId(policy.getMonitorId());
+      event.setMonitorId(policy.getMonitorTemplateId());
     }
     if (event.getMonitorId() == null) {
       handlePolicyOptOutEvent(event);
@@ -1462,7 +1462,7 @@ public class MonitorManagement {
           return null;
         }))
         .filter(Objects::nonNull)
-        .filter(p -> p.getMonitorId().equals(monitorId))
+        .filter(p -> p.getMonitorTemplateId().equals(monitorId))
         .map(p -> monitorRepository.findByTenantIdAndPolicyId(tenantId, p.getId()).orElse(null))
         .filter(Objects::nonNull)
         .collect(Collectors.toList());
@@ -1673,7 +1673,7 @@ public class MonitorManagement {
     for (UUID policyId : policiesToAdd) {
       Optional<MonitorPolicy> policy = monitorPolicyRepository.findById(policyId);
       policy.ifPresent(
-          monitorPolicy -> cloneMonitorTemplate(tenantId, policyId, monitorPolicy.getMonitorId()));
+          monitorPolicy -> cloneMonitorTemplate(tenantId, policyId, monitorPolicy.getMonitorTemplateId()));
     }
   }
 
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -216,12 +216,12 @@ public class MonitorManagement {
   }
 
   /**
-   * Gets an individual policy monitor by id
+   * Gets an individual monitor template by id
    * @param id The unique value representing the monitor.
    * @return The monitor with the provided id.
    * @throws NotFoundException If the monitor does not exist under the POLICY tenant.
    */
-  public Optional<Monitor> getPolicyMonitor(UUID id) {
+  public Optional<Monitor> getMonitorTemplate(UUID id) {
     return getMonitor(POLICY_TENANT, id);
   }
 
@@ -239,9 +239,9 @@ public class MonitorManagement {
    * Get a selection of monitor objects associated to the policy tenant.
    *
    * @param page The slice of results to be returned.
-   * @return The policy monitors found that match the page criteria.
+   * @return The monitor templates found that match the page criteria.
    */
-  public Page<Monitor> getAllPolicyMonitors(Pageable page) {
+  public Page<Monitor> getAllMonitorTemplates(Pageable page) {
     return getMonitors(POLICY_TENANT, page);
   }
 
@@ -315,9 +315,9 @@ public class MonitorManagement {
   }
 
   /**
-   * Helper method for cloning a policy monitor.
+   * Helper method for cloning a monitor template.
    */
-  void clonePolicyMonitor(String tenantId, UUID policyId, UUID monitorId) {
+  void cloneMonitorTemplate(String tenantId, UUID policyId, UUID monitorId) {
     cloneMonitor(POLICY_TENANT, tenantId, monitorId, policyId);
   }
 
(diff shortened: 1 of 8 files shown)


itzg (commented):
It's outside the diff range, so I couldn't add a comment there: it seems like changing the request mapping paths like `/tenant/{tenantId}/policy-monitors` was important too.

Also, is this PR ready for full review? It's marked as draft.

shabd67:
> It's outside the diff range, so I couldn't add a comment there: it seems like changing the request mapping paths like `/tenant/{tenantId}/policy-monitors` was important too.
> 
> Also, is this PR ready for full review? It's marked as draft.

Yes, changing the request mapping paths in monitor-management along with public/admin apis part is still left. I wasn't sure about the changes to be done. Hence, created a draft PR. If its fine then will go ahead with other changes as well.

src/test/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiControllerTest.java
```diff
@@ -1254,4 +1255,82 @@ public void testDeleteAllTenantMonitorsException() throws Exception {
...
+//    mockMvc.perform(get("/api/tenant/{tenantId}/monitor-templates/{uuid}", "t-1", monitor.getId())
+//        .contentType(MediaType.APPLICATION_JSON))
+//        .andExpect(status().isOk())
+//        .andExpect(content()
+//            .contentTypeCompatibleWith(MediaType.APPLICATION_JSON))
+//        .andExpect(jsonPath("$.id", is(monitor.getId().toString())));
+//
+//    verify(monitorManagement).getMonitorUsingTemplatesForTenant(monitor.getId(), "t-1");
+//    verifyNoMoreInteractions(monitorManagement);
+
```
itzg, line 1313: Can be deleted?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2249,14 +2249,18 @@ private int rebalanceWithZoneBindingCounts(ResolvedZone zone,
...
+   * Retrieves a list of monitors using templates that are relevant to the provided tenant.
+   * @param tenantId The tenant to get the monitors using templates for.
    * @return A list of monitors.
    */
-  public Page<Monitor> getAllPolicyMonitorsForTenant(String tenantId, Pageable page) {
+  public Page<Monitor> getAllMonitorsUsingTemplatesForTenant(String tenantId, Pageable page) {
     return monitorRepository.findByTenantIdAndPolicyIdIsNotNull(tenantId, page);
   }
 
+  public Optional<Monitor> getMonitorUsingTemplatesForTenant(UUID uuid, String tenantId) {
```
itzg, line 2260: Seems like this method name doesn't need to reference "templates" and is more like `getMonitorForTenant`...which then makes me realize this method is probably redundant with `Optional<Monitor> getMonitor(String tenantId, UUID id)`

which I'm

itzg (commented):
Just some minor phrasing question/comments.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -2249,14 +2249,18 @@ private int rebalanceWithZoneBindingCounts(ResolvedZone zone,
...
+    return monitorRepository.findByIdAndTenantId(uuid, tenantId);
+  }
```
jjbuchan, line 2262: This method is a duplicate of `getMonitor`.  `getMonitor` can be used in any places you've used this.

src/test/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiControllerTest.java
```diff
@@ -1254,4 +1255,82 @@ public void testDeleteAllTenantMonitorsException() throws Exception {
...
+//
+//    mockMvc.perform(get("/api/tenant/{tenantId}/monitor-templates/{uuid}", "t-1", monitor.getId())
+//        .contentType(MediaType.APPLICATION_JSON))
+//        .andExpect(status().isOk())
+//        .andExpect(content()
+//            .contentTypeCompatibleWith(MediaType.APPLICATION_JSON))
+//        .andExpect(jsonPath("$.id", is(monitor.getId().toString())));
+//
+//    verify(monitorManagement).getMonitorUsingTemplatesForTenant(monitor.getId(), "t-1");
+//    verifyNoMoreInteractions(monitorManagement);
```
jjbuchan, line 1312: Need to remove this commented out section?

jjbuchan: Or remove the block below and uncomment this?

jjbuchan: Fyi, for debugging these kind of tests it can also be good to add a `.andDo(print())` into the `perform` request.  You'll see we have that in place on server of the controller tests.  It allows you to see the full result of the request instead of using `System.out.println` on individual parts.

jjbuchan (changes requested):
This looks good.  Just a couple small comments.

I'm assuming there's more PRs on the way for policy-management and docs?

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
jjbuchan: Not sure why your comments weren't showing up for me as I reviewed it.  Ended up duplicating a couple

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
```diff
@@ -153,75 +152,86 @@ public DetailedMonitorOutput getById(@PathVariable String tenantId,
     return monitorConversionService.convertToOutput(monitor);
   }
 
-  @GetMapping("/tenant/{tenantId}/policy-monitors")
-  @ApiOperation(value = "Gets all Policy Monitors for Tenant")
-  public PagedContent<DetailedMonitorOutput> getAllPolicyMonitorsForTenant(
+  @GetMapping("/tenant/{tenantId}/monitor-templates")
+  @ApiOperation(value = "Gets all Monitors using templates for Tenant")
```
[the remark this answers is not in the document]
jjbuchan: I think this is fine.  If I'm understanding your point correctly the difference is that you view "Monitor Templates" via the admin api call.  That call is returning the template that is not actually running anywhere - it's just a placeholder waiting to be cloned.

Once it is cloned to a tenant that new object is no longer a template, it's a real monitor just like all the others for that tenant.  The only difference it was created based on a template and remains tied to that template.  It is these cloned monitors or "monitors using templates" that is returned by the public api call.

src/test/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiControllerTest.java
shabd67: noted. will take care of that.

shabd67:
> This looks good. Just a couple small comments.
> 
> I'm assuming there's more PRs on the way for policy-management and docs?

@jjbuchan , do I need to make changes in policy-management as well? I think policy-management is responsible for monitor or metadata policy, not policy-monitor.

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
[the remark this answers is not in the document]
itzg: In that case it is adding to the confusion to use the same API path segment `/monitor-templates` to refer to both "monitor templates" (via admin API) and "monitor using template" (via public API).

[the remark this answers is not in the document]
jjbuchan: Ah yes, I think I overlooked that.  I'd say we should get rid of this endpoint.  @shabd67 can you confirm it's not used by anything and then remove it from here and the public api?

Similar to what Rohit was working on, I think if we want to filter based on using policy or not, we can add a query parameter to the main `/monitors` endpoint.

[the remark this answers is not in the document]
shabd67: its not being used anything. 

we should remove below endpoint as well, na?
`
  @GetMapping("/tenant/{tenantId}/monitor-templates/{uuid}")
  @ApiOperation(value = "Gets Monitor using template for Tenant")
`

jjbuchan:
@shabd67 I see several uses in policy-management:

image


...and based on that there should be more changes in this PR as well because the names of the APIs used will be updated.

src/main/java/com/rackspace/salus/monitor_management/web/controller/MonitorApiController.java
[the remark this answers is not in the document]
jjbuchan: Yes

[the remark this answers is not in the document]
jjbuchan: Both these endpoints were likely leftover from the previous design where policy monitors were not quite the same as a "normal" monitor.

itzg (commented):
Changes look good from my previous review, but we'll let @jjbuchan provide final approval.

itzg (approved):

PR #226, opened 2020-11-02
Send MonitorChangeEvents on all write operations

jjbuchan:
# What

Send `MonitorChangeEvent`s to kafka for any create, clone, update, delete.

Also formatted MonitorEventProducer.java so you'll want to ignore white space changes.

# How to test

Updated existing tests to additionally check for the new producer operations.

This addition identified some previous tests were incorrect, so I've updated them as well.

# Why

Event-Engine-Processor will read these and clear its own caches whenever a monitor changes.

That service only really needs update/delete events but I've included them in all places to make it consistent with what is needed in the task service.

### DIFF
--- a/src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
+++ b/src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
@@ -36,6 +36,7 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.argThat;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.reset;
@@ -71,6 +72,7 @@ import com.rackspace.salus.telemetry.etcd.services.ZoneStorage;
 import com.rackspace.salus.telemetry.etcd.types.EnvoyResourcePair;
 import com.rackspace.salus.telemetry.etcd.types.ResolvedZone;
 import com.rackspace.salus.telemetry.messaging.MonitorBoundEvent;
+import com.rackspace.salus.telemetry.messaging.MonitorChangeEvent;
 import com.rackspace.salus.telemetry.messaging.ResourceEvent;
 import com.rackspace.salus.telemetry.model.AgentType;
 import com.rackspace.salus.telemetry.model.ConfigSelectorScope;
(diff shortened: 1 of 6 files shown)


src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -3252,41 +3349,74 @@ public void testSearchOnId() {
   }
 
   @Test
-  public void deleteAllMonitorsForTenant() {
-    when(boundMonitorRepository
-        .findAllByTenantIdAndMonitor_IdIn(any(), any(), any()))
+  public void deleteAllMonitorsForTenant_withEvents() {
```
jjbuchan, line 3352: Due to the mock only returning `Page.empty()` this test and the one below it were not adequately testing the with or without events option before.

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -295,6 +297,8 @@ private void createMonitorsForTenant(int count, String tenantId) {
       create.setInterval(Duration.ofSeconds(60));
       create.setResourceId(null);
       monitorManagement.createMonitor(tenantId, create);
+      // the mocked operations triggered here should not be verified by tests
+      reset(monitorEventProducer, boundMonitorRepository);
```
jjbuchan, line 301: This is what helps show the issues in the two tests I had to fix.

itzg (approved):

src/test/java/com/rackspace/salus/monitor_management/services/MonitorManagementTest.java
```diff
@@ -3252,41 +3349,74 @@ public void testSearchOnId() {
...
+    assertThat(expectedMonitorIds).hasSize(20);
 
     monitorManagement.removeAllTenantMonitors("t-1", true);
 
     Page<Monitor> result = monitorManagement.getMonitors("t-1", Pageable.unpaged());
 
     assertThat(result.getNumberOfElements(), equalTo(0));
-    verify(monitorEventProducer, times(20)).sendMonitorEvent(any());
+
+    // this is called once for every empty result and once more for the non-empty result
```
GeorgeJahad, line 3374: not following how this adds up to 21.  seems like it should be once for each monitor or 20.

jjbuchan: Because it's hit by this - https://github.com/racker/salus-telemetry-monitor-management/blob/27de75b4597891a0acce9ef1d121ef1a7b74af95/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java#L1920-L1932

At least 2 hits when a bound monitor is returned.

If mockito allowed you to use `times` for `when` like it does for `verify` I would have set it up to return an item for each request, which would update this assert to be `40`, but since it doesn't I made it only return an item for the first one (hence the double repository hit) but nothing for all the rest (which only hit it once).

GeorgeJahad (approved):

PR #227, opened 2020-11-04
Metadata policy is being cashed even after deletion

shabd67:
# Resolves

[SALUS-1033](https://jira.rax.io/browse/SALUS-1033)

# What

BugFix - Metadata policy is being cashed even after deletion

### DIFF
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1565,6 +1565,7 @@ public class MonitorManagement {
     log.info("Handling metadata policy event={}", event);
 
     String tenantId = event.getTenantId();
+    policyApi.getEffectiveMonitorMetadataMap(tenantId, event.getTargetClassName(), event.getMonitorType(), false);
     List<MonitorMetadataPolicyDTO> effectivePolicies = policyApi.getEffectiveMonitorMetadataPolicies(tenantId, false);
 
     Optional<MonitorMetadataPolicyDTO> policyOptional = effectivePolicies.stream()
--- a/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
@@ -1565,7 +1565,7 @@ public class MonitorManagement {
     log.info("Handling metadata policy event={}", event);
 
     String tenantId = event.getTenantId();
-    policyApi.getEffectiveMonitorMetadataMap(tenantId, event.getTargetClassName(), event.getMonitorType(), false);
+    policyApi.evictEffectiveMonitorMetadataMap(tenantId, event.getTargetClassName(), event.getMonitorType());
     List<MonitorMetadataPolicyDTO> effectivePolicies = policyApi.getEffectiveMonitorMetadataPolicies(tenantId, false);
 
     Optional<MonitorMetadataPolicyDTO> policyOptional = effectivePolicies.stream()
--- a/src/main/java/com/rackspace/salus/monitor_management/utils/MetadataUtils.java
+++ b/src/main/java/com/rackspace/salus/monitor_management/utils/MetadataUtils.java
@@ -211,7 +211,7 @@ public class MetadataUtils {
     } else if (!patchOperation &&
         monitor.getMonitorMetadataFields() != null &&
         !monitor.getMonitorMetadataFields().isEmpty()) {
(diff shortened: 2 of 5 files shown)


src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1565,6 +1565,7 @@ void handleMetadataPolicyEvent(MetadataPolicyEvent event) {
     log.info("Handling metadata policy event={}", event);
 
     String tenantId = event.getTenantId();
+    policyApi.getEffectiveMonitorMetadataMap(tenantId, event.getTargetClassName(), event.getMonitorType(), false);
```
itzg, line 1568: Rather than utilizing the "side effect" of the `@CacheEvict` on this method I'd prefer you add and call a specific method called something like `evictEffectiveMonitorMetadataMap` that just has the `@CacheEvict` on it. Admittedly it's an odd fit for the API interface, but will make the intent so much clearer.

itzg (changes requested):
It seems like a unit test should have been impacted by the additional call to `PolicyApi`. Can you double check that? Maybe there's a `verifyNoMoreInteractions` missing on the corresponding unit test.

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
shabd67: Should I remove `@CacheEvict ` from `getEffectiveMonitorMetadataMap` if we are using `@CacheEvict` on `evictEffectiveMonitorMetadataMap ` ?

itzg: Have you done a usage search to see if the "use cache" parameter is ever passed as false? If it isn't, then yes, it seems safe/good to remove the annotation and possibly adjust the other on it.

shabd67: yes, checked for "use cache" parameter passed as false, its not there apart from test cases. So, will be safe to delete that annotation from `getEffectiveMonitorMetadataMap`

itzg: Sounds good then. Can also remove the parameter unless that's already what you're thinking. Hopefully the ripple effect of removing the parameter won't cause too many tedious changes to refactor.

shabd67:
> It seems like a unit test should have been impacted by the additional call to `PolicyApi`. Can you double check that? Maybe there's a `verifyNoMoreInteractions` missing on the corresponding unit test.

It seems there is no corresponding unit test case for this scenario.

jjbuchan (commented):
> It seems there is no corresponding unit test case for this scenario.

If you see there are no tests for a particular function you should always aim to add some.

---

Once my comments in https://github.com/racker/salus-policy-management/pull/49 are addressed I'd expect there to me some more changes in this PR as well.  Please push changes over there prior to requesting a review over here.

Thanks

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
```diff
@@ -1565,6 +1565,7 @@ void handleMetadataPolicyEvent(MetadataPolicyEvent event) {
     log.info("Handling metadata policy event={}", event);
 
     String tenantId = event.getTenantId();
+    policyApi.evictEffectiveMonitorMetadataMap(tenantId, event.getTargetClassName(), event.getMonitorType());
```
jjbuchan, line 1568: One other thing... it might be possible to move this line down into the `if (policyOptional.isEmpty()) {` condition.

Can you look into that to see if my thought is correct.

If it is a comment should be added to explain why it is called there.

itzg: You **must** reply to all PR comments explaining why you won't make the change or what change you plan. PR's will not be approved with unanswered comments.

itzg (commented):
Address all comments

src/main/java/com/rackspace/salus/monitor_management/services/MonitorManagement.java
shabd67: I think reverse should be applicable i.e. if the policy is relevant to this tenant, in that case we should remove it from cache. So irrespective of that, I have removed that entry from cache whether that policy is applicable for that tenant or not.

itzg: That's reasonable. Thanks.

itzg (approved):

PR #228, opened 2020-11-19
made deeper response message from JDBC error in REST Response.

shabd67:
# Resolves

[SALUS-1067](https://jira.rax.io/browse/SALUS-1067)

# What

made deeper response message from JDBC error in REST Response.

# How

Used getRootCause().getMessage to get original throwable message


itzg (approved):
Very good, it was already in an if-instanceof

